PR #142, opened 2023-04-27
⚡️ Improve docker incremental build time

coyotte508:


martinkozle:
Can this PR be given higher priority? Because currently making changes to the config and rebuilding the image we need to wait for it install all dependencies every time.
Even at the current state without fixing the review comments it would solve the problem on the master branch. I believe it should be merged as is for now.

coyotte508:
Ok @martinkozle 

Feel free to open another PR for further improvements :)

PR #193, opened 2023-05-11
Login flow

### LINKED ISSUE #191: Sign in Modal and frontend

Grsmto:
This PR adds support for user signin to their HF account.

If you want to test locally you will need moon-landing running on the PR mentionned above (and follow instructions).
You will need to set the following env variables as well:
```
OPENID_CLIENT_ID=id
OPENID_CLIENT_SECRET=secret
```

- add Openid Connect flow with `openid-client` lib like on moon-landing (csrf token, auth url, etc.)
- for existing users, will merge their profile with their HF infos
  - migrate their conversations to a user id
  - migrate their settings to a user id
- for new users, will create a new user entry in DB
- for existing users (ex a user already signed in before but clearer its cookies and signin again) it will update entry in DB

image

closes #191


src/lib/server/auth.ts
```diff
@@ -1,9 +1,93 @@
...
+
+// Mostly taken from https://github.com/huggingface/moon-landing/tree/main/server/lib/Auth.ts
+export function generateCsrfToken(sessionId: string): string {
+	const data = { expiration: addDays(new Date(), 1).getTime(), sessionId };
+	const token = JSON.stringify(data);
+
+	return Buffer.from(
+		JSON.stringify({ data, signature: instantSha256(token + "##" + AUTH_SECRET) })
+	).toString("base64");
+}
```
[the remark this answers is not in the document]
Grsmto: Was it on purpose that you removed `AUTH_SECRET` from the signature?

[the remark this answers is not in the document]
coyotte508: yes

.env
```diff
@@ -9,6 +9,8 @@ HF_ACCESS_TOKEN=#hf_<token> from from https://huggingface.co/settings/token
 # Parameters to enable "Sign in with HF"
 HF_CLIENT_ID=
 HF_CLIENT_SECRET=
+AUTH_SECRET=
```
coyotte508, line 12:
```suggestion
```

src/hooks.server.ts
```diff
@@ -9,19 +9,21 @@ import { addYears } from "date-fns";
...
 
 	const user = await collections.users.findOne({ sessionId: event.locals.sessionId });
 
-	if (user) {
+	if (user?.hfUserId) {
 		event.locals.userId = user._id;
 	}
 
 	if (
+		!event.url.pathname.startsWith(`${base}/login`) &&
```
coyotte508, line 26: The redirect on `/login` can probably be moved to a `GET`, avoiding this change

src/lib/components/LoginModal.svelte
```diff
@@ -31,17 +28,15 @@
 		<p class="px-2 text-sm text-gray-500">
 			Your conversations will be shared with model authors unless you disable it from your settings.
 		</p>
-		<form action="{base}/settings" use:enhance method="POST">
-			<input type="hidden" name="ethicsModalAccepted" value={true} />
-			{#each Object.entries(settings) as [key, val]}
-				<input type="hidden" name={key} value={val} />
-			{/each}
+		<form action="{base}/login" use:enhance method="POST">
+			<input type="hidden" name="callbackUrl" value={base} />
```
coyotte508, line 32: I think we still want to allow users who don't define `HF_CLIENT_ID` to be able to run the app locally, so for them, the form should be unchanged.

Regarding `<input type="hidden" name="callbackUrl" value={base} />`: it's currently not used by the backend.

src/routes/login/callback/+server.ts
```diff
@@ -0,0 +1,89 @@
...
+		.object({
+			preferred_username: z.string(),
+			name: z.string(),
+			picture: z.string(),
+			sub: z.string(),
+		})
+		.parse(userData);
+
+	// find sessionId in db if existing and migrate it to a user
+	const anonymousUser = await collections.users.findOne({ sessionId: locals.sessionId });
```
coyotte508, line 39: It was something like this I had in mind:

```ts
const existingUser = collections.users.findOne({ sessionId: hfUserId });

await collections.users.updateOne(
	{ hfUserId },
	{ $set: { username, name, avatarUrl, updatedAt: new Date() }, $setOnInsert: { createdAt, sessionId } },
  { upsert: true }
);

if (existingUser) {
  // todo: set sessionId cookie to existingUser.sessionId, to merge sessions
} else {
  // migrate settings for session if they exist
}

await db.conversations.updateMany({ sessionId, userId: {$exists: false} }, { $set: { userId }, $unset: { sessionId: "" } })
```

- Move all conversations for `sessionId` to the existing or new user
- Move the settings for `sessionId` to the new user
- In case of existing user, change the `sessionId` in the cookie to the one of the existing user

src/hooks.server.ts
Grsmto: SvelteKit doc says:
> Actions always use POST requests, since GET requests should never have side-effects.

So I suppose if we want to keep using an action form this is necessary?

coyotte508: but we don't need to use an action form for it, it can just be a link to `/login`

src/routes/login/+server.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export async function GET({ url, locals }) {
+	// TODO: Handle errors if provider is not responding
+	const ssoAuthorizationUrl = await getOIDCAuthorizationUrl(
+		{ redirectURI: getRedirectURI(url) },
+		{ sessionId: locals.sessionId }
+	);
+
+	throw redirect(303, ssoAuthorizationUrl);
+}
```
coyotte508, line 12: Apologies :pray: 

We should go back to a POST, just to make sure the user clicked the "sign in" button and so we can set the `"ethicsModalAcceptedAt"` with peace of mind. (i.e. they didn't follow a link to hf.co/chat/login from someone else without reading the modal)

src/routes/+layout.server.ts
```diff
@@ -69,5 +69,6 @@ export const load: LayoutServerLoad = async ({ locals, depends, url }) => {
 			parameters: model.parameters,
 		})),
 		oldModels,
+		user: { id: locals.userId?.toString(), sessionId: locals.sessionId },
```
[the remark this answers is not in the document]
Grsmto: Good catch 🙏  it's not being used in the frontend, this was not intended.

src/hooks.server.ts
```diff
@@ -1,23 +1,23 @@
...
-export const handle: Handle = async ({ event, resolve }) => {
+const authorization: Handle = async ({ event, resolve }) => {
 	const token = event.cookies.get(COOKIE_NAME);
 
 	event.locals.sessionId = token || crypto.randomUUID();
 
 	const user = await collections.users.findOne({ sessionId: event.locals.sessionId });
 
-	if (user) {
+	if (user?.hfUserId) {
```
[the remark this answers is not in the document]
Grsmto: But user can be an anonymous user (present in DB) if they already used the app before we implemented this signin modal. It has a `sessionId` but not `hfUserId` in that case.
If you check the behaviour of this PR at the moment, as an anonymous user you can still use the app and view your conversations (and save them) in a read only state. You don't need to sign in until you want to interact.
I don't think we could keep this behaviour with the change you suggested?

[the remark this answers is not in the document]
coyotte508: > But user can be an anonymous user (present in DB) if they already used the app before we implemented this signin modal.

No? Currently we do not store any users in `db.users`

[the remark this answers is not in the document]
gary149: > If you check the behaviour of this PR at the moment, as an anonymous user you can still use the app and view your conversations (and save them) in a read only state. You don't need to sign in until you want to interact. I don't think we could keep this behaviour with the change you suggested?

I don't think we want this behavior. We want login to be mandatory.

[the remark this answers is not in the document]
coyotte508: yes we want at least the signin popup to show directly, maaaybe we could make it dismissable but then it's on the user to refresh to be able to have new chats

src/hooks.server.ts
```diff
@@ -41,7 +41,7 @@ export const handle: Handle = async ({ event, resolve }) => {
 			);
 		}
 
-		if (!event.url.pathname.startsWith(`${base}/settings`)) {
+		if (!requiresUser && !event.url.pathname.startsWith(`${base}/settings`)) {
 			const hasAcceptedEthicsModal = await collections.settings.countDocuments({
 				sessionId: event.locals.sessionId,
```
coyotte508, line 46:
```suggestion
				...authCondition(locals),
```

src/routes/login/callback/+server.ts
Grsmto: That should be better now!

src/hooks.server.ts
coyotte508: That way the `!requiresUser` above can be removed too

src/hooks.server.ts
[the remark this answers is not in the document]
gary149: No need to make it dismissable imo let's make it as simple as possible.

src/routes/+layout.svelte
```diff
@@ -128,14 +133,13 @@
 			on:editConversationTitle={(ev) => editConversationTitle(ev.detail.id, ev.detail.title)}
 		/>
 	</nav>
-	{#if currentError}
+	{#if $error === ERROR_MESSAGES.authOnly}
+		<LoginModal />
+	{:else if currentError}
```
coyotte508, line 138: `LoginModal` should pop up directly, not wait until there's an error (this will also remove the need for `ERROR_MESSAGES.authOnly`)

src/routes/login/+page.server.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const actions = {
+	default: async function ({ url, locals }) {
+		// TODO: Handle errors if provider is not responding
+		const ssoAuthorizationUrl = await getOIDCAuthorizationUrl(
+			{ redirectURI: getRedirectURI(url) },
+			{ sessionId: locals.sessionId }
+		);
+
+		throw redirect(303, ssoAuthorizationUrl);
```
coyotte508, line 12:
```suggestion
		const authorizationUrl = await getOIDCAuthorizationUrl(
			{ redirectURI: getRedirectURI(url) },
			{ sessionId: locals.sessionId }
		);

		throw redirect(303, authorizationUrl);
```

nit: this isn't really sso (there's also  a `SSOUserInformation` type)

src/routes/login/callback/+server.ts
```diff
@@ -0,0 +1,103 @@
...
+			{ sessionId: locals.sessionId },
+			{ $set: { userId: insertedId, updatedAt: new Date() }, $unset: { sessionId: "" } }
+		);
+
+		if (matchedCount) {
+			// migrate pre-existing conversations if any
+			await collections.conversations.updateMany(
+				{ sessionId: locals.sessionId },
+				{ $set: { userId: insertedId }, $unset: { sessionId: "" } }
+			);
```
coyotte508, line 89: This should **always** happen, whether the user has settings or not (they can have convos w/o settings with early version), and whether there is already an existing user or not

src/routes/login/callback/+server.ts
```diff
@@ -0,0 +1,103 @@
...
+			username,
+			name,
+			avatarUrl,
+			hfUserId,
+			sessionId: locals.sessionId,
+		});
+
+		// update pre-existing settings
+		const { matchedCount } = await collections.settings.updateOne(
+			{ sessionId: locals.sessionId },
```
coyotte508, line 80:
```suggestion
			authConditions(locals),
```

src/routes/login/callback/+server.ts
```diff
@@ -0,0 +1,103 @@
...
+			{ $set: { userId: insertedId, updatedAt: new Date() }, $unset: { sessionId: "" } }
+		);
+
+		if (matchedCount) {
+			// migrate pre-existing conversations if any
+			await collections.conversations.updateMany(
+				{ sessionId: locals.sessionId },
```
coyotte508, line 87:
```suggestion
				authConditions(locals),
```

src/routes/login/callback/+server.ts
Grsmto: Isn't this making things a bit more confusing? Since `authCondition` in this case should always return `{ sessionId: locals.sessionId }` anyway?

coyotte508: It would return `{sessionId: locals.sessionId, user: {$exists: false}}`

coyotte508:
Also, there should be a "log out" button in the navigation menu! (we can see later about adding the username + avatar)

src/routes/login/callback/+server.ts
Grsmto: Yes, I mean we are in the callback of the signin so why do we need to call a function to know if we have a user or not? We know we don't have one otherwise we couldn't be signing in? Or I'm missing something 😮

src/lib/utils/sha256.ts
```diff
@@ -1,7 +1,15 @@
...
 	const hashArray = Array.from(new Uint8Array(hashBuffer));
 	const hashHex = hashArray.map((bytes) => bytes.toString(16).padStart(2, "0")).join("");
 	return hashHex;
 }
+
+export function instantSha256(input: string | Buffer): string {
+	const sha256Stream = crypto.createHash("sha256");
+	sha256Stream.update(input);
+	return sha256Stream.digest().toString("hex");
+}
```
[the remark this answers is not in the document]
Grsmto: Yep that was a shameless copy/paste from moon-landing. I just tested and the two functions provide an equivalent output. Will update!

src/routes/login/callback/+server.ts
coyotte508: It's in case there are db documents with same sessionId but a userid already filled in, shouldn't happen, but if that's the case we don't want to change the user of the already migrated documents

Grsmto:
> Also, there should be a "log out" button in the navigation menu! (we can see later about adding the username + avatar)

cc @gary149 yeah it could be in the settings modal, or in the menu directly?

gary149:
> cc @gary149 yeah it could be in the settings modal, or in the menu directly?

I like to have a visual indicator that I'm logged in. I would do something like this:

image

coyotte508 (approved):
Thanks!

Regarding the tests you mentioned @Grsmto , will open another issue (it doesn't have to be in this PR)

PR #197, opened 2023-05-11
Update settings.json

ss9901:
Improved editor configuration to enable formatting on save, add support for Svelte components, and specify the workspace folder.


.vscode/settings.json
```diff
@@ -1,8 +1,10 @@
 {
+	"root": true,
```
coyotte508, line 2: image

Seems VSCode doesn't like it? It's unused

Issue #202, opened 2023-05-11
Help wanted: Installing `@huggingface` package from NPM registry


eertmanhidde:
👋🏻 

Sorry if I am opening a dumb issue but I was just looking into fixing some UI issues and not entirely sure how to run this project locally. I've created a `.env.local` with:
```
MONGODB_URL= 
HF_ACCESS_TOKEN=XXX
```
Haven't actually set the `MONGODB_URL` but did create an access token for HF.

Running into the following error when running `yarn`
```
yarn install v1.22.11
info No lockfile found.
warning package-lock.json found. Your project contains lock files generated by tools other than Yarn. It is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files. To clear this warning, remove package-lock.json.
[1/4] 🔍  Resolving packages...
error Couldn't find package "@huggingface/shared@*" required by "@huggingface/inference@^2.2.0" on the "npm" registry.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```
I suppose I need a secret or something for Yarn to be able to fetch that package from a different registry than NPM?

**use NPM instead of Yarn?**
Yes, I've also tried using NPM, ran into the same issue.

Again sorry if I am mistaking the readme and doing things wrong.

Thanks! 👋🏻

Grsmto:
Hey @eertmanhidde thanks for wanting to contribute! It's an issue with the dependency https://github.com/huggingface/huggingface.js we are looking into it!

coyotte508:
@eertmanhidde You need to use `pnpm`, not `yarn`

eertmanhidde:
@coyotte508 yeah I've did that already, experienced the same issue.

PR #211, opened 2023-05-12
📌 Specifiy packageManager to prevent yarn accidents

coyotte508:
To prevent things like #202 :)

Now `yarn` will explicitly display a message

(although I'd have liked to use pnpm :smiling_face_with_tear:)

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
 	"name": "chat-ui",
 	"version": "0.1.0",
 	"private": true,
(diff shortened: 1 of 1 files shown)


Grsmto:
Tbf user was already warned in its original error message as seen in the linked issue:
```
warning package-lock.json found. Your project contains lock files generated by tools other than Yarn. ...
```
But I'm fine adding this as it doesn't really have any drawback. In the same spirit we are prob missing an `engines` prop.

PR #220, opened 2023-05-14
Fix chat input

TP-O:
Before:
image

After:
image

### DIFF
--- a/src/lib/components/chat/ChatInput.svelte
+++ b/src/lib/components/chat/ChatInput.svelte
@@ -37,7 +37,7 @@
 
 <div class="relative min-w-0 flex-1">
 	<pre
-		class="invisible whitespace-pre-wrap p-3"
+		class="scrollbar-custom invisible overflow-x-hidden overflow-y-scroll whitespace-pre-wrap break-words p-3"
 		aria-hidden="true"
 		style="min-height: {minHeight}; max-height: {maxHeight}">{(value || " ") + "\n"}</pre>
 


TP-O:
> Hey @TP-O thanks for your contribution! This seems like an elegant refactor! The only problem I can find is that on Firefox I sometimes have the scrollbar appearing where it's not needed: image
> 
> Also notice that to fix the bug on your screenshot while keeping the original code you could have added the break-words class on the pre element.

It seems that the `textarea` in Firefox shows the scrollbar even though the height of the content is equal to its height. The only solution I can think of is to add one more pixel to the `textarea` height :confused:

Grsmto:
Thanks for your fix, but now you get a jump when first inputting something in the textarea. :(

I gave it a try and I don't think you can avoid any kind of "jump". Because as soon as you defined a rounded "height" to the textarea, it will change from it's natural height (which is defined by the text height and can be on a sub-pixel height, like 45.56px for ex) to a fixed, rounded, height (result of .scrollHeight() which returns rounded values).

I guess that is why the original solution with the `pre` tag is more bullet proof because it keeps the natural height calculated by the browser rather than forcing it.

You can keep trying with your solution (maybe there is a way?) but otherwise as I said previously the original bug can be fixed with adding `break-words` to the `pre` tag.

TP-O:
@Grsmto I checked your fix but it didn't work in case the next line has 1 character.

image

Grsmto:
That was quite the roller coaster 😄 I had hope we could get your nice refactor in but it seems more complicated than I thought! This is only because the app is server-side rendered otherwise that would not be an issue (I think for example ChatGPT is using something like your solution!).
Anyway thanks a lot for your contribution!

PR #227, opened 2023-05-15
226 sign in tests

### LINKED ISSUE #226: Add tests for "sign in with HF"
Related to #193 

When someone signs in with HF:

- When no existing user and no existing settings, create user + settings with default values, migrate convos (make sure convos aren't accessible anymore without being signed in)
- When no existing user and existing settings, create user + migrate settings / convos (make sure convos aren't accessible anymore without being signed in)
- When existing user, migrate convos + update session cookie to existing user's session cookie (so that the existing user is not logged out). Do not migrate settings, at least do not overwrite user's settings

Grsmto:
- Add Vitest lib/setup
- Add basic tests for the DB updates going on when logging in. 

- For now this not tested in a browser environment, I just extracted the part of the code in a separate file.

closes #226


coyotte508:
Yes the dbs should have `-test` added.

For MongoDB you can take inspiration of what's done with moonlanding, it uses this docker-compose file to run a mongo with replica set

```
version: "3.9"
services:
  mongo:
    image: mongo:5.0.9
    ports:
      - ${LOCAL_MONGO_PORT:-27017}:27017
    command: --replSet rs0 --bind_ip_all
    healthcheck:
      test: test $$(echo "rs.initiate().ok || rs.status().ok" | mongo --quiet) -eq 1
      interval: 5s
    # logging:
```

It can be run in the `.github/workflows` file

package.json
```diff
@@ -11,7 +11,7 @@
 		"check:watch": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json --watch",
 		"lint": "prettier --plugin-search-dir . --check . && eslint .",
 		"format": "prettier --plugin-search-dir . --write .",
-		"test": "vitest"
+		"test": "MONGODB_URL=mongodb://0.0.0.0:27017/ MONGODB_DB_NAME=chat-ui-test vitest"
```
coyotte508, line 14:
```suggestion
		"test": "MONGODB_URL=mongodb://127.0.0.1:27017/ vitest"
```

`MONGODB_NAME` should be edited in `database.ts`: 

```
const db = client.db(MONGODB_DB_NAME + import.meta.vitest ? "-test" : "");
```

.github/workflows/lint-and-test.yml
```diff
@@ -40,5 +40,7 @@ jobs:
           npm ci
       - name: "Tests"
         run: |
+          mv .env _.env
           docker-compose --env-file=/dev/null up -d 
           npm run test
+          mv _.env .env
```
coyotte508: Why is this needed?

coyotte508:
> what do you think of using a Github Action [github.com/marketplace/actions/mongodb-in-github-actions](https://github.com/marketplace/actions/mongodb-in-github-actions) instead of docker-compose for this? Since we only need it for launching MongoDB anyway.

Well if we don't need replica sets we can just use docker instead of docker compose

vite.config.ts
```diff
@@ -1,5 +1,5 @@
 import { sveltekit } from "@sveltejs/kit/vite";
-import { defineConfig } from "vite";
+import { defineConfig } from "vitest/config";
```
coyotte508, line 2: Is it ok to do this?

coyotte508 (approved):
Nice !

vite.config.ts
Grsmto: I think it's ok but we actually don't need this change since we are using only default config! I'll revert.

PR #235, opened 2023-05-16
✨ Add pre-commit hook to run Prettier/Lint

Grsmto:
- Add commit hook that will run Prettier automatically on staged files only.


coyotte508:
:grimacing: :grimacing: :grimacing: :grimacing: 

Is it possible to make it _opt-in_?

Grsmto:
I'm not sure but happy to have it opt-in if we can. What issue do you have with it?

The biggest issue that this solves is that since we now have a CI that performs lint checks, you constantly have to back and forth between code and PR cause of linting issues. When an issue arise in the CI, you have to receive an email notification, then fix it, etc. The earlier you are aware of the issue the better imo!

Now that I said it, I realised this PR is missing the linting 😅

Grsmto:
Ok yeah I don't want to be pushy with this if it only benefits to me. We don't want to slow down commits to the point it makes you want to commit less often.
Also this was intended to catch Eslint "non-autofixable" issues, just so you avoid getting spammed by Github that you forgot to remove an unused variable. I'll make it opt-in or not do it at all and deal with it then.

PR #236, opened 2023-05-16
🐛 Fix "signin with HF" within space + CSRF

coyotte508:


src/routes/login/callback/+server.ts
```diff
@@ -31,9 +31,9 @@ export async function GET({ url, locals, cookies }) {
 		throw error(403, "Invalid or expired CSRF token");
 	}
 
-	const { userData } = await getOIDCUserData({ redirectURI: getRedirectURI(url) }, code);
+	const { userData } = await getOIDCUserData({ redirectURI: isValidToken.redirectUrl }, code);
```
Grsmto, line 34: And here then `if (!token.redirectUrl)` maybe instead of `isValidToken`?

src/routes/login/+page.server.ts
```diff
@@ -1,11 +1,13 @@
 import { redirect } from "@sveltejs/kit";
-import { getOIDCAuthorizationUrl, getRedirectURI } from "$lib/server/auth";
+import { getOIDCAuthorizationUrl } from "$lib/server/auth";
+import { base } from "$app/paths";
 
 export const actions = {
-	default: async function ({ url, locals }) {
+	default: async function ({ url, locals, request }) {
 		// TODO: Handle errors if provider is not responding
+		const referer = request.headers.get("referer");
```
Grsmto, line 8: Currently you get redirected to the frameless space instead of https://huggingface.co/spaces/. I suppose we would need another env var to solve that...? Not a big deal anyway.

Grsmto (approved):

PR #237, opened 2023-05-17
Add initial support for web browsing

nsarrazin:
https://github.com/huggingface/chat-ui/assets/25119303/d56b4f56-099d-4718-8577-b92a3f21a55c

## Why
Since the model has a knowledge cutoff and has a tendency to hallucinate it would be great to complement the chat history with relevant results from the web.

## How
We use serpapi for performing the search query. The LLM determines which query best answers the question, and we pass this query to serpapi. The results are parsed and injected into the conversation prompt. 

The web search feature can be toggled using a switch. If the key is not defined, the feature is disabled globally.

## What's missing
* Reading the content of pages. Right now we only inject the headlines from google and the text snippet that goes with it. The answers would be a lot more relevant if it could read the actual pages.
* Some UI indication that something is going on. Inferring the search query and then executing the search takes time and right now to the user it's not obvious what is happening.
* Maybe some rate limit mechanism on search-enhanced queries, to avoid burning through the API credits too fast ?

## How to test
* Create a [serpapi](https://serpapi.com/) account
* Get you api key and paste it in `.env.local` under `SERPAPI_KEY`
* You can now ask questions with and without web search by toggling the switch in the ui.
* You can see what search query is used and what snippets get injected in the prompt in the console.


src/lib/server/searchWeb.ts
```diff
@@ -0,0 +1,34 @@
...
+export async function getQueryFromPrompt(messages: Pick<Message, "from" | "content">[]) {
+	let prompt = "What query would you ask Google to answer this question?\n\n";
+
+	messages.forEach((message) => {
+		prompt += `${message.from}: ${message.content}\n`;
+	});
+
+	prompt += "\nQuery:";
+
+	return await generateFromDefaultEndpoint(prompt);
```
Grsmto, line 33: I don't know if it's intended but it seems like when you have more than one message, the prompt sent to get the Google query is including the assistant answer. So it seems like it gets confused. I have been testing with this example:
- What is the capital of Spain?
- What are the colors of the football squad?

And the prompt is something like:
```
What query would you ask Google to answer this question?

user: what is the capital of Spain?
assistant:  The current capital of Spain since the year 1916 is Madrid. This was established by the law of 10/04/1617 issued by Felipe III. However before this time, other Spanish kings chose Segovia, Toledo, Valladolid & Cordoba to exercise their rule as well as residences such as El Pardo
user: what are the colors of the football squad?

Query:
```

nsarrazin: I changed it so that it only uses the user prompts to formulate the web query. Might make it easier to get a good query out.

src/routes/conversation/[id]/+page.svelte
```diff
@@ -43,6 +44,7 @@
 			{
 				id: messageId,
 				is_retry: isRetry,
+				useSearch: $webSearchParameters.useSearch,
```
Comment (coyotte508) [inline src/routes/conversation/[id]/+page.svelte:47]: Let's keep casing consistent among the parameters.

Here it's snake case because the other options passed to inference API are snake case too

src/lib/server/searchWeb.ts
```diff
@@ -0,0 +1,39 @@
...
+
+	// Show result as JSON
+	const response = await getJson("google", params);
+
+	return response;
+}
+
+export async function getQueryFromPrompt(messages: Pick<Message, "from" | "content">[]) {
+	let prompt =
+		"<|prompter|>The following messages were written by a user, trying to answer a question.\n";
```
coyotte508, line 26:
```suggestion
		`${defaultModel.userMessageToken}The following messages were written by a user, trying to answer a question.\n`;
```

src/lib/server/searchWeb.ts
```diff
@@ -0,0 +1,39 @@
...
+
+	messages
+		.filter((message) => message.from === "user")
+		.forEach((message) => {
+			prompt += `<|${message.from}|> ${message.content}</s>`;
```
coyotte508, line 31:
```suggestion
			prompt += `<|user|> ${message.content}${model.messageEndToken ?? ""}`;
```

?

(if `message.from` was from the assistant, we'd want to use `defaultModel.assistantMessageToken` instead)

We also probably want to default `model.messageEndToken` to `""` in `server/models.ts` to simplify code here & in `buildPrompt`

src/lib/server/searchWeb.ts
```diff
@@ -0,0 +1,39 @@
...
+		});
+
+	prompt +=
+		"<|prompter|>What plain-text short (less than 10 words) sentence would you input into Google to answer the last question? Answer with just the query.\n</s>";
+	prompt += "<|assistant|>";
```
coyotte508, line 36: Same comments about using `model.messageEndToken`, `model.userMessageToken` and `model.assistantMessageToken`.

Also, not sure we need to add `\n` before `model.messageEndToken` (does it improve results? if not, we can keep to building the meta-prompt with an implementation as close as possible to existing `buildPrompt`)

src/routes/conversation/[id]/summarize/+server.ts
```diff
@@ -1,16 +1,12 @@
...
-import { modelEndpoint } from "$lib/server/modelEndpoint";
+import { generateFromDefaultEndpoint } from "$lib/server/generateFromDefaultEndpoint.js";
 import { defaultModel } from "$lib/server/models";
-import { trimPrefix } from "$lib/utils/trimPrefix";
-import { trimSuffix } from "$lib/utils/trimSuffix";
-import { textGeneration } from "@huggingface/inference";
 import { error } from "@sveltejs/kit";
 import { ObjectId } from "mongodb";
 
-export async function POST({ params, locals, fetch }) {
```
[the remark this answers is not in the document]
Comment (nsarrazin) [reply src/routes/conversation/[id]/summarize/+server.ts:13]: I'm calling the endpoint from `generateFromDefaultEndpoint` so `fetch` was an unused var. But then indeed I'm using the wrong fetch implementation.

Is there a clean way to access the fetch implementation in the function above, or should I just pass `fetch` as an argument to `generateFromDefaultEndpoint`?

src/lib/server/searchWeb.ts
nsarrazin: Fixed those, i'm using the correct model tokens now.

src/routes/conversation/[id]/+page.svelte
Comment (nsarrazin) [reply src/routes/conversation/[id]/+page.svelte]: Fixed!

src/lib/buildPrompt.ts
```diff
@@ -25,12 +45,65 @@ export function buildPrompt(
...
+			const dom = new JSDOM(htmlString ?? "");
+			const body = dom.window.document.querySelector("body");
+
+			if (body === null) throw new Error("body is null");
+
+			const text = (naiveInnerText(body) ?? "").replace(/ {2}|\r\n|\n|\r/gm, "");
+
+			console.log(text);
+			// if there is a text, summarize it,
+			// TODO: This could be improved by using a dedicated summarization model if we have an endpoint for it.
```
julien-c, line 75: we could spawn one on Inference Endpoints if needed

src/lib/components/WebSearchToggle.svelte
```diff
@@ -0,0 +1,31 @@
...
+
+<label class="relative inline-flex cursor-pointer items-center">
+	<input
+		type="checkbox"
+		value=""
+		class="peer sr-only"
+		on:click={() => {
+			webSearchParameters.update((value) => ({ ...value, useSearch: !value.useSearch }));
+		}}
+		checked={$webSearchParameters.useSearch}
```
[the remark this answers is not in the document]
nsarrazin: Done!

src/lib/buildPrompt.ts
nsarrazin: So far it seems to work okay like this. Maybe if we have a fine tuned model already available for summarization we could spawn one in the future but it's not a requirement to ship this, and I don't know if that's something we already have ?

gary149 (approved):

PR #239, opened 2023-05-17
Feature/delete all conv

Grsmto:
image

<details> 
  <summary>or this option</summary>

</details>

image

### DIFF
--- a/src/lib/components/SettingsModal.svelte
+++ b/src/lib/components/SettingsModal.svelte
@@ -10,56 +10,94 @@
 
 	export let settings: Pick<Settings, "shareConversationsWithModelAuthors">;
 
+	let isConfirmingDeletion = false;
+
 	const dispatch = createEventDispatcher<{ close: void }>();
 </script>
 
 <Modal on:close>
-	<form
-		class="flex w-full flex-col gap-5 p-6"
-		use:enhance={() => {
-			dispatch("close");
-		}}
-		method="post"
-		action="{base}/settings"
-	>
+	<div class="flex w-full flex-col gap-5 p-6">
 		<div class="flex items-start justify-between text-xl font-semibold text-gray-800">
 			<h2>Settings</h2>
 			<button type="button" class="group" on:click={() => dispatch("close")}>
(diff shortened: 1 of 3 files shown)


src/lib/components/SettingsModal.svelte
```diff
@@ -10,56 +10,94 @@
...
+					class="underline decoration-gray-300 hover:decoration-gray-700">Open Assistant</a
+				>.
+			</p>
+			<button
+				type="submit"
+				class="mt-2 rounded-full bg-black px-5 py-2 text-lg font-semibold text-gray-100 ring-gray-400 ring-offset-1 transition-all focus-visible:outline-none focus-visible:ring hover:ring"
+			>
+				Apply
+			</button>
+		</form>
```
coyotte508, line 66: Nit: we can still have a global form two ways:

- Adding `formaction` prop to a specific submit button => will change the action of the form when clicking the button
- Adding a `form` prop to an input => will attach the input to the form, even if it's located in another form or outside any forms

It can also help go towards the no-js route - I know that settings can only be opened with JS at the moment, but we could also switch it to `href="?settings"` later-on for people w/o JS

src/lib/components/SettingsModal.svelte
```diff
@@ -10,56 +10,94 @@
...
 		<button
-			type="submit"
-			class="mt-2 rounded-full bg-black px-5 py-2 text-lg font-semibold text-gray-100 ring-gray-400 ring-offset-1 transition-all focus-visible:outline-none focus-visible:ring hover:ring"
+			type="button"
+			class="text-lg text-gray-500 underline hover:no-underline"
+			on:click={() => (isConfirmingDeletion = true)}
 		>
-			Apply
+			Delete all conversations
 		</button>
```
coyotte508, line 73: Maybe something like this:

```ts
<form action="/..../?delete" on:submit|preventDefault={() => isConfirmationDeletion = true}>
  <button type="submit">Delete all conversations</button>
</form>
```

Just to facilitate No-JS later on

coyotte508 (approved):
LGTM, thanks for looking into it!

gary149 (approved):
ok

PR #240, opened 2023-05-17
🐛 Fix settings modal not resetting its state

Grsmto:
Fix the following bug:
- Open the settings modal, change "Share conversations with model authors", close the modal without "applying"
- Re-open modal, setting is still the same, making you think it was taken into consideration

This is because the input is mutating the `settings` prop.

### DIFF
--- a/src/lib/components/SettingsModal.svelte
+++ b/src/lib/components/SettingsModal.svelte
@@ -10,6 +10,8 @@
 
 	export let settings: Pick<Settings, "shareConversationsWithModelAuthors">;
 
+	let shareConversationsWithModelAuthors = settings.shareConversationsWithModelAuthors;
+
 	const dispatch = createEventDispatcher<{ close: void }>();
 </script>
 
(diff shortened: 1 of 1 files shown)


src/lib/components/SettingsModal.svelte
```diff
@@ -10,6 +10,8 @@
 
 	export let settings: Pick<Settings, "shareConversationsWithModelAuthors">;
 
+	let shareConversationsWithModelAuthors = settings.shareConversationsWithModelAuthors;
```
coyotte508, line 13:
```suggestion
	$: shareConversationsWithModelAuthors = settings.shareConversationsWithModelAuthors;
```

Maybe? If the data is reloaded for X or Y

coyotte508 (approved):

src/lib/components/SettingsModal.svelte
Grsmto: Ahh actually! this variable represents the local state of the form, it shouldn't get updated by anything external since it represents user's choice while it is configuring its settings. Once user "applied" (ie when modal is closed), then at this point it can be updated.

PR #242, opened 2023-05-18
Make data sharing disclosure a little clearer

yjernite:
Quick proposed update so the disclosure draws the eye a little more - currently it's easy to miss

image

### DIFF
--- a/src/lib/components/LoginModal.svelte
+++ b/src/lib/components/LoginModal.svelte
@@ -33,8 +33,8 @@
 			misinformation. Do not use this application for high-stakes decisions or advice.
 		</p>
 		<p class="px-2 text-sm text-gray-500">
-			⚠️ <b>Data sharing:</b> Your conversations will be shared with the authors of the model
-			you use unless you disable data sharing in the HuggingChat settings.
+			⚠️ <b>Data sharing:</b> Your conversations will be shared with the authors of the model you use
+			unless you disable data sharing in the HuggingChat settings.
 		</p>
 		<form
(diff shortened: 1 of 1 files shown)


yjernite:
Simple rephrasing to start the conversation, but I'd love to hear from @ImaBello how the text could be improved!

src/lib/components/LoginModal.svelte
```diff
@@ -33,7 +33,8 @@
 			misinformation. Do not use this application for high-stakes decisions or advice.
 		</p>
 		<p class="px-2 text-sm text-gray-500">
-			Your conversations will be shared with model authors unless you disable it from your settings.
+			⚠️ <b>Data sharing:</b> Your conversations will be shared with the authors of the model
+			you use unless you disable data sharing in the HuggingChat settings.
```
julien-c, line 37:
```suggestion
   Your conversations will be shared with model authors unless you disable data sharing in the HuggingChat settings.
```

ImaBello:
I suggest:

"By default, your conversations are shared with the authors of the model to improve their training data and model. You can disable data sharing in the HuggingChat settings. See this [privacy note ](https://huggingface.co/chat/privacy)for further details."

cc @annatrdj for visibility

annatrdj:
Thanks for the tag Ima! I actually prefer the initial language :) 
1/ "unless you disable" makes it clear that they *have* to do something to stop data sharing
2/ no mention of the purpose to keep it vague (and in case data sharing is for another purpose)

PR #244, opened 2023-05-19
Fix stop generating button

TP-O:
The current behavior overlays user input, so I think making it similar to chatGPT would be better.
image

New in desktop
image

New in mobile
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,6 +26,7 @@
 			},
 			"devDependencies": {
 				"@iconify-json/carbon": "^1.1.16",
+				"@iconify-json/eos-icons": "^1.1.6",
 				"@sveltejs/adapter-node": "^1.2.4",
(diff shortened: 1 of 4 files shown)


Grsmto:
Hey @TP-O thanks for your contrib again!
There is no way you could do this without relying on JS to display the desktop or mobile version? The difference between the two versions doesn't seem to huge to need some JS (unless the positioning is a deal breaker).

TP-O:
> Hey @TP-O thanks for your contrib again! There is no way you could do this without relying on JS to display the desktop or mobile version? (and instead do with CSS Media Queries). The difference between the two versions doesn't seem to huge to need some JS (unless the positioning is a deal breaker).

I will fix it :rofl:

gary149:
LGTM, nice one.

PR #249, opened 2023-05-19
🐛 Vote not working with new messages

Grsmto:
- Fix the retry button not working after refactor
- Fix the voting not working for new messages: the problem was that new messages have different ids on client and server (because we generate them randomly). 
Solution doesn't seem like the most efficient but it was the most elegant one. Other solutions would be to modify the tokens stream to inject the message id or generate a deterministic id from both client/server based on the message value...But not sure those are much better.


coyotte508:
if you want a quick fix, passing the responses' id along with the message's id from the front-end to the backend is probably the simplest.

coyotte508 (approved):
Thanks!

PR #251, opened 2023-05-19
Feature/rate limit

Grsmto:
Very basic rate limitation for the message API.

### DIFF
--- a/src/routes/conversation/[id]/+server.ts
+++ b/src/routes/conversation/[id]/+server.ts
@@ -23,12 +23,16 @@ export async function POST({ request, fetch, locals, params }) {
 	const date = new Date();
 
 	// rate limit users by messages per day
-	if (locals.user) {
+	if (locals.user && MESSAGES_RATE_LIMIT) {
 		const messagesCount = await collections.messageEvents.countDocuments({
 			userId: locals.user._id,
 		});
-		if (messagesCount > parseInt(MESSAGES_RATE_LIMIT ?? 0)) {
-			throw error(429, ERROR_MESSAGES.overRateLimit);
+
+		if (messagesCount + 1 > parseInt(MESSAGES_RATE_LIMIT)) {
+			return new Response(JSON.stringify({ error: ERROR_MESSAGES.overRateLimit }), {
+				status: 429,
+				headers: { "Content-Type": "application/json" },
+			});
 		}
 	}
 


coyotte508:
A few things:

- Rate limit should be in `.env` (not defined = unlimited)
- Since it's only calls on posting a message that are limited, we don't need to define it in `hooks.server.ts`
- We can add a mongodb collection `messageEvents`, with a `expiresAfterSeconds` index on `createdAt` that is `3600*24`: We can then `collections.messageEvents.count({userId: ...})` to count the number of messages in the last day. The collection would only contain timestamps, userId (and maybe `messageId` / `conversationId`)

src/routes/conversation/[id]/+server.ts
```diff
@@ -14,12 +14,24 @@ import type { TextGenerationStreamOutput } from "@huggingface/inference";
...
 	const id = z.string().parse(params.id);
 	const convId = new ObjectId(id);
 	const date = new Date();
 
+	// rate limit users by messages per day
+	if (locals.user) {
+		const messagesCount = await collections.messageEvents.countDocuments({
+			userId: locals.user._id,
+		});
+		if (messagesCount > parseInt(MESSAGES_RATE_LIMIT ?? 0)) {
```
Comment (coyotte508) [inline src/routes/conversation/[id]/+server.ts:30]:
```suggestion
		if (messagesCount > parseInt(MESSAGES_RATE_LIMIT)) {
```

src/routes/conversation/[id]/+server.ts
```diff
@@ -14,12 +14,28 @@ import type { TextGenerationStreamOutput } from "@huggingface/inference";
...
+	// rate limit users by messages per day
+	if (locals.user && MESSAGES_RATE_LIMIT) {
+		const messagesCount = await collections.messageEvents.countDocuments({
+			userId: locals.user._id,
+		});
+
+		if (messagesCount + 1 > parseInt(MESSAGES_RATE_LIMIT)) {
+			return new Response(JSON.stringify({ error: ERROR_MESSAGES.overRateLimit }), {
+				status: 429,
+				headers: { "Content-Type": "application/json" },
```
Comment (Grsmto) [inline src/routes/conversation/[id]/+server.ts:34]: @coyotte508 FYI I had to do this instead of throwing an error because of the way errors are handle in hf.js [there](https://github.com/huggingface/huggingface.js/blob/main/packages/inference/src/tasks/custom/streamingRequest.ts#L33). Do you think hf.js should forward the error instead of changing it?

julien-c:
> * We can then `collections.messageEvents.count({userId: ...})` to count the number of messages in the last day.

Or we could just have added a `numOfMessagesSinceMidnight` (or a `numOfMessagesRemainingToday`) on users, i.e. compute rate limits on windowed days, not past 24 hours

and then either reset everyone to 0 as a cronjob, or store a additional lastMessageTime

Grsmto:
> > * We can then `collections.messageEvents.count({userId: ...})` to count the number of messages in the last day.
> 
> Or we could just have added a `numOfMessagesSinceMidnight` (or a `numOfMessagesRemainingToday`) on users, i.e. compute rate limits on windowed days, not past 24 hours
> 
> and then either reset everyone to 0 as a cronjob, or store a additional lastMessageTime

That's pretty much how I implemented it on the first commit. I have no problem either ways.

coyotte508:
No strong opinion, more of a product decision

src/hooks.server.ts
```diff
@@ -81,6 +82,35 @@ export const handle: Handle = async ({ event, resolve }) => {
...
+				{ _id: user._id },
+				{
+					$inc: {
+						requestCount: 1,
+					},
+					$set: {
+						lastSeenAt: new Date(),
+					},
+				}
+			);
```
coyotte508: In case we go the other way:

- `user.lastSeenAt` should be renamed to `lastMessageWrittenAt`
- The current code requires the user to wait a whole 24 hours after last message, rather than waiting until next day or 24 hours after first message
- The `requestCount` isn't actually reset, it should be reset  to `0` when going the next day

julien-c: no strong opinion either!

julien-c: pick whatever's the simplest technically, IMO

src/lib/stores/errors.ts
```diff
@@ -3,6 +3,7 @@ import { writable } from "svelte/store";
 export const ERROR_MESSAGES = {
 	default: "Oops, something went wrong.",
 	authOnly: "You have to be logged in.",
+	overRateLimit: "You have exceeded your rate limit. Try again later.",
```
coyotte508, line 6:
```suggestion
	overRateLimit: "You have exceeded the amount of messages per day. Try again tomorrow.",
```

src/routes/conversation/[id]/+server.ts
```diff
@@ -128,6 +143,16 @@ export async function POST({ request, fetch, locals, params }) {
 				},
 			}
 		);
+
+		// if messages API rate is limited, save message event
+		if (locals.user && typeof MESSAGES_RATE_LIMIT !== "undefined") {
```
Comment (coyotte508) [inline src/routes/conversation/[id]/+server.ts:148]:
```suggestion
		if (locals.user && MESSAGES_RATE_LIMIT) {
```

It won't be undefined, it'll be empty string at most I think

src/routes/login/callback/updateUser.spec.ts
coyotte508, line 1: can be reverted

src/routes/login/callback/updateUser.ts
coyotte508, line 1: can be reverted

coyotte508 (commented):
Good to merge after the changes requested I think!

src/hooks.server.ts
coyotte508: I think the collection solution is less error prone (even if the other one is much shorter w/ code)

PR #259, opened 2023-05-24
Add incremental build + multi layer docker for size reduction

XciD:
Reduce the docker layer size + enable incremental builds

PS: This `--mount=type=secret,id=DOTENV_LOCAL,dst=.env.local` fell old, we should not need secret at build time.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,7 +1,6 @@
 Dockerfile
 .vscode/
 .idea
-.env
 .gitignore
 LICENSE
 README.md
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,3 +4,5 @@ Dockerfile
 .gitignore
 LICENSE
 README.md
+node_modules/
+.svelte-kit/
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,7 @@
+Dockerfile
+.vscode/
+.idea
+.env
+.gitignore
+LICENSE
+README.md
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,18 +1,31 @@
 # read the doc: https://huggingface.co/docs/hub/spaces-sdks-docker
 # you will also find guides on how best to write your Dockerfile
-
-FROM node:19
-
-RUN npm install -g pm2
+FROM node:19 as builder-production
 
 WORKDIR /app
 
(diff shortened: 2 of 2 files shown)


.dockerignore
```diff
@@ -0,0 +1,7 @@
+Dockerfile
+.vscode/
+.idea
+.env
```
coyotte508, line 4:
```suggestion
```

coyotte508:
`COPY --from=builder /app/package.json /app/package.json` is not cached :thinking: 

```

--> RUN npm install -g pm2
CACHED

--> COPY --from=builder-production /app/node_modules /app/node_modules
CACHED

--> COPY --from=builder /app/package.json /app/package.json
DONE 3.3s

DONE 3.3s

--> COPY --from=builder /app/build /app/build
DONE 0.1s

--> Pushing image
DONE 0.7s

--> Exporting cache
DONE 0.9s
```

XciD:
I've updated, can you retry ?

coyotte508:
LGTM, @nsarrazin do you want to take a look at it?

nsarrazin:
Yep I'll have a quick look

Dockerfile
```diff
@@ -1,18 +1,31 @@
 # read the doc: https://huggingface.co/docs/hub/spaces-sdks-docker
 # you will also find guides on how best to write your Dockerfile
-
-FROM node:19
-
-RUN npm install -g pm2
+FROM node:19 as builder-production
```
nsarrazin, line 3: Any reason not to use node:19-slim for this builder stage ?

.dockerignore
```diff
@@ -0,0 +1,6 @@
+Dockerfile
+.vscode/
```
nsarrazin, line 2: Could we add `.svelte-kit` and `node_modules` to speed up local development as well ?

Dockerfile
XciD: slim container dont always have everything for build. I dont know here if it will work, but some deps needs gcc etc.

.dockerignore
XciD: yes

Dockerfile
nsarrazin: Ok fair enough, thanks!

nsarrazin (approved):
LGTM!

PR #152, opened 2023-05-02
add thumb up/down voting system for messages

Grsmto:
**desktop (on message hover or when voted)**

image

**mobile**

image

PR #265, opened 2023-05-24
Export to parquet: also export score

julien-c:
Follow-up to #151 to account for #152

### DIFF
--- a/src/routes/admin/export/+server.ts
+++ b/src/routes/admin/export/+server.ts
@@ -34,7 +34,14 @@ export async function POST({ request }) {
 		title: { type: "UTF8" },
 		created_at: { type: "TIMESTAMP_MILLIS" },
 		updated_at: { type: "TIMESTAMP_MILLIS" },
-		messages: { repeated: true, fields: { from: { type: "UTF8" }, content: { type: "UTF8" } } },
+		messages: {
+			repeated: true,
+			fields: {
+				from: { type: "UTF8" },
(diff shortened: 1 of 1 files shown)


src/routes/admin/export/+server.ts
```diff
@@ -34,7 +34,14 @@ export async function POST({ request }) {
 		title: { type: "UTF8" },
 		created_at: { type: "TIMESTAMP_MILLIS" },
 		updated_at: { type: "TIMESTAMP_MILLIS" },
-		messages: { repeated: true, fields: { from: { type: "UTF8" }, content: { type: "UTF8" } } },
+		messages: {
+			repeated: true,
+			fields: {
+				from: { type: "UTF8" },
+				content: { type: "UTF8" },
+				score: { type: "INT_8", optional: true },
```
julien-c, line 42: @lhoestq @severo Out of curiosity, does our viewer support `INT_8` (and optional) data cells in Parquet files?

coyotte508 (approved):

src/routes/admin/export/+server.ts
severo: do you have an example file to share?

lhoestq: also note that the viewer is not enabled on private datasets !

julien-c: yes i know!

PR #271, opened 2023-05-31
Shared convos can be opened without being logged in (#266, #270)

### LINKED ISSUE #266: Shared conversations should not require a login
Currently if you try to access a shared conversation and the `.env` is setup to support logging in, you will have to login before you can see it.

Since shared convos are read only, it makes sense to be able to see them without logging in first.

### LINKED ISSUE #270: Replying to a shared conversation leads to a 500 error
While trying to solve #266 I seem to have found a bug:

1. Create a new conversation, write a message
2. Share the resulting conversation, then open the link in a new tab
3. Try to send a message in the chat input of the shared convo -> 500 internal server error

I could replicate it when using the same account for sharing & trying to send the message, but I also replicated it by using two different accounts and I also get a 500.

nsarrazin:
Some users were complaining that you have to be logged in as a user to see a shared conversation.  This PR lets you see shared convos without logging in but as soon as you type in the chat input field, the login modal pops up.

~~This doesn't solve #270 where you can't currently reply to shared conversations anyway. I'm currently investigating a fix for this one.~~ Fixed #270 as well!

Closes #266 
Closes #270

### DIFF
--- a/src/routes/+layout.svelte
+++ b/src/routes/+layout.svelte
@@ -91,6 +91,11 @@
 	});
 
 	$: if ($error) onError();
+
+	const requiresLogin =
+		!$page.error &&
+		(data.requiresLogin ? !data.user : !data.settings.ethicsModalAcceptedAt) &&
+		!$page.route.id?.startsWith("/r/");
 </script>
 
 <svelte:head>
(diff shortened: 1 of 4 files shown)


src/routes/+layout.svelte
```diff
@@ -91,6 +91,11 @@
 	});
 
 	$: if ($error) onError();
+
+	const requiresLogin =
+		!$page.error &&
+		(data.requiresLogin ? !data.user : !data.settings.ethicsModalAcceptedAt) &&
```
gary149, line 97: It's preferable to use ternary only for assignments.

gary149 (approved):

Issue #247, opened 2023-05-19
UI refactor for menu items


nsarrazin:
![image](https://github.com/huggingface/chat-ui/assets/25119303/cac249f0-40e5-4265-9964-c6e05852b9fa)

It's hard to distinguish between past chats and the menu items when the bar is full. I think a low hanging fruit would be to add:
* A horizontal separator between the two
* Icons for the menu tabs 
* Maybe play around with the typography, increasing font weight on menu items for example.

coyotte508:
See also https://github.com/huggingface/chat-ui/issues/143

fredguth:
I also would move Theme inside Settings.

gary149:
Closing, will probably do a V2 redesign at some point.

Issue #143, opened 2023-04-27
Slight UI issue in left panel dark mode


coyotte508:
![image](https://user-images.githubusercontent.com/342922/234986004-b9f636e6-f907-4227-9334-068a49e653cd.png)

Or is this intended?

PR #272, opened 2023-06-01
Add icons & avatar to menu items

nsarrazin:
| Before  | After |
| image  | image  |

The goal was to make it a bit clearer where conversations end and where menu items begin. Hopefully this solves #247 #143


gary149:
We removed this icon to keep it minimalist (cc https://github.com/huggingface/chat-ui/pull/101) so I would close for now.

Issue #273, opened 2023-06-01
Documentation about how to configure custom model endpoints is missing


djmaze:
It seems it has been removed in https://github.com/huggingface/chat-ui/commit/fae93d9fc3be9a39d8efd9ab9993dea13f0ae844.

nsarrazin:
Thanks for opening an issue. I'm not sure what the issue is, could you tell me what you're missing from https://github.com/huggingface/chat-ui#custom-models ?

djmaze:
The `endpoints` key in the `MODELS` variable is not documented anymore.

averad:
@nsarrazin 

```
Running your own models
If you want to, you can even run your own models, by having a look at our endpoint project, [text-generation-inference](https://github.com/huggingface/text-generation-inference). You can then add your own endpoint to the MODELS variable in .env.local and it will be picked up as well.
```
As described above the endpoints key is not documented on either the chat-ui or text-generation-inference documentation.

PR #264, opened 2023-05-24
Added a few sections to the readme and reorganized it

nsarrazin:
Just a simple readme update to hopefully make it easier to get started with the project.

PR #288, opened 2023-06-10
➕ Update README.md to Include endpoints Variable

averad:
**📝  Summary:**
Information about adding endpoints to .env.local was removed in Pull request #264 - Added a few sections to the readme and reorganized it

**⚠️ Related Issue(s):**
#273 - Documentation about how to configure custom model endpoints is missing

**🔍 Code Change:**
[.env](https://github.com/huggingface/chat-ui/compare/main...averad:chat-ui:main#diff-e9cbb0224c4a3d23a6019ba557e0cd568c1ad5e1582ff1e335fb7d99b7a1055d)
`# "endpoints": [{"url": "http://127.0.0.1:8080/generate_stream"}],`

[README.md](https://github.com/huggingface/chat-ui/compare/main...averad:chat-ui:main#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5)
` "endpoints": [{"url": "http://127.0.0.1:8080/generate_stream"}],`

### DIFF
--- a/.env
+++ b/.env
@@ -25,7 +25,6 @@ MODELS=`[
     "datasetName": "OpenAssistant/oasst1",
     "description": "A good alternative to ChatGPT",
     "websiteUrl": "https://open-assistant.io",
-    # "endpoints": [{"url": "http://127.0.0.1:8080/generate_stream"}],
     "userMessageToken": "<|prompter|>",
     "assistantMessageToken": "<|assistant|>",
     "messageEndToken": "</s>",
--- a/.env
+++ b/.env
@@ -25,6 +25,7 @@ MODELS=`[
     "datasetName": "OpenAssistant/oasst1",
     "description": "A good alternative to ChatGPT",
     "websiteUrl": "https://open-assistant.io",
+    # "endpoints": [{"url": "http://127.0.0.1:8080/generate_stream"}],
     "userMessageToken": "<|prompter|>",
     "assistantMessageToken": "<|assistant|>",
     "messageEndToken": "</s>",
--- a/README.md
+++ b/README.md
@@ -85,6 +85,7 @@ MODELS=`[
     "datasetName": "OpenAssistant/oasst1",
     "description": "A good alternative to ChatGPT",
     "websiteUrl": "https://open-assistant.io",
+    "endpoints": [{"url": "http://127.0.0.1:8080/generate_stream"}],
     "userMessageToken": "<|prompter|>",
     "assistantMessageToken": "<|assistant|>",
     "messageEndToken": "</s>",
--- a/README.md
+++ b/README.md
@@ -129,11 +129,9 @@ If you want to, you can even run your own models, by having a look at our endpoi
 
 Custom endpoints may require authorization. In those situations, we will need to generate a base64 encoding of the username and password.
 
-```
-echo -n "USER:PASS" | base64
-```
+`echo -n "USER:PASS" | base64`
 
-VVNFUjpQQVNT
+> VVNFUjpQQVNT
 
 You can then add the generated information and the `authorization` parameter to your `.env.local`.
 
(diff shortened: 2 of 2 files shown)


.env
```diff
@@ -25,6 +25,7 @@ MODELS=`[
     "datasetName": "OpenAssistant/oasst1",
     "description": "A good alternative to ChatGPT",
     "websiteUrl": "https://open-assistant.io",
+    # "endpoints": [{"url": "http://127.0.0.1:8080/generate_stream"}],
```
nsarrazin, line 28: Having the endpoint commented out causes an internal server error with:
 
```Internal server error: Unexpected token # in JSON at position 214```

Maybe we could remove it from the `.env` but expand the readme a bit ?

Basically the behavior is that without an `endpoints` array defined, it will try to use the hosted hugging face API inference endpoint. (see [here](https://github.com/huggingface/chat-ui/blob/df5a2eb42b37a540e63674e47358ef55934ae38f/src/lib/server/modelEndpoint.ts#LL13C3-

If it's defined we use those endpoints instead!

File: README.md [hunk 2]
```diff
@@ -85,6 +85,7 @@ MODELS=`[
     "datasetName": "OpenAssistant/oasst1",
     "description": "A good alternative to ChatGPT",
     "websiteUrl": "https://open-assistant.io",
+    "endpoints": [{"url": "http://127.0.0.1:8080/generate_stream"}],
```
nsarrazin, line 88: So it would be dope if we could leave it out of the default `.env` and then add a note under "Running your own models" in the readme about the `endpoints` key (with the example you gave)

nsarrazin (commented):
Hey :wave: Thanks for finding this and contributing a fix!

I made a few comments, let me know if you need any help ! :smile:

README.md
averad: Good point, I will make the requested changes

a-ghorbani:
Thanks @averad. This helped me big time. If it's not too much trouble, could you possibly incorporate additional parameters of the `endpoints`, like `authorization` and `weight`? (otherwise I am happy to create a PR as well 😄).

For instance, if the endpoint requires a username and password, we should use the following steps.
Initially, we must generate a base64 encoding of the username and password.

```bash
> echo -n "USER:PASS" | base64
VVNFUjpQQVNT
```

Next, the endpoints settings should be configured as:

```json
"endpoints": [
    {
        "url": "https://HOST/generate_stream", 
        "authorization": "Basic VVNFUjpQQVNT"
    }
]

```

Finally, if we have more that one endpoint, the application employs a weighted random selection approach to select an "endpoint" from the "endpoints" array. Therefore, when multiple endpoints are present, we can use a 'weight' to each through the `weight` parameter. 
(@nsarrazin is my understanding of the `weight` param correct?)

nsarrazin:
@a-ghorbani Yes pretty much correct!

Would love to see this added to the readme, @averad I'm available if you need any help.

pervrosen:
Since the chat-ui supoorts many endpoints for a single model we need also the weight parameter for completeness.

"endpoints": [
    {
        "url": "https://HOST/generate_stream", 
        "authorization": "Basic VVNFUjpQQVNT",
        "weight": 1
    }
]

averad:
Good catch on the additional parameters, I will make sure they are added into the readme

.env
averad: I have removed the requested information from .env and and updated the readme. Let me know if you have any other concerns or feedback.

README.md
averad: I have removed the requested information and updated the readme. Let me know if you have any other concerns or feedback.

averad:
Great feedback everyone! I have rolled the information into the requested readme section and removed my suggested changes from .env

averad:
@nsarrazin want me to roll the requested change #300 into this PR?

If not, this PR is ready for review.

nsarrazin:
Thanks for contributing, this is looking pretty good! I'll merge it as is for now. :smile:

nsarrazin (approved):

PR #293, opened 2023-06-13
My changes

sstasik645:


nsarrazin:
@sstasik645 Hey it seems like you committed your HF auth token here, you might want to change it in [your profile](https://huggingface.co/settings/tokens)  to avoid abuse :sweat_smile:

PR #294, opened 2023-06-14
Add optional timestamps to messages

nsarrazin:
We keep track of conversation timestamps, but we don't do it yet for messages. 

With those timestamps we could in the future get a better understanding of usage but also we could compute average response time from the model and cool things like that.

### DIFF
--- a/src/lib/types/Message.ts
+++ b/src/lib/types/Message.ts
@@ -1,7 +1,9 @@
-export interface Message {
+import type { Timestamps } from "./Timestamps";
+
+export type Message = Partial<Timestamps> & {
 	from: "user" | "assistant";
 	id: ReturnType<typeof crypto.randomUUID>;
 	content: string;
 	webSearchId?: string;
 	score?: -1 | 0 | 1;
-}
+};
--- a/src/routes/conversation/[id]/+server.ts
+++ b/src/routes/conversation/[id]/+server.ts
@@ -59,12 +59,18 @@ export async function POST({ request, fetch, locals, params }) {
 			}
 			return [
 				...conv.messages.slice(0, retryMessageIdx),
-				{ content: newPrompt, from: "user", id: messageId as Message["id"] },
+				{ content: newPrompt, from: "user", id: messageId as Message["id"], updatedAt: new Date() },
 			];
 		}
 		return [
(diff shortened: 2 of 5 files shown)


src/routes/conversation/[id]/+server.ts
```diff
@@ -59,12 +59,18 @@ export async function POST({ request, fetch, locals, params }) {
...
 		}
 		return [
 			...conv.messages,
-			{ content: newPrompt, from: "user", id: (messageId as Message["id"]) || crypto.randomUUID() },
+			{
+				content: newPrompt,
+				from: "user",
+				id: (messageId as Message["id"]) || crypto.randomUUID(),
+				createdAt: new Date(),
+				updatedAt: new Date(),
```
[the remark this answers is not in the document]
Comment (nsarrazin) [reply src/routes/conversation/[id]/+server.ts:72]: When we use retry to generate an answer again it gets updated but not created, we could get rid of it if it's not worth the DB space though.

src/lib/types/Message.ts
```diff
@@ -1,7 +1,9 @@
-export interface Message {
+import type { Timestamps } from "./Timestamps";
+
+export type Message = Partial<Timestamps> & {
```
coyotte508, line 3: We should use the whole `Timestamps` imo.

If you only want `createdAt` or `updatedAt`, you can use `Pick`

coyotte508:
> WDYT @coyotte508?

Sounds good! I guess be careful with sharing convos / hash computation (that the hash is only on non-timestamped data)

src/lib/types/Message.ts
coyotte508: up to you :)

nsarrazin:
> > WDYT @coyotte508?
> 
> Sounds good! I guess be careful with sharing convos / hash computation (that the hash is only on message content)
> 
> it's maybe the case already

Ha good catch, we use [the whole message array.](https://github.com/huggingface/chat-ui/blob/643433996e573c744a569ac6d8580ea9090a4b04/src/routes/search/%5Bid%5D/+server.ts#L26)

I will need to fix this before this can be merged.

nsarrazin:
Should be better now, finally got around to it :smile:

PR #300, opened 2023-06-15
Fix code example preview in Readme

arifszn:
Currently the code example preview has syntax error.
image


nsarrazin:
Good catch! Thank you for contributing :smile:

nsarrazin (approved):

PR #302, opened 2023-06-16
Added Serper API as an alternative web search provider

rogerserper:
Added Serper API as an alternative web search provider.  Serper API is a fast and low-cost Google Search API alternative to SerpAPI.
* approximately 2x lower latencies
* 10x cheaper
* more generous free tier (2,500 vs 100 free queries)

Developers can sign up for a free account at [serper.dev](https://serper.dev/) and obtain an api key.


nsarrazin:
Thank you for the contribution @rogerserper! This is looking pretty good, tested it locally with an account and it´s working great. Merging it!

PR #326, opened 2023-06-28
Add support for Sagemaker endpoint

Mehdi-Aiyad:


Mehdi-Aiyad:
Hey, added support for sagemaker endpoint. In .env.local, if *host* argument (in model endpoints) is set to "aws", user must provide the url to his sagemaker endpoint along with the access_key, secret_access_key and region (in the .env.local file just like HF_token), if not host is provided code behaves the same as before.

nsarrazin:
Hi, thank you for your contribution! I think supporting multiple endpoints is a super valuable feature and I appreciate that you tackled it!

I just want to make sure we do it in a way that's modular so we can scale to supporting even more endpoints in the future, so I'll try to have a look at how we can do this and then we can merge your work into that framework.

rrfaria:
Perhaps, instead of loading an entire library just for formatting UTC, you could consider using:

```js
function getFormattedDate(format) {
  const now = new Date();
  const year = now.getUTCFullYear();
  const month = String(now.getUTCMonth() + 1).padStart(2, '0');
  const day = String(now.getUTCDate()).padStart(2, '0');
  const hours = String(now.getUTCHours()).padStart(2, '0');
  const minutes = String(now.getUTCMinutes()).padStart(2, '0');
  const seconds = String(now.getUTCSeconds()).padStart(2, '0');

  let formattedDate = format
    .replace("yyyy", year)
    .replace("MM", month)
    .replace("DD", day)
    .replace("HH", hours)
    .replace("mm", minutes)
    .replace("ss", seconds);

  return formattedDate;
}

const customFormat = "yyyyMMDD"; 
const formattedDate = getFormattedDate(customFormat);
console.log(formattedDate);
```
I hope it could help

philschmid:
@nsarrazin any timeline on when this would be available?

PR #332, opened 2023-07-08
Factor out HF_API_ROOT to allow different inference endpoints

DayOfThePenguin:
Changed the hard-coding of the inference api endpoint into something configured in `.env` or `.env.local`

This is similar to #326, but is generic enough to allow arbitrary inference endpoints to be swapped in.


nsarrazin:
I see no downsides to this, thanks for contributing!

nsarrazin (approved):

PR #345, opened 2023-07-13
rm open assistant branding

flozi00:


nsarrazin:
Hi! Thanks for opening a PR.

Trying to understand the intention here, I guess you want to remove the branding because it's hardcoded and if you run a different model it doesn't make sense to display this branding?

If so would it make sense to use [the content of the `MODELS` env variable](https://github.com/huggingface/chat-ui/blob/1eff97d9fd47d8c486480d4d9a5208437c519cbb/.env#L28C6-L29) to make the branding dynamic?

flozi00:
Yeah, that makes sense
I just thought the model card is linked in the UI too, so that would be a better reference than the website.
Many open source models have no own website like open assistant project

nsarrazin (approved):
Hey! Thanks again for contributing. I added a few things to your PR but it's good to go now!

PR #352, opened 2023-07-19
Option to disable login on first N messages

nsarrazin:


gary149 (approved):
Didn't try but it seems 👍

PR #359, opened 2023-07-20
Fix ToC + move up space template

osanseviero:
* ToC link is currently broken
* Move up the Spaces option as a fast deploy option


nsarrazin (approved):
Thanks for the fix! Looks good to me

merveenoyan (approved):
Thank you!

julien-c (commented):
nice idea

PR #367, opened 2023-07-25
Clarify that model 'tokens' are not actual tokens

AndreasMadsen:
The `userMessageToken`, `assistantMessageToken`, `messageEndToken`, and `parameters.stop` settings in `MODELS` do not have to be a token. They can be any string.

This also uses `<|endoftext|>` rather than `</s>` as that is the stop token according to the documentation, and it sets that as the stop token in `parameters.stop` to demonstrate the usage of `stop`.


nsarrazin (approved):
Looks good to me, thanks for contributing!

PR #369, opened 2023-07-26
trim and remove stop-suffixes from summary

AndreasMadsen:
The chat generation removes parameters.stop and <|endoftext|> from the generated text. And additionally trims trailing whitespace.

This PR copies that behavior to the summarize functionality, when the summary is produced by a the chat model.


nsarrazin (approved):
Thanks for the PR this is working great!

PR #373, opened 2023-07-28
Attempt to clarify how hosted API ≠ local endpoint

julien-c:


nsarrazin (approved):
Looks good to me! I like the mention of the template :smile:

PR #385, opened 2023-08-03
Leverage model link to modelUrl when informed

airibarne:
In the model link placed below the text input, link URL is being built from `currModel.name` and always heading to HuggingFace. 

To generalize to other models, potentially not placed in HuggingFace, this change leverages this `href` to `modelUrl`, when informed, preserving the old behavior, as already being done in other components.


nsarrazin (approved):
Good catch! Thanks for the contribution.

PR #393, opened 2023-08-09
feat: add support for endpoints requiring client authentication using PKI

cambriancoder:
feat: add support for endpoints requiring client authentication using PKI
* This enables a global mTLS context for client requests using the native fetch
API. It can be used when custom endpoints require client authentication via
PKI.

fix: Dockerfile syntax required for macOS
* Set the Dockerfile syntax explictly to enable new features in Dockerfile.
The default syntax used by Docker Desktop for macOS does not include support
for `--link` in the COPY command for example.

### DIFF
--- a/.env
+++ b/.env
@@ -19,6 +19,13 @@ OPENID_CLIENT_SECRET=
 OPENID_SCOPES="openid profile" # Add "email" for some providers like Google that do not provide preferred_username
 OPENID_PROVIDER_URL=https://huggingface.co # for Google, use https://accounts.google.com
 
+# Parameters to enable a global mTLS context for client fetch requests
+USE_CLIENT_CERTIFICATE=false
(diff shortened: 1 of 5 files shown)


nsarrazin:
Thanks for the PR! It seems super complete, love it. Out of curiosity what's your use case for this ? :eyes: 

I would like to test it locally before approving, do you have any specific infra you use in combination with this feature?

cambriancoder:
We're running the chat-ui in an environment that uses a private certificate authority. Due to security requirements we require all clients/servers in our environment authenticate through mutual TLS. 

We're using the text-generation-inference server as the backend with a reverse proxy (nginx) to handle the TLS termination.

nsarrazin (approved):
Looks good to me. Thanks for contributing!

PR #399, opened 2023-08-14
Support custom system prompts from the user

nsarrazin:
We want to support custom system prompts especially for llama-2.

This PR lets the user set their own prompt and reset to the default if needed. The settings are saved in the database in `settings` under `customPrompts`.

image
image


julien-c:
i like the UI^

Issue #382, opened 2023-08-02
Feature request: use template language for prompts 


AndreasMadsen:
I think the current solution of specific prompt properties (`preprompt`, `userMessageToken`, `messageEndToken`, etc.) is too inflexible. I keep running in to edge cases, which are not supported by such this. A more general solution would be to use a simple template language, such as https://handlebarsjs.com.

In particular falcon requires the following format:

```
{{system message}}

User: {{user message 1}}
Falcon: {{assistant message 1}}
User: {{user message 2}}
Falcon:
```

Note that there is no space after the last `Falcon:`. They say this is important for the tokenizer. However, the history does need the space. Such conditional logical is not currently possible.

I don't think the specific template language is important. But in the case of https://handlebarsjs.com, features like HTML escaping and auto indentation should be disabled. ~I would also propose adding a newline "helper": `{{ln}}` for printing a line break.~ This would make byte-accuate formatting possible.

 ## Example:

The input data could be:
```json
{
  "preprompt": "system message",
  "history": [{
    "user": "user message 1",
    "assistant": "assistant message 1"
  }, {
    "user": "user message 1",
    "assistant": null
  }]
}
```

Then different formats could be implemented with:

**Lamma 2:**

```html
{{#each history}}
<s>[INST] {{#if @first}}<<SYS>>{{{@root.preprompt}}}<</SYS>>

{{/if}}{{user}} [/INST]{{#unless @last}}{{assistant}} </s>{{/unless}}
{{/each}}
```

**Falcon:**

```html
{{{preprompt}}}
{{#each history}}
User: {{user}}
Falcon:{{#unless @last}} {{assistant}}{{/unless}}
{{/each}}
```

**Open Assistant:**

```html
{{{preprompt}}}
{{#each history}}
<|prompter|>{{user}}<|endoftext|><|assistant|>{{#unless @last}}{{assistant}}<|endoftext|>{{/unless}}
{{/each}}
```

**Default:**

The current logic could be implemented with: 

```html
{{{preprompt}}}
{{#each history}}
{{@root.userMessageToken}}{{user}}{{@root.userMessageEndToken}}
{{@root.assistantMessageToken}}{{#unless @last}}{{assistant}}{{@root.assistantMessageEndToken}}{{/unless}}
{{/each}}
```

Although, I would suggest not keeping such implementation in the code, but simple mention it in a migration guide.

nsarrazin:
[...]

PR #400, opened 2023-08-14
Make all prompt templates configurable

AndreasMadsen:
This PR enables all prompts to be configurable using handlebar templates as described in #382.

For backward compatibility the old hardcoded templates are reimplemented as the default templates. The old template parameters such as `preprompt`, `userMessageToken`, `userMessageEndToken`, `assistantMessageToken`, `assistantMessageEndToken` are now considered legacy. They still work as they are exposed as variables to the default
template. However, new prompt configurations should not use these. And it is recommended that the legacy support is eventually removed.

As an example, this is how the default chat prompt template is implemented:

```
{{preprompt}}
{{#each messages}}
{{#ifUser}}{{@root.userMessageToken}}{{content}}{{@root.userMessageEndToken}}{{/ifUser}}
{{#ifAssistant}}{{@root.assistantMessageToken}}{{content}}{{@root.assistantMessageEndToken}}{{/ifAssistant}}
{{/each}}
{{assistantMessageToken}}
```

In addition, this PR fixes an issue where the `model` configuration was used to generate the prompts in WebSearch. However, the `defaultModel` was used to query. This caused issues when the `model` and `defaultModel` uses
different prompt configurations. This has now been changed to always use the `defaultModel`.

Note, when developing this PR, it has been observed that the WebSearch prompts can violate typical model assumptions. For example, a query may be generated as:

```
Assistant: The following context ...
User: content ...
```

```
User: user message 1
User: user message 2
Assistant:
```

Models typically assume the prompts to be User -> Assistant -> User. For best compatability with existing configurations, this issues was not fixed. Instead, the old behavior is maintained with the default templates. This is also the reason why `defaultModel` was chosen as the WebSearch model instead of `model`. As `defaultModel` may allow the WebSearch format, while `model` might not.

This behavior, as well as the overall aritecture of chat-ui, necessitated that the template input maintained the format

```
messages: [{from: 'user' | 'assistant', content: string }]
```

For the template to be able to detect which source a message comes from a `ifUser` and a `ifAssistant` handlebar block-helper was implemented. The original proposed format in #382 was:

```
history: [{ user: string, assistant: string }]
```

However, using such format would require significant changes to the project and would make it impossible to implement the existing websearch templates.

Finally, there may be minor differences in how truncation is implemented. As in some cases, truncation is now applied to the entire prompt, rather than part of the prompt.

Fixes: #382

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,6 +262,24 @@ You can then add the generated information and the `authorization` parameter to 
 
 ```
 
+### Amazon SageMaker
+
+You can also specify your Amazon SageMaker instance as an endpoint for chat-ui. The config goes like this:
+
+```
+"endpoints": [
+    {
+      "host" : "sagemaker",
+      "url": "", // your aws sagemaker url here
+      "accessKey": "",
+      "secretKey" : "",
+      "sessionToken": "", // optional
+      "weight": 1
+    }
+```
+
+You can get the `accessKey` and `secretKey` from your AWS user, under programmatic access.
+
 #### Client Certificate Authentication (mTLS)
 
 Custom endpoints may require client certificate authentication, depending on how you configure them. To enable mTLS between Chat UI and your custom endpoint, you will need to set the `USE_CLIENT_CERTIFICATE` to `true`, and add the `CERT_PATH` and `KEY_PATH` parameters to your `.env.local`. These parameters should point to the location of the certificate and key files on your local machine. The certificate and key files should be in PEM format. The key file can be encrypted with a passphrase, in which case you will also need to add the `CLIENT_KEY_PASSWORD` parameter to your `.env.local`.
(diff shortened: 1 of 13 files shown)


nsarrazin:
:fire: Thanks for doing this, it looks great. I'll have a look at it and I hope we can combine it with #399 !

AndreasMadsen:
@nsarrazin Should be quite trivial to combine. The template already support a `preprompt` parameter.

nsarrazin:
Thanks, this is looking pretty good!

I'm using this template for llama2: 

```
"<s>[INST] <<SYS>>\n{{preprompt}}<</SYS>>\n\n{{#each messages}}{{#ifUser}}{{content}} [/INST] {{/ifUser}}{{#ifAssistant}}{{content}} </s><s>[INST] {{/ifAssistant}}{{/each}}"
```

It seems to work quite well. This is a great PR that's going to make things super flexible moving forward. Really appreciate the work done here!

julien-c:
thanks for this PR @AndreasMadsen it is very cool!

PR #427, opened 2023-09-04
[Websearch] update

mishig25:
### TLDR: update websearch demo

Specifically, there are 2 parts:
1. Improve/fix google query generation, better information retrieval from websites
3. Use embeddings to find relevant texts from sources

### Testing locally
1. Make sure you are using `meta-llama/Llama-2-70b-chat-hf` as your model
2. ~~Run `npm ci` since there is a new dependency to gradio client~~

### Detailed description
1. Request to generate google search query from user query was broken (fixed https://github.com/huggingface/chat-ui/pull/427/commits/0a38beced68264588f98797a04fe3e5dee8ad20b)
2. Update webSearchQueryPromptTemplate (https://github.com/huggingface/chat-ui/pull/427/commits/9145a3aa339f91feda798da65a21d43c94d5d89a and https://github.com/huggingface/chat-ui/pull/427/commits/fd8f654be3835e835651ad66f3b6e2e95b4cba5b and https://github.com/huggingface/chat-ui/pull/427/commits/19ef30bfeb2af498eb8d0221970a9defe3166466). So the new google query generator prompt template is `My question is: {{USER QUESTION}}. Based on the conversation history, give me an appropriate query to answer my question for google search. You should not say more than query. You should not say any words except the query. For the context, today is {{CURRENT DATE}}`. Adding `{{CURRENT DATE}}` has been very useful in asking questions such as: `what happened yesterday?`, which becomes `news September 3, 2023` (overall, the query is same as used https://huggingface.co/spaces/chansung/llama2-with-gradio-chat except current date is added)
3. On google results, get `top storeis` and `organic results` only (discard `knowledge graph` and `answer box`)
4. From the web pages: only get  texts from `<p>` paragraph elements (this seems to remove a lot of "noise" and actually get "useful" text information from websites)
5. Find relevant information on those webpages using ths gradio space https://huggingface.co/spaces/mishig/embeddings-similarity (the space is inspired by https://huggingface.co/spaces/smangrul/PEFT-Docs-QA-Chatbot)
   a. use [intfloat/e5-large-v2](https://huggingface.co/intfloat/e5-large-v2) for embeddings [here](https://huggingface.co/spaces/mishig/embeddings-similarity/blob/main/app.py#L11)
   b. use [hnswlib](https://github.com/nmslib/hnswlib) for finding releavant text chunks [here](https://huggingface.co/spaces/mishig/embeddings-similarity/blob/main/app.py#L71)
    c. and use gradio js client [here](https://github.com/huggingface/chat-ui/blob/2f0b2fa4fc609bab86783fc62ae63fea045bebd4/src/routes/conversation/%5Bid%5D/web-search/+server.ts#L108-L115) (docs [here](https://www.gradio.app/guides/getting-started-with-the-js-client))
6. Inject relevant info into the messages [here](https://github.com/huggingface/chat-ui/pull/427/commits/57e52bfe3bc4eba5af1e0d0c21d62b4a95a61dd6)

### Video cast

https://github.com/huggingface/chat-ui/assets/11827707/98e0ec08-51b5-42ac-9a42-3d5a75a4d536

### Notes

I'd say it is working quite well. ~~However, right now, I'm looking for a potential bug that results in generating a completely unrelated text. When I ask: what happened yesterday, it answered once: Why AWS is a good cloud provider. Looking for a bug whether it is in the prompt, frontend, or server.~~ (Update: https://github.com/huggingface/chat-ui/pull/427/commits/5bfab64cc97af819eb6b66a2ddb70277339c34c9 fixed the error)

### Examples
<details>
  <summary>Open the examples:</summary>
<div>
image
image
image
image
image
image

</div>
</details>

### Next steps
- [ ] get @gary149 feedback and decide


mishig25:
Some tests: 
image
image
image
image

src/routes/conversation/[id]/web-search/+server.ts
```diff
@@ -73,45 +61,60 @@ export async function GET({ params, locals, url }) {
...
-				webSearch.summary = await summarizeWeb(text, webSearch.searchQuery, model);
-				appendUpdate("Injecting summary", [JSON.stringify(webSearch.summary)]);
+				appendUpdate("Extracing relevant information");
+				const topKClosestParagraphs = 8;
+				const requestBody = {
+					paragraphs: paragraphChunks,
+					query: prompt,
+					top_k: topKClosestParagraphs,
+				};
+				const res = await fetch("https://mishig-embeddings-similarity.hf.space/", {
```
Comment (nsarrazin) [inline src/routes/conversation/[id]/web-search/+server.ts:104]: Would it be possible to make this an environment variable so users can have their own embeddings API? :eyes:

Comment (mishig25) [reply src/routes/conversation/[id]/web-search/+server.ts:104]: totally! I'm gonna move this one to https://huggingface.co/huggingchat org

Comment (julien-c) [reply src/routes/conversation/[id]/web-search/+server.ts:104]: so what i would suggest is that it should:
- run by default on server-side JS (transformers.js for instance) embedding + pure JS similarity search
- optionally, use your Space as a backend API (to provide better performance? though i'm pretty sure the local perf would be the same)

src/routes/conversation/[id]/web-search/+server.ts
```diff
@@ -73,45 +61,60 @@ export async function GET({ params, locals, url }) {
...
 					}
-					if (!text) throw new Error("No text found on the first 3 results");
 				} else {
 					throw new Error("No results found for this search query");
 				}
 
-				appendUpdate("Creating summary");
-				webSearch.summary = await summarizeWeb(text, webSearch.searchQuery, model);
-				appendUpdate("Injecting summary", [JSON.stringify(webSearch.summary)]);
+				appendUpdate("Extracing relevant information");
```
Comment (julien-c) [inline src/routes/conversation/[id]/web-search/+server.ts:97]:
```suggestion
				appendUpdate("Extracting relevant information");
```

src/routes/conversation/[id]/web-search/+server.ts
```diff
@@ -9,21 +8,12 @@ import { z } from "zod";
 import type { WebSearch } from "$lib/types/WebSearch";
 import { generateQuery } from "$lib/server/websearch/generateQuery";
 import { parseWeb } from "$lib/server/websearch/parseWeb";
-import { summarizeWeb } from "$lib/server/websearch/summarizeWeb";
+import { chunk } from "$lib/utils/chunk.js";
```
Comment (julien-c) [inline src/routes/conversation/[id]/web-search/+server.ts:11]:
```suggestion
import { chunk } from "$lib/utils/chunk";
```
no?

julien-c (commented):
looks cool!

in my local tests it does seem to work a bit better than the current search (and it's faster as it does hit the LLM with summarization of all web pages?)

(i suspect it's due a lot to `only get texts from <p> paragraph elements (this seems to remove a lot of "noise" and actually get "useful" text information from websites)`)

bilalazhar72:
will it make the search function of the hugging face better as well right now its very broken

mishig25:
@bilalazhar72  yes indeed it is gonna be better with this PR

mishig25:
### Updates

1. Creating embeddings and calculating similarity is working great on server-side JS using transfomrers.js
2. Show sources of web searches

image

src/lib/server/websearch/sentenceSimilarity.ts
```diff
@@ -0,0 +1,39 @@
+import type { Tensor } from "@xenova/transformers";
+import { pipeline, dot } from "@xenova/transformers";
+
+// see here: https://github.com/nmslib/hnswlib/blob/359b2ba87358224963986f709e593d799064ace6/README.md?plain=1#L34
+function innerProduct(tensor1: Tensor, tensor2: Tensor) {
+	return 1.0 - dot(tensor1.data, tensor2.data);
+}
+
+const extractor = await pipeline("feature-extraction", "Xenova/e5-small-v2");
```
mishig25, line 9: I can make the embedding model env variable. However, there is a specific preprocessing step https://github.com/huggingface/chat-ui/blob/a47c6f29ae752e1e2727bfb367aa98559c6cb1d2/src/lib/server/websearch/sentenceSimilarity.ts#L16-L18

nsarrazin:
Love the sources :fire: I'll test this locally!

src/routes/conversation/[id]/web-search/+server.ts
Comment (mishig25) [reply src/routes/conversation/[id]/web-search/+server.ts]: handled in https://github.com/huggingface/chat-ui/pull/427#issuecomment-1717070246

src/lib/server/websearch/generateQuery.ts
```diff
@@ -1,12 +1,25 @@
 import type { Message } from "$lib/types/Message";
+import { format } from "date-fns";
 import { generateFromDefaultEndpoint } from "../generateFromDefaultEndpoint";
 import { defaultModel } from "../models";
 
 export async function generateQuery(messages: Message[]) {
-	const promptSearchQuery = defaultModel.webSearchQueryPromptRender({ messages });
+	const currentDate = format(new Date(), "MMMM d, yyyy");
+	const lastMessage = messages[messages.length - 1];
```
nsarrazin, line 8: Because we use only the last message, you can't ask follow-up questions right?

Like I tried 
image and it had no context of the previous question.

Not that big of a deal I think, since when we add the agents flow it will be solved, but just thought I'd bring it up.

src/lib/server/models.ts
```diff
@@ -71,28 +67,14 @@ const modelsRaw = z
...
-						"{{userMessageEndToken}}" +
-						"{{assistantMessageToken}}Summary: "
-				),
 			webSearchQueryPromptTemplate: z
 				.string()
 				.default(
 					"{{userMessageToken}}" +
-						"The following messages were written by a user, trying to answer a question." +
+						'My question is: "{{message.content}}". ' +
+						"Based on the conversation history, give me an appropriate query to answer my question for google search. You should not say more than query. You should not say any words except the query. For the context, today is {{currentDate}}" +
```
nsarrazin, line 75: Could be interesting to give both the current date and the training data cutoff date, I think I've seen some other chat apps use both, but not a needed change.

src/lib/buildPrompt.ts
```diff
@@ -39,13 +40,16 @@ export async function buildPrompt({
 
 		if (!conversation) throw new Error("Conversation not found");
 
-		if (webSearch.summary) {
+		if (webSearch.context) {
+			const messagesWithoutLastUsrMsg = messages.slice(0, -1);
```
nsarrazin, line 44: Much better like this, this makes a lot of sense!

src/routes/conversation/[id]/web-search/+server.ts
```diff
@@ -73,45 +63,63 @@ export async function GET({ params, locals, url }) {
...
+				if (webSearch.results.length > 0) {
+					appendUpdate("Browsing results");
+					const promises = webSearch.results.map(async (result) => {
+						const { link } = result;
+						let text = "";
 						try {
-							text = await parseWeb(searchUrl);
-							if (!text) throw new Error("text of the webpage is null");
+							text = await parseWeb(link);
+							appendUpdate(`Browsing webpage: ${link}`);
```
Comment (nsarrazin) [inline src/routes/conversation/[id]/web-search/+server.ts:86]:
```suggestion
							appendUpdate("Browsing webpage", [link]);
```

Fairly small change but it puts the url as an argument below 

image

src/routes/conversation/[id]/web-search/+server.ts
```diff
@@ -73,45 +63,63 @@ export async function GET({ params, locals, url }) {
...
-					text = webSearch.knowledgeGraph;
-					appendUpdate("Found a Google knowledge page");
-				} else if (webSearch.results.length > 0) {
-					let tries = 0;
-
-					while (!text && tries < 3) {
-						const searchUrl = webSearch.results[tries];
-						appendUpdate("Browsing result", [JSON.stringify(searchUrl)]);
+				webSearch.results = webSearch.results
+					.filter(({ link }) => !link.includes("youtube.com")) // filter out youtube links
```
Comment (nsarrazin) [inline src/routes/conversation/[id]/web-search/+server.ts:75]: would love to know if there's a banlist somewhere for websites we know we can't easily parse

nsarrazin (approved):
Overall this is clearly much better! I tried it with Serper and it worked great. I also think we should switch to serper for prod tbh.

I like the sources UI, plus it's much faster and more accurate. I left a few minor comments but really nothing blocking, just notes for the future.

Minor nit is that the update collapse menu is getting pretty big and  I feel like some updates could be combined/removed imo.

 I think it makes sense to keep the ones that contain extra info (like which query is used, which webpages are visited) but the other ones could be dropped/merged into fewer updates imo.

image

src/lib/server/websearch/sentenceSimilarity.ts
julien-c: really cool cc @xenova @osanseviero ^

src/lib/server/websearch/sentenceSimilarity.ts
```diff
@@ -0,0 +1,39 @@
+import type { Tensor } from "@xenova/transformers";
+import { pipeline, dot } from "@xenova/transformers";
+
+// see here: https://github.com/nmslib/hnswlib/blob/359b2ba87358224963986f709e593d799064ace6/README.md?plain=1#L34
+function innerProduct(tensor1: Tensor, tensor2: Tensor) {
+	return 1.0 - dot(tensor1.data, tensor2.data);
```
julien-c, line 6: ah nice dot product is already implemented in there 🎉

src/lib/server/websearch/sentenceSimilarity.ts
xenova: Haha yes indeed! 🔥 

PS @mishig25, you can experiment with using the unquantized versions of the models too by passing `{quantized: false}` as a third parameter to the pipeline creation function.

i.e., 
```js
const extractor = await pipeline("feature-extraction", "Xenova/e5-small-v2", { quantized: false });
```

The unquantized models are ~4x larger, however, so just take note of that.


> using [Xenova/e5-small-v2](https://huggingface.co/Xenova/e5-small-v2) which provided better results than the popular [Xenova/all-MiniLM-L6-v2](https://huggingface.co/Xenova/all-MiniLM-L6-v2)

The [MTEB leaderboard](https://huggingface.co/spaces/mteb/leaderboard) corroborates this observation.

xenova:
For the transformers.js integration, I would also recommend the following:
1. Making copies of the models that will be used, and freezing them. For example, this is what [Supabase/gte-small](https://huggingface.co/Supabase/gte-small) does.
2. Host the WASM files on Hugging Face instead of jsdelivr. This isn't a massive issue, but it's always good to know if can never break in the future :) You can then set the path to the WASM files with (see [docs](https://huggingface.co/docs/transformers.js/custom_usage)):
    ```js
    import { env } from '@xenova/transformers';
    
    // Set location of .wasm files. Defaults to use a CDN.
    env.backends.onnx.wasm.wasmPaths = '/path/to/files/';
    ```
3. Freeze the transformers.js version. Of course, I don't intend to introduce any breaking changes 😅 ... but freezing this makes sure the correct ONNX + WASM versions are loaded and also prevents any issues that occur later down the line. The current version is v2.6.0, and perhaps you can wait until I release v2.6.1 which has a few fixes for minor issues relating to minification (which shouldn't affect embedding models). We can always upgrade later!

mishig25:
@xenova based on this codebase, do you have suggestion for the path? And I guess I should put that path in gitignore?

xenova:
> @xenova based on this codebase, do you have suggestion for the path? And I guess I should put that path in gitignore?

These can also be HTTP(s) links, so we could host them on the hub if needed. Alternatively, if you wish to serve these locally, they can be copied from `node_modules/@xenova/transformers/dist` into whatever the public folder is.

src/routes/conversation/[id]/web-search/+server.ts
Comment (gary149) [reply src/routes/conversation/[id]/web-search/+server.ts]: yes nice catch

src/lib/server/websearch/generateQuery.ts
gary149: yes, mishig addressed this 🔥

src/lib/server/models.ts
gary149: ok but let's do it in the system prompt!

src/routes/conversation/[id]/web-search/+server.ts
Comment (mishig25) [reply src/routes/conversation/[id]/web-search/+server.ts]: I've removed those messages altogether https://github.com/huggingface/chat-ui/pull/427/commits/dd155791604bcce58749cab32086164cf7d43e0e as it is very noisy and we show sources at the end. wdyt?

src/lib/server/websearch/sentenceSimilarity.ts
mishig25: yes, dot product - the operation of the brain

src/routes/conversation/[id]/web-search/+server.ts
```diff
@@ -83,10 +83,8 @@ export async function GET({ params, locals, url }) {
 						let text = "";
 						try {
 							text = await parseWeb(link);
-							appendUpdate(`Browsing webpage: ${link}`);
```
Comment (gary149) [inline src/routes/conversation/[id]/web-search/+server.ts]: I think this was fine (+ @nsarrazin comment)

Comment (mishig25) [reply src/routes/conversation/[id]/web-search/+server.ts]: > Minor nit is that the update collapse menu is getting pretty big and I feel like some updates could be combined/removed imo.

from @nsarrazin [here](https://github.com/huggingface/chat-ui/pull/427#pullrequestreview-1624289710)

mishig25:
@xenova in this PR, tfjs is running on server-side, not frontend. Therefore, is `import { env } from '@xenova/transformers';` needed?

gary149 (approved):

xenova:
> @xenova in this PR, tfjs is running on server-side, not frontend. Therefore, is `import { env } from '@xenova/transformers';` needed?

I guess it depends if you want to modify the default configuration at all (e.g., updating the path to the WASM files). It also depends how much extra configuration you want users to face. However, as you say, it's running server-side, so it doesn't actually use the WASM files - it uses their onnxruntime-node backend.

TLDR: Probably not necessary! Check [these options](https://huggingface.co/docs/transformers.js/api/env#envenv) just in case you want some custom behaviour.

pixelpandacreative:
“ through something hacky like selenium, Google won't like it (and they'll miss ad revenue).“ 
@nsarrazin of whom was this quoting on HF?

I'm trying not to implode with cringe and laughter fighting to make sense of such a ridiculous naivety and ill-logic.

Google owns searching the internet? We must not do anything Google doesn't like. Their feelings could get huwwrrrrt. We are not responsible for the eggs Google's business has restricted its success to. We are not responsible for prioritizing the Google-gets-a-check-for any web searchy things any creature partakes in. Cuz, duh, pay Google. I mean just like, they did a thing so pay up humanity! 🤦🏻‍♂️

xenova:
@pixelpandacreative I believe [this](https://huggingface.co/spaces/huggingchat/chat-ui/discussions/254#650217749a612aa33af48568) is the thread you are referring to? Your questions will be answered there :)

ShadowDawg:
Just curious, how does web search for this really work? I went through langchain's web search tutorial and according to it, the web search api usually returns html and we have to manually filter it by relevant tags (span, p, etc) to get the exact text information. And of course, this doesn't really seem to be applicable to every website. So how does huggingchat or perplexity manage to extract the necesarry information from any web page?

mishig25:
the overall process is described in https://github.com/huggingface/chat-ui?tab=readme-ov-file#web-search

you can see the code for the details: https://github.com/huggingface/chat-ui/tree/main/src/lib/server/websearch

PR #443, opened 2023-09-17
Add support for OpenAI API

julien-blanchon:
This PR add the support for the OpenAI API

This implementation add a new route `/openai/` that wrap the OpenAI API in a HuggingFace Inferencer fashion.
This route return a StreamingResponse if the `stream` parameter is set to `true`.
I did try to mimic the HuggingFace Inferencer as much as possible.

To call internal `/openai/` route in `generateFromDefaultEndpoint` I added the sveltekit fetch in argument.
Maybe it's not the best way to do it ...

I did a little update of the `README.md` to add some explanations about the OpenAI API.

https://github.com/huggingface/chat-ui/assets/11278197/9c6181ad-9a48-4ec5-bf90-143b5af30e23

---

TODO: 
- [x] Fix the `max_new_tokens` issue
- [ ] Improve the README.md documentation
- [x] Improve the /openai route to be more generic

### DIFF
--- a/.env
+++ b/.env
@@ -92,3 +92,6 @@ PUBLIC_APP_DISCLAIMER=#set to 1 to show a disclaimer on login page
 # PUBLIC_APP_COLOR=yellow
 # PUBLIC_APP_DATA_SHARING=1
 # PUBLIC_APP_DISCLAIMER=1
+
+CHATGPT_API_KEY=#sk-.... your chatgpt api key here
+CHATGPT_MODEL=#gpt-4-32k	the model to use for chatgpt: (https://platform.openai.com/docs/models)
--- a/.env
+++ b/.env
@@ -92,6 +92,3 @@ PUBLIC_APP_DISCLAIMER=#set to 1 to show a disclaimer on login page
 # PUBLIC_APP_COLOR=yellow
 # PUBLIC_APP_DATA_SHARING=1
 # PUBLIC_APP_DISCLAIMER=1
-
-CHATGPT_API_KEY=#sk-.... your chatgpt api key here
-CHATGPT_MODEL=#gpt-4-32k	the model to use for chatgpt: (https://platform.openai.com/docs/models)
--- a/README.md
+++ b/README.md
@@ -162,11 +162,11 @@ MODELS=`[
 
 You can change things like the parameters, or customize the preprompt to better suit your needs. You can also add more models by adding more objects to the array, with different preprompts for example.
 
-#### Custom prompt templates:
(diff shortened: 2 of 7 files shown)


julien-blanchon:
To avoid adding dependencies I didn't add the openai package but nice idea could be to add the langchain.js package and to adapt my script for any hosted service that langchain support and will support in the future.

What is your thoughts about this ?

Maemol:
I think it's a good idea to support langchain.js. I was working on implementing a similar solution with the @azure/openai package but it would make sense to use langchain instead. I'm going to look at your implementation tonight. Thank you!

julien-blanchon:
**About langchain.js:**

We should wait for feedback from the HuggingFace team before using langchain.js. Langchain has features like the SerpAPI web search, but HuggingFace is also making something similar called [Agents.js](https://huggingface.co/blog/agents-js). We don't want any clashes. So, it's best to add langchain.js if we're sure it's beneficial. I think it's a good idea to hear from an expert about this.

---- 

**Updates on the Current PR:**

I found a bug. When we change the max_new_tokens, there's an error. It seems like something's off with the /openai endpoint. I've marked this for future fixing.

Also, I want to make the /openai route more flexible. We might use it again with other tools.

About the token format: I've suggested a layout to make our work with the OpenAI API easier:
```json
  "userMessageToken": "",
  "assistantMessageToken": "",
  "userMessageEndToken": "</s>",
  "assistantMessageEndToken": "</s>",
  "preprompt": "You are a helpful assistant...</s>",
```
Is this format okay for other tools we might use later?

helterskelterr:
really good change, would this also work with the new gpt-3.5 instruct model that just got released

julien-blanchon:
> gpt-3.5 instruct

Yes of course every OpenAI model work (include `gpt-3.5-turbo-instruct`), you just need to change the model name in the .env config. The only OpenAI features that don't work is [ `function`](https://openai.com/blog/function-calling-and-other-api-updates)

chenhunghan:
Looks like we have similar ideas :) https://github.com/huggingface/chat-ui/pull/452

julien-blanchon:
> Looks like we have similar ideas :) #452

Nice ! I like

julien-blanchon:
> Looks like we have similar ideas :) #452

Nice 🤗 

I like your PR, especially the `openAIChatToTextGenerationStream`, it's a the same mecanism but with a quite better style.
I think I will close this PR and we will continue to work there.

See you in https://github.com/huggingface/chat-ui/pull/452 🙂

Issue #421, opened 2023-08-30
Web search results disappear after message completion


nsarrazin:
The search results disappear after the assistant message is fully generated on https://huggingface.co/chat/.

This is probably due to how web search updates are passed. The switch from the streaming of updates (for the latest streaming message) to a stored message with web search updates in it, i must be going wrong somehow.

nsarrazin:
Should be fixed with #444 !

PR #444, opened 2023-09-18
Backend refactor for streaming endpoints

nsarrazin:
## Why
This refactor was on the critical path for agents. Our old backend basically called `textGenerationStream` from the front-end, which meant that we expected only to recieive stream of tokens from the backend. 

This is the reason why websearch updates (which are json, not just plaintext tokens) had to be streamed from another endpoint, then the websearch ID was passed around on the front-end between the two endpoints. It was clunky but back then I wanted to avoid rewriting everything.

We did [some workarounds](https://github.com/huggingface/chat-ui/blob/8e300b6175549f09965634f573ceaad2f68d3e1f/src/routes/conversation/%5Bid%5D/%2Bpage.svelte#L56C4-L62) to inject things into `textGenerationStream` options (bypassing typescript in the process) but the problem was that we could only return a stream of tokens. 

Now that we are adding agents, we need to stream not only the tokens but also arbitrary updates from the scratchpad, the websearch, any tools being used etc. It was not practical to do all this in separate endpoints so I rewrote the backend so that the main endpoint can just stream any updates.

## How
I moved the call to `textGenerationStream` from the front-end to the back-end and rerouted everything accordingly. 

The POST endpoint now streams JSON updates that get parsed in the front-end. They're [defined as a discriminated union](https://github.com/huggingface/chat-ui/blob/ef7d7fd1c133fefa82e3cea0c80f8c3b625cf3d9/src/lib/types/MessageUpdate.ts) so we can do different things based on update type. If it's a `webSearch` update we append it to the websearch updates, if it's a `stream` we append the token inside to the latest message, etc. This is **trivial to expand for agents** now.

It also means the **`web-search` endpoint is gone**, everything happens through the main endpoint now! 🎉 No more passing around of websearch id, past messages, auth check, etc. The whole thing happens in the conversation.  

## Todo:
- [x] Fix legacy type errors from old websearch
- [x] Websearch doesn't save nicely to DB, fix serialization
- [x] Loading indicator shows even when it shouldn't


nsarrazin:
Should fix #421 as well

gary149:
Let's try to fix this too?

https://github.com/huggingface/chat-ui/assets/3841370/6d143adb-2b15-4b5e-9ff7-a11afc6aff26

src/lib/buildPrompt.ts
```diff
@@ -14,44 +12,31 @@ interface buildPromptOptions {
...
-		}
+	if (webSearch && webSearch.context) {
+		const messagesWithoutLastUsrMsg = messages.slice(0, -1);
+		const lastUserMsg = messages.slice(-1)[0];
+		const currentDate = format(new Date(), "MMMM d, yyyy");
+		messages = [
+			...messagesWithoutLastUsrMsg,
+			{
+				from: "user",
+				content: `Please answer my question "${lastUserMsg.content}" using the supplied context below (paragrpahs from various websites). For the context, today is ${currentDate}:
```
julien-blanchon, line 33: typo: `paragraphs`

nsarrazin:
> Let's try to fix this too?
> Screen.Recording.2023-09-19.at.10.11.27.mov

Ok should be fixed, now if you switch conversation it saves the message

[Screencast from 09-20-2023 01:56:37 PM.webm](https://github.com/huggingface/chat-ui/assets/25119303/9055f063-1a7d-4702-a979-beb13776e54c)

@gary149

nsarrazin:
cc @mishig25 I moved the websearch from a separate endpoint to [a function](https://github.com/huggingface/chat-ui/pull/444/files#diff-32cb113e5edc7224bb9c60ac48202bbf5a847207c881ff8174be89ddb756b390) that gets called within the main `POST` endpoint. 

I didn't change any of the websearch functionality in this PR so it should be functionally identical, but just wanted to give you a heads up

PR #445, opened 2023-09-18
fix JSON.parse for summerize

jackielii:
When serving with TGI, summerize calls this function and it errors with `SyntaxError: Unexpected token d in JSON at position 0`

This PR fixes the problem and keeps existing behaviour.

Example `result` value:

```
data:{"token":{"id":29871,"text":" ","logprob":0.0,"special":false},"generated_text":null,"details":null}

data:{"token":{"id":15043,"text":" Hello","logprob":0.0,"special":false},"generated_text":null,"details":null}

data:{"token":{"id":29892,"text":",","logprob":0.0,"special":false},"generated_text":null,"details":null}

data:{"token":{"id":920,"text":" how","logprob":0.0,"special":false},"generated_text":null,"details":null}

data:{"token":{"id":526,"text":" are","logprob":0.0,"special":false},"generated_text":null,"details":null}

data:{"token":{"id":366,"text":" you","logprob":0.0,"special":false},"generated_text":null,"details":null}

data:{"token":{"id":29973,"text":"?","logprob":0.0,"special":false},"generated_text":null,"details":null}

data:{"token":{"id":2,"text":"</s>","logprob":0.0,"special":true},"generated_text":" Hello, how are you?","details":null}
```

As you can see, it's not a valid json


src/lib/server/generateFromDefaultEndpoint.ts
```diff
@@ -87,7 +87,12 @@ export async function generateFromDefaultEndpoint(
 	// Close the reader when done
 	reader.releaseLock();
 
-	const results = await JSON.parse(result);
+	let results;                                                                                                                                                                                                                                                                                            
+	if (result.startsWith("data:")) {                                                                                                                                                                                                                                                                       
+		results = [JSON.parse(result.split("data:").pop())]
```
julien-blanchon, line 92: A `await JSON.parse` may be needed here ?

jackielii: you mean add `await` in front of `JSON.parse`? I don't think it's needed JSON.parse is not async, is it?

julien-blanchon: Yes you're right, so the original results = await JSON.parse(result) might be useless

nsarrazin (approved):
Thanks for the contribution! Looks good, I'll merge it now.

PR #451, opened 2023-09-19
Improve error message when the .env MODELS is not well formatted

julien-blanchon:
## Summary
Improved error messaging for malformed MODELS format.

## Description

When using our application, I've often encountered a non-descriptive error message when the custom MODELS is not formatted correctly. Even a simple mistake like an extra comma at the end can trigger this error.

This PR aims to provide clearer feedback to the developer by replacing the default error message with a more meaningful one. By doing so, we can ensure faster debugging and a better development experience.

## Before

Screenshot 2023-09-19 at 23 27 27

## After

Screenshot 2023-09-19 at 23 26 55

### DIFF
--- a/src/lib/server/models.ts
+++ b/src/lib/server/models.ts
@@ -2,6 +2,7 @@ import { HF_ACCESS_TOKEN, MODELS, OLD_MODELS } from "$env/static/private";
 import type { ChatTemplateInput, WebSearchQueryTemplateInput } from "$lib/types/Template";
 import { compileTemplate } from "$lib/utils/template";
 import { z } from "zod";
+import { error } from '@sveltejs/kit';
 
 type Optional<T, K extends keyof T> = Pick<Partial<T>, K> & Omit<T, K>;
 
@@ -37,6 +38,14 @@ const combinedEndpoint = endpoint.transform((data) => {
 	}
 });
 
+let parsedModels;
+try {
+	parsedModels = JSON.parse(MODELS);
+} catch (e) {
+	throw error(500, {
+		message: "There's a formatting error in your MODELS .env file. Please check it (missing/extra commas, etc.).",
+	});
+}
 const modelsRaw = z
 	.array(
 		z.object({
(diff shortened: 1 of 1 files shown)


src/lib/server/models.ts
```diff
@@ -37,6 +38,14 @@ const combinedEndpoint = endpoint.transform((data) => {
 	}
 });
 
+let parsedModels;
+try {
+	parsedModels = JSON.parse(MODELS);
+} catch (e) {
+	throw error(500, {
+		message: "There's a formatting error in your MODELS .env file. Please check it (missing/extra commas, etc.).",
```
julien-c, line 46: would be nice to be able to surface the line or property around which the JSON parser fails, as it's quite hard to find what's broken otherwise

julien-blanchon: Yes the only way is to use another kind of JSON.parse for this one. Maybe https://github.com/zaach/jsonlint

julien-blanchon: Resolve with jsonlint

nsarrazin:
Hey, thanks for the contribution. I like the idea of showing that the issue is with the `MODELS` var, as it can be confusing, however I'm not a fan of showing part of the config in the error message. 

If we mess up the `.env` in prod, we could end up leaking private parts of our config to everyone, stuff like endpoints, auth keys, etc. I'd either only show it for `dev` mode or just show the line number + error without showing the actual content if that's possible.

Do you agree @julien-c ?

nsarrazin:
> > If we mess up the `.env` in prod, we could end up leaking private parts of our config to everyone, stuff like endpoints, auth keys, etc. I'd either only show it for `dev` mode or just show the line number + error without showing the actual content if that's possible.
> 
> Yes you're right ! Maybe just display this in the server console and write a generic message for the client side without more info

Yeah I think that would be a great solution!

nsarrazin:
Not sure how but this PR broke in production :sweat_smile: @julien-blanchon I'll investigate and try to merge again!

PR #453, opened 2023-09-20
Add a message wide copy button

julien-blanchon:
Add a 'Copy' button next to the 'Like/Dislike' buttons, allowing users to easily copy raw messages. This feature is particularly beneficial for complex Markdown/LaTeX messages that might be challenging to copy directly from the browser.

Screenshot 2023-09-20 at 15 35 05

(Note that oasst-sft-4-pythia-12b-epoch-3.5 is quite stupid in this example)


julien-blanchon:
We may need to change CopyToClipBoardBtn to fix the padding or to introduce tailwindMerge. I think if it's worth it for a such little features

src/lib/components/chat/ChatMessage.svelte
```diff
@@ -195,6 +196,10 @@
 				>
 					<CarbonThumbsDown class="h-[1.14em] w-[1.14em]" />
 				</button>
+				<CopyToClipBoardBtn
```
gary149, line 199: Good idea! There are some adjustments to be made to make it look exactly like the other buttons. I can do it in your PR if you want :)
image

gary149: > We may need to change CopyToClipBoardBtn to fix the padding or to introduce tailwindMerge. I don't know if it's worth it for a such little features

Have you tried overwriting the classes you pass with the [important modifier](https://tailwindcss.com/docs/configuration#important-modifier)? I think it could work.

julien-blanchon: Yes this sould definively work. I didn't know this tailwind feature. Sure if you want to edid go on ;). Otherwise I may do it later in the evening (3 hours max I think)

gary149: Zero pressure 👍 as you want

julien-blanchon:
Done, I move the button before the like and dislike button. I can't remember where but I've see this layout on another huggingface app (copy, like, dislike)

gary149 (approved):
Thanks! I made a few small layout changes (added a gap between the votes and the copy button, also put it at the end makes more sense imo) also to make it stay when you tap on mobile.

image

julien-blanchon:
Yep, it kinda of make sense to add a small gap between the votes and the copy button. It's okay for me, you can merge

PR #458, opened 2023-09-21
Use `@postlight/parser` for parsing html text

mishig25:
Comparing HTML text parsers:
* jsdom (parse only `<p>` elements) (current approach)
* https://github.com/mozilla/readability
* https://github.com/postlight/parser

From the tests below, that you can see https://github.com/postlight/parser is producing slightly better results. Please let me know what you think of these prompt tests on different parsers

Find all the data of prompts and generations: <a href="https://docs.google.com/spreadsheets/d/1pNoDa00d-yoOQt5SwVczDeOO8atsbY3YTK002FwUIPU/edit#gid=719616982"> Google Sheet image </a>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,7 @@
 			"dependencies": {
 				"@huggingface/hub": "^0.5.1",
 				"@huggingface/inference": "^2.2.0",
+				"@postlight/parser": "^2.2.3",
 				"@xenova/transformers": "^2.6.0",
 				"autoprefixer": "^10.4.14",
 				"aws4fetch": "^1.0.17",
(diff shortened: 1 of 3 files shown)


gururise:
Very interesting. It does appear that the postlight/parser did give better results for the "**Who won the US Open?**" question.

Might be beneficial to test on a few more questions, specifically related to product searches, as that is also quite a common use case.

mishig25:
@gururise if you have some good questions to test, please comment them below 👍

gururise:
> @gururise if you have some good questions to test, please comment them below 👍

Sorry, here are a few questions that I came up with over the weekend:

1. What is the print width of the Prestige A4 DTF printer?  (answer: 7.87")
2. Does the DigiRIP RIP Software only run on windows? (answer: yes, windows only)
3. How many print heads does the Prestige R2 DTF printer have? (answer: 2 Epson L805 print heads)
4. What type of printhead does the Epson F2270 DTG printer use? (answer: PrecisionCore MicroTFP printhead with Nozzle Verification Technology)

These are all equipment that are new and won't be in any LLM's knowledge. Hope this helps!

mishig25:
Thanks so much. Will test on them

nsarrazin:
@mishig25 just curious, do you think we should switch parsers in the end? (And does this PR overlap with the changes in #502 ?)

mishig25:
Indeed. this PR is superceded by #502

PR #442, opened 2023-09-16
Add PDF to Markdown feature

julien-blanchon:
This pull request introduces the PDF to Markdown conversion. This enhancement aims to provide users with a seamless experience when discussing and referencing content from PDFs, especially in academic and research contexts.

PR #462, opened 2023-09-25
Agents 

nsarrazin:
Still a WIP

Requires https://github.com/huggingface/huggingface.js/pull/249 as a linked package in local dev mode to work

image

Roadblocks:
- [x] Need to hook-up agents updates to websearch collapsable modal
- [ ] Need to check the streaming tokens to extract the final answer as it comes. (right now we only parse the JSON when it's done streaming but we would like to stream as it comes in)
- [ ] SDXL pipeline requires base + refiner. We don't currently have an endpoint for the refiner, but it's being worked on internally. base model works fine though.
- [ ] need to check prompts, final answer doesn't reference files for some reason.

### DIFF
--- a/.env
+++ b/.env
@@ -87,6 +87,7 @@ PUBLIC_APP_COLOR=blue # can be any of tailwind colors: https://tailwindcss.com/d
 PUBLIC_APP_DATA_SHARING=#set to 1 to enable options & text regarding data sharing
 PUBLIC_APP_DISCLAIMER=#set to 1 to show a disclaimer on login page
 
+TOOLS = []
 # PUBLIC_APP_NAME=HuggingChat
 # PUBLIC_APP_ASSETS=huggingchat
 # PUBLIC_APP_COLOR=yellow
--- a/README.md
+++ b/README.md
@@ -162,11 +162,11 @@ MODELS=`[
 
 You can change things like the parameters, or customize the preprompt to better suit your needs. You can also add more models by adding more objects to the array, with different preprompts for example.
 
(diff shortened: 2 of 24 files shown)


gururise:
WOW! This is amazing. I hope the Agents will be modular/extensible. 
Would love to be able to use Agents with OpenAI or other LLM's (see PR #452)

nsarrazin:
Tests are failing because we need the new agents release for them to pass.

Road to merging:

1. Merge https://github.com/huggingface/huggingface.js/pull/249
2. Do a release on huggingface.js
3. Add the new release to the deps of chat-ui
4. Merge this PR 
5. Deploy to prod

Notes:
* We still don't have a pipeline for SDXL that's base + refiner, or optimized for speed, not sure if we want it before merging
* We should let the team know before doing comms so we can ensure the sdxl endpoint doesn't die

Let me know what you think @gary149

nsarrazin:
Closing this for now, will come back to this later :grin:

hungryalgo:
This is such an amazing and huge feature, when are you @nsarrazin planning to come back to this?  
I am guess that you are waiting for the SDXL base + refiner endpoint to be ready?

Will you integrate some of the file drop UI from #442?

julien-blanchon:
@hungryalgo, I think the dropzone UI was already integrated with https://github.com/huggingface/chat-ui/pull/531 and live here https://github.com/huggingface/chat-ui/blob/3cbea3450ddaf06d09a7790c9068bc08340dd425/src/lib/components/chat/FileDropzone.svelte

PR #470, opened 2023-10-03
Store preprompt in conversation (#422)

### LINKED ISSUE #422: System prompts should be included in shared chats as well
Given system prompts are configured at user level, they should be stored when saving them for reproducibility

nsarrazin:
This PR fixes #422. We now store preprompts in the conversation. 

The custom preprompt is fetched from settings and passed in the POST request when the user creates a new conversation. It then gets used in the prompt building. If it's undefined we default to the model defined preprompt.

It also gets passed when sharing a conversation and displayed at the top of the conversation.

image

### DIFF
--- a/src/lib/components/chat/ChatMessages.svelte
+++ b/src/lib/components/chat/ChatMessages.svelte
@@ -44,7 +44,10 @@
 	<div class="mx-auto flex h-full max-w-3xl flex-col gap-6 px-5 pt-6 sm:gap-8 xl:max-w-4xl">
 		{#each messages as message, i}
 			{#if i === 0 && preprompt}
-				<div class="text-sm text-gray-500">Custom system prompt: {preprompt}</div>
+				<div class="mx-auto text-sm text-gray-500">
+					Custom system prompt:
+					<span class="font-semibold">{preprompt}</span>
+				</div>
 			{/if}
 			<ChatMessage
 				loading={loading && i === messages.length - 1}
--- a/src/lib/components/chat/ChatMessages.svelte
+++ b/src/lib/components/chat/ChatMessages.svelte
@@ -18,6 +18,7 @@
 	export let currentModel: Model;
 	export let settings: LayoutData["settings"];
 	export let models: Model[];
+	export let preprompt: string | undefined;
 	export let readOnly: boolean;
 
 	let chatContainer: HTMLElement;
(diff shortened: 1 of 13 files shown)


nsarrazin:
@gary149 if you have any thoughts on how to style the preprompt, it's done [here](https://github.com/huggingface/chat-ui/pull/470/files#diff-4762d8266ac7d57eadf31bd7cc3bbab055a2b003a2d4fdfdf10b671819719f0cR47-R50), could probably be better

src/lib/components/chat/ChatMessages.svelte
```diff
@@ -42,6 +43,12 @@
 >
 	<div class="mx-auto flex h-full max-w-3xl flex-col gap-6 px-5 pt-6 sm:gap-8 xl:max-w-4xl">
 		{#each messages as message, i}
+			{#if i === 0 && preprompt}
```
gary149, line 46: Can you share a screenshot? not sure we want to render that here.

src/lib/components/chat/ChatMessages.svelte
```diff
@@ -18,6 +18,7 @@
 	export let currentModel: Model;
 	export let settings: LayoutData["settings"];
 	export let models: Model[];
+	export let preprompt: string | undefined;
```
gary149, line 21: systemPrompt instead of preprompt?

nsarrazin:
image

----

image

--- 

image

---

thoughts ? @gary149

gary149:
Made some small tweaks! let's do [this](https://github.com/huggingface/chat-ui/pull/470#discussion_r1345783191) then merge?

src/lib/components/chat/ChatMessages.svelte
nsarrazin: Looking at it, we've been calling it `preprompt` in the .env for [a while now](https://github.com/huggingface/chat-ui/commit/cd6894da7d7940315f2e69bbbe670dc1d4ecb575#diff-9a6d1a5a34d34f4a64bbe39b0cbf3d6a6e6cbd98943f8503a6aeecc091829d25), and if we change it, it means a breaking change for users, as they will need to update their .env. 

If you think it's worth the change I can still do it though

gary149 (approved):

PR #488, opened 2023-10-09
Version upgrade to Svelte 4

nsarrazin:
This will make sure we're up to the latest version and bring about quite a few benefits, including [DX and performance related](https://svelte.dev/blog/svelte-4).


src/lib/components/Toast.svelte
```diff
@@ -7,7 +7,7 @@
 </script>
 
 <div
-	transition:fade={{ duration: 300 }}
+	transition:fade|global={{ duration: 300 }}
```
gary149, line 10: it's the default now?

nsarrazin: `|local` is the default now and `|global` needs to be specified. This used to be the opposite!

Issue #497, opened 2023-10-09
Feature request: Allow disable ask LLM for conversation summery


itaybar:
This feature requiers to much resources for this extra LLM api call (which can be very expensive) so I tought it can be configured rather it just put the first user input as tab name or asking for lllm to summerize that

nsarrazin:
Closed with #498 ! Thanks for the contrib

PR #498, opened 2023-10-09
Added optional llm summerization

itaybar:
#497


src/lib/server/summarize.ts
```diff
@@ -1,8 +1,14 @@
...
 import { generateFromDefaultEndpoint } from "$lib/server/generateFromDefaultEndpoint";
 import { defaultModel } from "$lib/server/models";
+import {
+	LLM_SUMMERIZATION
+} from "$env/static/private";
 
 export async function summarize(prompt: string) {
+	if (!LLM_SUMMERIZATION) {
+		return prompt;
+	}
```
julien-blanchon, line 11: Maybe we can slice the prompt so it will not take too much space in the database. 
```suggestion
	if (!LLM_SUMMERIZATION) {
		return prompt.slice(0, 15);
	}
```

src/lib/server/summarize.ts
```diff
@@ -1,8 +1,14 @@
 import { buildPrompt } from "$lib/buildPrompt";
 import { generateFromDefaultEndpoint } from "$lib/server/generateFromDefaultEndpoint";
 import { defaultModel } from "$lib/server/models";
+import {
+	LLM_SUMMERIZATION
+} from "$env/static/private";
 
 export async function summarize(prompt: string) {
+	if (!LLM_SUMMERIZATION) {
+		return prompt;
```
nsarrazin, line 10:
```suggestion
		return prompt.split(/\s+/g).slice(0,5).join(' ');
```
Maybe we could split the prompt on white spaces and only take the first 5 words, what do you think ?

nsarrazin (commented):
Hi, thanks for the contrib! Great idea, just left some minor feedback

nsarrazin:
lol @julien-blanchon beat me to the punch :grin:

src/lib/server/summarize.ts
julien-blanchon: Sound like even better

nsarrazin:
Thanks for the contribution! This is great, merging it.

src/lib/server/summarize.ts
```diff
@@ -1,8 +1,12 @@
 import { buildPrompt } from "$lib/buildPrompt";
 import { generateFromDefaultEndpoint } from "$lib/server/generateFromDefaultEndpoint";
 import { defaultModel } from "$lib/server/models";
+import { LLM_SUMMERIZATION } from "$env/static/private";
```
julien-c, line 4: nit: typo => LLM_SUMMARIZATION

ruizcrp:
Just tried with version 0.8.1
LLM_SUMMERIZATION=false
(as well as because of the typo
LLM_SUMMARIZATION=false)
and it didn't turn it off. 

I looked into summarize.ts and don't see why it doesn't work. 
Have you tried it?
I see it not only because the summary was generated, but also because the respective connection is made to the api.

PR #486, opened 2023-10-08
[WebSearch] Update 3

mishig25:

PR #502, opened 2023-10-11
[Websearch] Turns website text into markdown and use sliding window chunker

mishig25:
### Websearch update 

1. Update scraping: use [turndown](https://github.com/mixmark-io/turndown) to turn html string (scraped website content) into markdown string (since most models understand markdown quite well)
2. Update chunking: sliding window chunking based on makrdown headings structure (Read below for example)

#### Example of chunking with markdown structure

let's a website gets turn into this markdown:
```
# Title

Some text

## Subtitle 1

Some text 2

## Subtitle 2

Some text 3
```

then section texts will be chunked to have its heading and parent headings. For example, the chunks will look like:
```ts
const chunks: string[] = [
"# Title\n\nSome text",
"# Title\n\n## Subtitle 1\n\nSome text 2",
"# Title\n\n## Subtitle 12n\nSome text 3",
]
```

the reasoning was that: preserving headings structure in every chunk would be a better context rather than chunking naively. Example of naive chunking that doesn't preserve the markdown structure:
```ts
const chunks: string[] = [
"# Title\n\nSome text\n\n## Subtitle 1",
"Some text 2\n\n## Subtitle 2\n\nSome text 3",
]
```

#### Explanation of chunking with sliding window

Instead of chunking naively, sliding window chunking (as the name suggests) chunks in a way that neighbouring chunks share commend end/beginning.

## Testing locally

Make sure to run `npm ci` to get the new dependencies

** note: in https://github.com/huggingface/chat-ui/pull/486, also tried Auto Merging Retriever. However, the results weren't better than what it currently is.


nsarrazin:
Hi @mishig25 quick question, what's the status on this PR, still a draft or should it be reviewed? 👀

mishig25:
As I tested it more, this PR implementation didn't improve the results on what we have currently. Therefore, closing this PR for now. However, will keep experimenting with other chunking/text parsing methods to improve the results

PR #506, opened 2023-10-12
fix inert attribute didn't removed properly after modal close

Bkyaro:
Hi team, I noticed a bug that after close the setting modal, whole site will be ignored due to #app node didn't remove the inert attribute.
It seems like the document.querySelectorAll('[role="dialog"]:not(#app *)') in onDestroy hook in Modal.svelte file selected 2 nodes because it runs before rerender.So I modify the condition to 2 instead of 1.
love your huggingface chat bot,best wishes :D
modal close error

### DIFF
--- a/src/lib/components/Modal.svelte
+++ b/src/lib/components/Modal.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-	import { createEventDispatcher, onDestroy, onMount } from "svelte";
+	import { afterUpdate, createEventDispatcher, onDestroy, onMount } from "svelte";
 	import { cubicOut } from "svelte/easing";
 	import { fade } from "svelte/transition";
 	import Portal from "./Portal.svelte";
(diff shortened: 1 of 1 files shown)


nsarrazin:
Hi! I can't reproduce the issue this solves, just out of curiosity what's your OS/browser config?

julien-blanchon:
I can't reproduce either on Firefox/Brave/Chrome

Bkyaro:
> Hi! I can't reproduce the issue this solves, just out of curiosity what's your OS/browser config?

you can reproduce in this site https://huggingface.co/chat
the https://huggingface.co/spaces/huggingchat/chat-ui site can't reproduce.
I reproduce on MAC & win10 both chrome :D

Bkyaro:
I tried in incognito window,can not reproduce. It should be my own fault for that. Sorry for disturbing, case closed :D

PR #508, opened 2023-10-13
Update CI scripts for dotenv management

nsarrazin:
You can  now run `npm run updateLocalEnv`. It will merge the contents of a local file `.env.SECRET_CONFIG` with `.env.template` into your `.env.local`

This lets you get the exact same config as prod really easily by just running a script. Handy for debugging model configs

I also replaced the python script used for updating prod by a TS script which does the same thing.


scripts/updateProdEnv.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+// Read the content of the file .env.template
+const PUBLIC_CONFIG = fs.readFileSync(".env.template", "utf8");
+
+// Prepend the content of the env variable SECRET_CONFIG
+const full_config = `${PUBLIC_CONFIG}\n${SECRET_CONFIG}`;
+
+// Make an HTTP POST request to add the space secre
+fetch(`https://huggingface.co/api/spaces/huggingchat/chat-ui/secrets`, {
```
nsarrazin, line 21: I don't think `@huggingface/hub` supports the spaces secrets API yet so I made a manual API call here. Tested it on my own space and it works 

@coyotte508 for visibility, not sure if it's planned or not haha

PR #512, opened 2023-10-16
Title update occur when available

nsarrazin:
Before the title would only change when invalidating the dependency (after message is finished streaming)

Now the title is passed from the backend during the streaming and is updated when available. 

This required the use of a store in order to update the conv title in both  `src/routes/+layout.svelte` and `src/routes/conversation/[id]/+page.svelte`


src/lib/stores/titleQueue.ts
```diff
@@ -0,0 +1,32 @@
+import { writable } from "svelte/store";
+
+export interface TitleUpdate {
+	convId: string;
+	title: string;
+}
+
+const queueStore = () => {
```
gary149, line 8: Why do you need a queue?

nsarrazin: In case we end up with multiple updates for multiple conversations at the same time somehow, it's better handled as a queue

gary149: This seems like an edge case no? I would have kept it as simple as possible.

nsarrazin: see https://github.com/huggingface/chat-ui/pull/512/commits/59d978a27bd6dd34b49370bb42487737eee1c031

PR #520, opened 2023-10-18
Add checkbox for displaying emojis in sidebar

nsarrazin:
https://github.com/huggingface/chat-ui/assets/25119303/98d15803-a37b-48e7-a539-90ace6ffccee

### DIFF
--- a/src/lib/components/SettingsModal.svelte
+++ b/src/lib/components/SettingsModal.svelte
@@ -75,6 +75,14 @@
 					</ul>
 				</div>
 			{/if}
+			<label class="flex cursor-pointer select-none items-center gap-2 text-sm text-gray-500">
+				<input
+					type="checkbox"
+					name="hideEmojiOnSidebar"
+					bind:checked={settings.hideEmojiOnSidebar}
+				/>
+				Hide emoji on sidebar
+			</label>
 			<form
 				method="post"
 				action="{base}/conversations?/delete"
--- a/src/lib/components/SettingsModal.svelte
+++ b/src/lib/components/SettingsModal.svelte
@@ -81,7 +81,7 @@
 					name="hideEmojiOnSidebar"
 					bind:checked={settings.hideEmojiOnSidebar}
 				/>
-				Hide emoji on sidebar
+				Show emoticons in conversation topics
 			</label>
 			<form
 				method="post"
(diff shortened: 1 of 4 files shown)


nsarrazin:
I think it would be nicer if the form was blocking while applying the change to settings. 

Now it closes immediately and the changes happens in the background which makes UX a bit confusing, could be done later though. wdyt @gary149 ?

src/lib/components/SettingsModal.svelte
```diff
@@ -75,6 +75,14 @@
 					</ul>
 				</div>
 			{/if}
+			<label class="flex cursor-pointer select-none items-center gap-2 text-sm text-gray-500">
+				<input
+					type="checkbox"
+					name="hideEmojiOnSidebar"
+					bind:checked={settings.hideEmojiOnSidebar}
+				/>
+				Hide emoji on sidebar
```
gary149, line 84: nit but i's not only is it in the sidebar, it's also in the top-level mobile navigation.

gary149: I think my proposed wording was fine: "Show emoticons in conversation topics"

src/routes/+layout.server.ts
```diff
@@ -51,7 +51,7 @@ export const load: LayoutServerLoad = async ({ locals, depends, url }) => {
 			})
 			.map((conv) => ({
 				id: conv._id.toString(),
-				title: conv.title,
+				title: settings?.hideEmojiOnSidebar ? conv.title.replace(/\p{Emoji}/gu, "") : conv.title,
```
gary149, line 54: It's a bit weird to hide there? I would hide at bottom level (Svelte component)

julien-c:
i haven't followed, but is there really a need for a end-user option for this?

gary149:
> Now it closes immediately and the changes happens in the background which makes UX a bit confusing, could be done later though. wdyt @gary149 ?

I'm not too shocked by it. I think it's fine for now.

src/routes/+layout.server.ts
nsarrazin: Thinking about this more, I think it makes more sense to keep it at the top level.

Both the page title, top-level mobile nav, and `NavConversationItem` use this layout data as a source of truth, so I would prefer to change it once at the top level than in three places.

src/lib/components/SettingsModal.svelte
nsarrazin: Updated the wording!

PR #521, opened 2023-10-18
[Websearch] UI fix sources

mishig25:
Fixes a bug where at the beginning of the model answer it seems to show 1 source and all other sources appear at the end of the answer


nsarrazin (approved):
Works as intended!

PR #522, opened 2023-10-18
manifest: /chat/

gary149:


nsarrazin (approved):
I think in the long run we probably want to find a way to use the env var `APP_BASE` for determining this path, because for now [the manifest for chat-ui](https://github.com/huggingface/chat-ui/blob/9d761a91034c8d70266bab47d8cc651278cc74af/static/chatui/manifest.json) will only work if `APP_BASE` is not set

This fix will work for huggingchat in prod though so we should just go ahead with this for now

gary149:
Yes, this is a quick fix but you are right.

PR #523, opened 2023-10-18
Break up SECRET_CONFIG into individual secrets

nsarrazin:


nsarrazin:
@glegendre01 I think this is better now, I broke everything down into individual secrets.

glegendre01:
seems better ! next step renewing yours keys :)

PR #530, opened 2023-10-24
Add a more visible retry button on failure to generate

nsarrazin:
How to test:

1. Edit `.env.template` so that inside of `MODELS` one of the `"name"` is not a real model
2. `npm run updateLocalEnv` 
3. Then start the server and start a conversation with the non-existent model, you should see the new button/message.
image

### DIFF
--- a/src/lib/components/chat/ChatWindow.svelte
+++ b/src/lib/components/chat/ChatWindow.svelte
@@ -17,6 +17,7 @@
 	import type { WebSearchUpdate } from "$lib/types/MessageUpdate";
 	import { page } from "$app/stores";
 	import DisclaimerModal from "../DisclaimerModal.svelte";
+	import RetryBtn from "../RetryBtn.svelte";
 
 	export let messages: Message[] = [];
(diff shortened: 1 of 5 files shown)


src/lib/components/chat/ChatWindow.svelte
```diff
@@ -45,6 +46,8 @@
 		dispatch("message", message);
 		message = "";
 	};
+
+	$: lastIsError = messages[messages.length - 1]?.from === "user" && !loading;
```
nsarrazin, line 50: The logic check for displaying the error is here, I think that covers all use cases. 

Basically if the last message is from a user but it's not currently loading a new message, it means something went wrong.

gary149: Ok 👀, looks a bit weird but if it works :)

gary149:
Simplified a bit (stop/retry button always on the right side, no colors)

image

gary149 (approved):

Issue #441, opened 2023-09-16
Feature Request: Convert PDF to Markdown inside Chat UI


julien-blanchon:
## Feature Description:

I'd like to propose the following feature:  Add a **PDF to Markdown converter and they include this Markdown content directly in the chat**. This feature would not only enhance the user experience but also provide a seamless way to discuss and reference content from PDFs, especially research papers.

It's worth noting that similar features exist in platforms like the Anthropic Claude 2 and Perplexity.AI interface. However, their implementations primarily focus on a basic text extraction, which often results in a lossy conversion. Specifically, with  mathematical equations, tables, and other complex formatting are not retaine. My proposed implementation aims to address this limitation by ensuring a more comprehensive and lossless conversion using the **Mathpix Paid API**, and maybe other service in the future.

## Design Implementation:

* Dropdown on Hover: When a user hovers the chat form with a file, a dropdown menu will appear. This dropdown will provide an option to "Upload PDF".
* Once the file is released this will send the file to be converted.
* After the conversion, the markdown content will be append in the chat.

## Example

![demo (1)](https://github.com/huggingface/chat-ui/assets/11278197/f3bf954d-043c-4862-af29-44cd9496ac2b)

## Pull Request

https://github.com/huggingface/chat-ui/pull/442

nsarrazin:
Hey @julien-blanchon thanks for contributing all of this! I'm a bit busy atm with the agents feature but I have seen your PR #449 and I'll review it once I have a bit more time 😁 Just so you know it hasn't been forgotten haha

julien-blanchon:
Yes, but this is still a draft. As it don't fit the usage of 99% of the hf-chat user I don't know if we sould merge it or not (or maybe disable the feature by default). 
However we can maybe generalize the file features with:
- The file drag and drop compotents
- A abstract system for processing file object dependings of the user settings (process to mathpix could be one of them).

nsarrazin:
Yeah I think in particular your drag&drop feature looks great! Would it be okay if we used it for agents?

julien-blanchon:
Yes of course, you're talking about https://github.com/huggingface/chat-ui/pull/462 ?

nsarrazin:
Yeah! We already support in the backend uploading images & audios for agents (stuff like speech transcription, image description), there's just no UI component for it so I think yours could work well when we implement that feature!

gary149:
Hi, good news! @mishig25 has started working on this feature so it should come soon!

[...]

PR #531, opened 2023-10-24
Add support for tgi multimodal models

nsarrazin:
Working for now but things that still need fixing:

- [x] Better UI for deleting an uploaded image (like a ❌ button or something)
- [x] Better UI for displaying an image in the chat
- [x] Handle resizing of images in the front-end
- [x] Add a button for uploading an image without dragging it (for mobile for example)
- [x] Disable dragging feature if `multimodal:true` is not set in the config for the model
- [x] Check image size on backend
- [x] Retry button doesn't retry with images
- [x] share conversation with images
- [x] when sending a message, show images in optimistic updates
- [x] Only allow 1 image upload max

## How to test
1. pull this pr
2. `npm run updateLocalEnv`
3. `npm run dev`

Start a conversation with IDEFICS to see the new ui

## Screenshots

image
image


nsarrazin:
@julien-blanchon I reused your dropzone component for this feature 🤗 thanks a lot for making it!

julien-blanchon:
Nice ! I'm pretty hype by this PR btw 👀

julien-blanchon:
Hey @nsarrazin I'm thinking of dropping Mathpix dependencies in my implementation of **Convert PDF to Markdown inside Chat UI** (https://github.com/huggingface/chat-ui/issues/441). 
And include two text extractors:
- A basic text extractor that extracts pure text from the PDF
- A more advanced text extractor that uses an advanced OCR like https://huggingface.co/facebook/nougat-base, and uses the hosted inference API of these templates with the user token API.

Are you interested in this functionality on the huggingchat side?

If so, how can we work together? Is this functionality included in your multimodal tgi implementation?
We could refactor the code a bit to enable the use of multiple file types and multiple "agents", what are your plans in this regard?

nsarrazin:
Hey @julien-blanchon!

I think as a rule of thumb it's good to decouple any dependencies (especially remote APIs) from the feature itself if possible (see web search for example where we support three different providers now), so that people can configure the pdf parsing that they want. Maybe have some kind of standardized interface that takes a pdf file and returns the extracted text, so that people can copy the method and implement their own versions in future PRs?

And I think that could be a cool feature, maybe when this (#531) PR is merged we can have a look to see how to hook it up? This PR already adds support for passing files to the backend,so we could have a logic check that handles files differently based on mime type, like you mentioned.

nsarrazin:
This is pretty much done and ready for review, I think I covered every edge case of the feature! 😄

.env.template
```diff
@@ -109,6 +109,36 @@ MODELS=`[
...
+          "title": "Code a snake game",
+          "prompt": "Code a basic snake game in python, give explanations for each step."
+        }, {
+          "title": "Assist in a task",
+          "prompt": "How do I make a delicious lemon cheesecake?"
+        }
+      ]
+    },
+        {
+      "name": "HuggingFaceM4/idefics-80b-instruct",
```
nsarrazin, line 126: We can remove this change to .env.template if we don't want IDEFICS in production for HuggingChat

nsarrazin: cc @julien-c to confirm this

src/lib/components/chat/ChatMessage.svelte
```diff
@@ -234,36 +234,57 @@
...
-						on:click={() => dispatch("retry", { content: message.content, id: message.id })}
-					>
-						<CarbonRotate360 />
-					</button>
-				{/if}
+		<div class="flex flex-col">
+			{#if message.files && message.files.length > 0}
+				<div class="mx-auto grid w-fit grid-cols-2 gap-5 px-5">
+					{#each message.files as file}
+						{#if file.length === 64}
```
mishig25, line 241: what is this number `64`?

src/lib/components/chat/ChatMessage.svelte
```diff
@@ -234,36 +234,57 @@
...
+							<img
+								src={$page.url.pathname + "/output/" + file}
+								alt="input from user"
+								class="my-2 aspect-auto max-h-48 rounded-lg shadow-lg"
+							/>
+						{:else}
+							<img
+								src={"data:image/png;base64," + file}
```
mishig25, line 249: is it the response guranteed to be `png` all the time?

src/lib/components/chat/ChatWindow.svelte
```diff
@@ -47,7 +53,25 @@
 		message = "";
 	};
 
+	let lastTarget: EventTarget | null = null;
+
+	let onDrag = false;
+
+	const onDragEnter = (e: DragEvent) => {
+		lastTarget = e.target;
+		onDrag = true;
```
mishig25, line 62: two comment on the `FileDropZone`:
1. when `onDrag=true`, there should be some kind of visual feedback (like the background changed to different color etc.)
2. At the moment, the `FileDropZone` only works if you drag/drop file on the `Upload image` btn. It may make sense to have a similar behaviour as ChatGPT, where `FileDropZone` works on the wntire window, not just `Upload image` btn

src/lib/components/chat/FileDropzone.svelte
```diff
@@ -0,0 +1,97 @@
+<script lang="ts">
```
mishig25, line 1: made some suggestions https://github.com/huggingface/chat-ui/pull/544

src/lib/types/Message.ts
```diff
@@ -10,4 +10,5 @@ export type Message = Partial<Timestamps> & {
 	webSearchId?: WebSearch["_id"]; // legacy version
 	webSearch?: WebSearch;
 	score?: -1 | 0 | 1;
+	files?: string[]; // filenames
```
mishig25, line 13: nit: can it be renamed to `fileNames` ?

src/lib/buildPrompt.ts
```diff
@@ -49,6 +53,38 @@ export async function buildPrompt({
...
 	}
 
+	// section to handle potential files input
+	if (model.multimodal) {
+		messages = await Promise.all(
+			messages.map(async (el) => {
+				let content = el.content;
+
+				if (el.from === "user") {
+					if (el.files && url && fetch && id && el.files?.length > 0) {
```
mishig25, line 63:
```suggestion
					if (el?.files?.length > 0 && url && fetch && id) {
```
I think we can combine the two `el.files` checks here

mishig25:
except the nits, I left looks very close to being merged 👍

gary149 (approved):
LGTM. We could add paste support and full-screen dropzone in another PR.

src/routes/conversation/[id]/+page.svelte
```diff
@@ -79,14 +81,34 @@
 				retryMessageIndex = messages.length;
 			}
 
+			const module = await import("browser-image-resizer");
+
+			const resizedImages = await Promise.all(
+				files.map(async (file) => {
+					return await module
+						.readAndCompressImage(file, {
+							maxHeight: 224,
```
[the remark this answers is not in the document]
Comment (nsarrazin) [reply src/routes/conversation/[id]/+page.svelte:90]: Agreed that it should be in the config in the long run! The reason I didn't make it configurable for now is that TGI hardcodes the images to 224px anyway for now. When TGI starts supporting more models we should make this configurable, wdyt?

src/lib/components/chat/ChatMessage.svelte
nsarrazin: I ended up using a wildcard mime type `image/*` since I don't have an easy way of getting the mime type in that situation, seems to works fine for all formats I tested (png, jpeg, webp)

src/lib/types/Message.ts
nsarrazin: I fixed the comment so it's more clear what's inside!

src/lib/components/chat/ChatMessage.svelte
nsarrazin: I added some comments to make things clearer

src/routes/conversation/[id]/+page.svelte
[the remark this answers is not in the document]
Comment (nsarrazin) [reply src/routes/conversation/[id]/+page.svelte:93]: added a comment about it

src/lib/components/chat/ChatWindow.svelte
nsarrazin: I think [we'll handle that in a separate PR](https://github.com/huggingface/chat-ui/pull/531#pullrequestreview-1731936802) but definitely good points

nsarrazin:
Going to remove IDEFICS from the prod template, add a readme note about using it and then merge this

PR #539, opened 2023-10-31
Set zephyr 7B beta for small tasks

nsarrazin:


gary149:
Doesn't seems to work better than current. I think we can close for now.

PR #428, opened 2023-09-05
Implement general awsEndpoint + lambda integration

EikeKohl:
This small refactoring PR implements a general `awsEndpoint`, enabeling the user to not only connect sagemaker endpoints, but also lambda function URLs that require AWS_IAM authentication.
I had to implement this change for my personal setup, because upon requesting an LLM generation, the lambda performs additional scaling logic to provision / delete the sagemaker endpoint to save LLM hosting costs. I think that everyone could benefit from this functionality.

PR #541, opened 2023-11-01
Modular backends & support for openAI & AWS endpoints

nsarrazin:
This PR adds support for openAI endpoint with a refactor that make it easier to handle adding independent backends with library dependencies in the future. For example if you want to add `gpt-3.5-turbo`:

Add to your `MODELS` : 
```
{
      "name": "gpt-3.5-turbo",
      "displayName": "ChatGPT",
      "preprompt": "",
      "parameters": {
        "temperature": 0.1,
        "top_p": 0.95,
        "repetition_penalty": 1.2,
        "top_k": 50,
        "truncate": 1000,
        "max_new_tokens": 2048
      },
      "endpoints" : [{
        "type": "openai"
      }]
}
```

and add `OPENAI_API_KEY` to your .env.local

Closes #452 
Closes #428


nsarrazin:
cc @chenhunghan thanks for all your work on the openai compatible endpoints. If you have time can you test this? I refactored things to make it more modular in the future, let me know what you think

README.md
```diff
@@ -258,23 +307,45 @@ You can then add the generated information and the `authorization` parameter to
...
-      "url": "", // your aws sagemaker url here
+      "type" : "aws",
+      "service" : "sagemaker"
+      "url": "",
       "accessKey": "",
       "secretKey" : "",
-      "sessionToken": "", // optional
-      "weight": 1
+      "sessionToken": "",
+      "weight": 1,
```
chenhunghan, line 325: json probably disallows trailing commas

README.md
```diff
@@ -168,6 +168,55 @@ MODELS=`[
...
+      "top_p": 0.95,
+      "repetition_penalty": 1.2,
+      "top_k": 50,
+      "truncate": 1000,
+      "max_new_tokens": 1024,
+      "stop": []
+    },
+    "endpoints": [{
+      "type" : "openai",
+      "baseURL": "http://localhost:8000/v1",
```
chenhunghan, line 193:
```suggestion
      "baseURL": "http://localhost:8000/v1"
```
Same as https://github.com/huggingface/chat-ui/pull/541/files#r1381215132

README.md
```diff
@@ -168,6 +168,55 @@ MODELS=`[
 
 You can change things like the parameters, or customize the preprompt to better suit your needs. You can also add more models by adding more objects to the array, with different preprompts for example.
 
+##### OpenAI API compatible models
+
+Chat UI can be used with any API server that supports OpenAI API compatibility, for example [oobabooga](https://github.com/oobabooga/text-generation-webui/tree/main/extensions/openai), [LocalAI](https://github.com/go-skynet/LocalAI), [FastChat](https://github.com/lm-sys/FastChat/blob/main/docs/openai_api.md), [llama-cpp-python](https://github.com/abetlen/llama-cpp-python), and [ialacol](https://github.com/chenhunghan/ialacol).
+
+The following example config makes Chat UI works with [oobabooga](https://github.com/oobabooga/text-generation-webui/tree/main/extensions/openai), the `endpoint.baseUrl` is the url of the OpenAI API compatible server, this overrides the baseUrl to be used by OpenAI instance. The `endpoint.type` determine which endpoint to be used, default is `chat_completions` which uses `v1/chat/completions`, change to `endpoint.type` to `completions` to use the `v1/completions` endpoint.
```
chenhunghan, line 175: We are using `.completion` in this PR
```suggestion
The following example config makes Chat UI works with [oobabooga](https://github.com/oobabooga/text-generation-webui/tree/main/extensions/openai), the `endpoint.baseUrl` is the url of the OpenAI API compatible server, this overrides the baseUrl to be used by OpenAI instance. The `endpoint.completion` determine which endpoint to be used, default is `chat_completions` which uses `v1/chat/completions`, change to `endpoint.completion` to `completions` to use the `v1/completions` endpoint.
```

src/lib/server/models.ts
```diff
@@ -73,13 +44,16 @@ const modelConfig = z.object({
 			})
 		)
 		.optional(),
-	endpoints: z.array(combinedEndpoint).optional(),
+	endpoints: z.array(endpointSchema).optional(),
 	parameters: z
 		.object({
 			temperature: z.number().min(0).max(1),
 			truncate: z.number().int().positive(),
```
chenhunghan, line 51: Not sure `truncate` should be mandatory, does not make much sense use with OAI

src/lib/server/endpoints/openai/endpointOai.ts
```diff
@@ -0,0 +1,82 @@
...
+	baseURL,
+	apiKey,
+	completion,
+	model,
+}: z.infer<typeof endpointOAIParametersSchema>): Promise<Endpoint> {
+	let OpenAI;
+	try {
+		OpenAI = (await import("openai")).OpenAI;
+	} catch (e) {
+		throw new Error("Failed to import OpenAI");
```
chenhunghan, line 29: Better error propagation 
```suggestion
		throw new Error("Failed to import OpenAI", { cause: e });
```

README.md
nsarrazin: Yep you're right, thanks for catching it

src/lib/server/models.ts
nsarrazin: Yeah I agree, I also think parameters should be defined per endpoint, not per model. That would allow for specifying the right params per endpoint type.

nsarrazin:
Tested this locally with openAI, the llama.cpp local server with openAI adapter and also plain TGI and everything works great. 

I think this is good to go imo!

nsarrazin:
cc @coyotte508 if you want to take a quick peek, especially in [endpoints.ts](https://github.com/huggingface/chat-ui/pull/541/files#diff-0945ef3419fa99d22cf33eb6d699d7c81dd5dc302411df485b5d4cea5e0643c0) and [models.ts](https://github.com/huggingface/chat-ui/pull/541/files#diff-ac854d5672af10370323e57d41c389221362fd685d219fc5b74c6ff0d23fc6e8R75-R115)

README.md
```diff
@@ -168,6 +168,55 @@ MODELS=`[
 
 You can change things like the parameters, or customize the preprompt to better suit your needs. You can also add more models by adding more objects to the array, with different preprompts for example.
 
+##### OpenAI API compatible models
+
+Chat UI can be used with any API server that supports OpenAI API compatibility, for example [oobabooga](https://github.com/oobabooga/text-generation-webui/tree/main/extensions/openai), [LocalAI](https://github.com/go-skynet/LocalAI), [FastChat](https://github.com/lm-sys/FastChat/blob/main/docs/openai_api.md), [llama-cpp-python](https://github.com/abetlen/llama-cpp-python), and [ialacol](https://github.com/chenhunghan/ialacol).
```
mishig25, line 173:
```suggestion
Chat UI can be used with any API server that supports OpenAI API compatibility, for example [text-generation-webui](https://github.com/oobabooga/text-generation-webui/tree/main/extensions/openai), [LocalAI](https://github.com/go-skynet/LocalAI), [FastChat](https://github.com/lm-sys/FastChat/blob/main/docs/openai_api.md), [llama-cpp-python](https://github.com/abetlen/llama-cpp-python), and [ialacol](https://github.com/chenhunghan/ialacol).
```

correct me if I'm wrong. But I think this repo is more known as `text-generation-webui` rather than `oobabooga`

mishig25: in the readme, there are other instances of `oobabooga `

src/routes/conversation/[id]/+server.ts
```diff
@@ -191,105 +184,25 @@ export async function POST({ request, fetch, locals, params, getClientAddress })
...
-					accessToken: randomEndpoint.host === "sagemaker" ? undefined : HF_ACCESS_TOKEN,
-				},
-				{
-					use_cache: false,
-					fetch: usedFetch,
-				}
-			);
+			const endpoint = await model.getEndpoint();
 
-			for await (const output of tokenStream) {
```
Comment (chenhunghan) [inline src/routes/conversation/[id]/+server.ts:281]: Please consider refactor to move `await endpoint({ conversation: conv }))`  out of `ReadableStream()` and/or out of `Response(stream)`

In practice it's quite common `await endpoint()` throws, for example getting `409` from server because of hitting rate limiter from server, for configuration was wrong and getting `404`.

IMO it make little sense to start streaming with we don't know if the response return OK

for example if something throw an exception in start handler in `ReadableStream`:
```ts
const stream = new ReadableStream({
    async start(controller) {
        throw new Error("broken")
    }
});
return new Response(stream);
```

Client will just get an `Network` error, IMO we should only return `Response`, i.e., start streaming when external API are green.
```ts
// handle error before returning `new Response(stream)`.
try {
  const example = await endpoint({ conversation: conv }));
} catch (e) {
  throw error(500, e.message);
}
const stream = new ReadableStream({
    async start(controller) {
        for await (const output of example) {
            ...
        }
    }
});
return new Response(stream);
```

wdhorton:
I've been testing this branch and from what I've seen, it seems to break the LLM Summarization for the conversation title

Maemol (commented):
I'm using mostly Azure Open AI so I tested this branch with an Azure Endpoint. With some minor tweaks it works well so thanks for this PR!

I also tested the websearch but I had some issues with the GenerateQuery that didn't understand the preprompt (system prompt) correctly.

Appart from that, it's working well for me.

src/lib/server/endpoints/openai/endpointOai.ts
```diff
@@ -0,0 +1,82 @@
...
+	});
+
+	if (completion === "completions") {
+		return async ({ conversation }) => {
+			return openAICompletionToTextGenerationStream(
+				await openai.completions.create({
+					model: model.id ?? model.name,
+					prompt: await buildPrompt({
+						messages: conversation.messages,
+						webSearch: conversation.messages[conversation.messages.length - 1].webSearch,
```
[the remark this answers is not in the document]
golgeek: Had the same experience and fixed it by changing:
```preprompt: conversation.preprompt```
with
```preprompt: conversation.preprompt ?? model.preprompt,```

From what I understood from the code, `conversation.preprompt` is defined only when someone customized the preprompt in the UI.

**Comment by Fadude** [comment]
Looking forward for this PR 👀

**Comment by nsarrazin** [comment]
I was on holiday last week, I'll be back Tuesday and implement the feedback/merge this then!

File: src/lib/server/endpoints/openai/endpointOai.ts (hunk 8 above)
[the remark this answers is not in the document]
Comment (nsarrazin) [reply src/lib/server/endpoints/openai/endpointOai.ts:44]: Should be fixed! Thanks for catching this

File: src/lib/server/endpoints/openai/openAICompletionToTextGenerationStream.ts [hunk 9]
```diff
@@ -0,0 +1,32 @@
...
+ * Transform a stream of OpenAI.Completions.Completion into a stream of TextGenerationStreamOutput
+ */
+export async function* openAICompletionToTextGenerationStream(
+	completionStream: Stream<OpenAI.Completions.Completion>
+) {
+	let generatedText = "";
+	let tokenId = 0;
+	for await (const completion of completionStream) {
+		const { choices } = completion;
+		const { text } = choices[0];
```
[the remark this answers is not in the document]
nsarrazin: Should be fixed!

src/lib/server/endpoints/openai/openAIChatToTextGenerationStream.ts
```diff
@@ -0,0 +1,34 @@
...
+export async function* openAIChatToTextGenerationStream(
+	completionStream: Stream<OpenAI.Chat.Completions.ChatCompletionChunk>
+) {
+	let generatedText = "";
+	let tokenId = 0;
+	for await (const completion of completionStream) {
+		const { choices } = completion;
+		const {
+			delta: { content },
+		} = choices[0];
```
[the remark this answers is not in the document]
nsarrazin: Should be fixed!

src/lib/server/websearch/generateQuery.ts
```diff
@@ -62,10 +61,8 @@ Current Question: Where is it being hosted ?`,
 		},
 	];
 
-	const promptQuery = smallModel.chatPromptRender({
-		preprompt: `You are tasked with generating web search queries. Give me an appropriate query to answer my question for google search. Answer with only the query. Today is ${currentDate}`,
+	return await generateFromDefaultEndpoint({
 		messages: convQuery,
+		preprompt: `You are tasked with generating web search queries. Give me an appropriate query to answer my question for google search. Answer with only the query. Today is ${currentDate}`,
```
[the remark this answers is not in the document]
nsarrazin: Can you try again? I made a mistake with the openai endpoint, should be better now

src/routes/conversation/[id]/+server.ts
Comment (nsarrazin) [reply src/routes/conversation/[id]/+server.ts:281]: I think that's a valid point but a bit more intense of a refactor than I think I want to do in this PR specifically. 

We currently need to be in the stream before creating the token stream. For example, when we do requests to the web search that happens before creating the token stream but we also send updates about that so we need to be streaming already...

README.md
nsarrazin: fixed!

nsarrazin:
I also just added support directly for the llama.cpp API server directly since the protocol is a bit different!

PR #544, opened 2023-11-03
Update filedropzone

mishig25:


src/lib/components/chat/FileDropzone.svelte
```diff
@@ -1,11 +1,12 @@
 <script lang="ts">
+	import { onDestroy } from "svelte";
 	import CarbonImage from "~icons/carbon/image";
 	// import EosIconsLoading from "~icons/eos-icons/loading";
```
julien-blanchon, line 4: I think we can remove this old import

julien-blanchon (approved):

src/lib/components/chat/FileDropzone.svelte
julien-blanchon, line 4: I think we can remove this old import

nsarrazin (approved):
Looks good to me! Will merge it into the other PR and continue there

Issue #547, opened 2023-11-08
Stream Generation on the UI seems not works well.


jason9693:
**Description:**
Currently, I am facing an issue in my service where tokens enqueued from the server should be used to display data in the UI as a stream, but there seems to be a problem in this process. Below is a detailed explanation of the issue.

**Scenario:**

Tokens are enqueued from the server.
The UI should receive enqueued tokens and display them as a streaming data feed.
However, the streaming functionality in the UI is not working as expected.

**Problem:**

Enqueued tokens are not being displayed in the UI as a stream especially short answer.
The UI server receives tokens but fails to render or update them.

https://github.com/huggingface/chat-ui/assets/11185336/404deb1c-9ca2-43dd-8fea-4d40b47b86ce

**Cause:**

I am currently in the process of debugging to determine the root cause of this issue. The exact cause has not been identified yet.
Reproducible Steps:

1. enqueue the token in `src/routes/conversation/[id]/+server.ts/`
- controller.enqueue(JSON.stringify(newUpdate) + "\n");
2. try dequeue with stream.read() in `src/routes/conversation/[id]/+page.svelte`
- `await reader?.read().then(async ({ done, value }) => {...`

**Expected Behavior:**

Enqueued tokens from the server should be displayed correctly as a streaming feed in the UI.

hungryalgo:
any updates? 

I find token generation in general is an issue, even with default settings, I either get a network error, or no tokens generated. This is deployment with the main branch, with no code changes on my end.

However, if you try:
https://huggingface.co/chat/
That deployment somehow has no issues. 

Many users seem to have similar issues:
https://huggingface.co/spaces/huggingchat/chat-ui/discussions/306
https://huggingface.co/spaces/huggingchat/chat-ui/discussions/300
https://huggingface.co/spaces/huggingchat/chat-ui/discussions/294

jason9693:
It seems 

`await reader?.read().then(async ({ done, value }) => {
`
function in src/routes/conversation/[id]/+page.svelte/writeMessage 
is blocked while they receive 4096 characters from the server stream sent.


<img width="527" alt="스크린샷 2023-11-13 오후 12 47 54" src="https://github.com/huggingface/chat-ui/assets/11185336/3c09d208-adee-4d12-8429-f0cde87b7b8d">

Currently I'm finding to change the stream buffer size.

nsarrazin:
[...]

PR #561, opened 2023-11-17
[#547] fix stream generation bug

jason9693:
[#547] Fix Stream Generation Bugs.

1. When the token is sliced by edge of buffer, it can merge with previous chunk. previous chunk is stored when the chunk raised parse error.
2. As the final generation is hanged when the length of final stream is less than buffer size, push the placeholder when the stream is final stage.

### DIFF
--- a/src/routes/conversation/[id]/+server.ts
+++ b/src/routes/conversation/[id]/+server.ts
@@ -200,7 +200,7 @@ export async function POST({ request, locals, params, getClientAddress }) {
 					// 4096 of spaces to make sure the browser doesn't blocking buffer that holding the response
 					controller.enqueue(" ".repeat(4096));
 				}
-				if (newUpdate.token === "") {
+				if (newUpdate.type === "stream" && newUpdate.token === "") {
 					return;
 				}
 			}
--- a/src/routes/conversation/[id]/+server.ts
+++ b/src/routes/conversation/[id]/+server.ts
@@ -194,7 +194,7 @@ export async function POST({ request, locals, params, getClientAddress }) {
 				if (newUpdate.token === "") {
 					return;
 				}
-				
+
 				if (newUpdate.type !== "stream") {
 					updates.push(newUpdate);
 				}
--- a/src/routes/conversation/[id]/+server.ts
+++ b/src/routes/conversation/[id]/+server.ts
@@ -191,12 +191,12 @@ export async function POST({ request, locals, params, getClientAddress }) {
 			const updates: MessageUpdate[] = [];
 
 			function update(newUpdate: MessageUpdate) {
-				if (newUpdate.token === "") {
-					return;
-				}
-
 				if (newUpdate.type !== "stream") {
 					updates.push(newUpdate);
+				}
+
+				if (newUpdate.token === "") {
+					return;
 				}
 				controller.enqueue(JSON.stringify(newUpdate) + "\n");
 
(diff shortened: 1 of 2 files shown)


src/routes/conversation/[id]/+server.ts
```diff
@@ -195,6 +195,15 @@ export async function POST({ request, locals, params, getClientAddress }) {
 					updates.push(newUpdate);
 				}
 				controller.enqueue(JSON.stringify(newUpdate) + "\n");
+
+				if (newUpdate.type === "finalAnswer") {
+					// 4096 of spaces to make sure the browser doesn't blocking buffer that holding the response
+					let placeholder = " ".repeat(4096);
+					controller.enqueue(placeholder);
```
Comment (nsarrazin) [inline src/routes/conversation/[id]/+server.ts:202]:
```suggestion
					controller.enqueue(" ".repeat(4096));
```
for linting 😄

src/routes/conversation/[id]/+server.ts
```diff
@@ -195,6 +195,15 @@ export async function POST({ request, locals, params, getClientAddress }) {
...
+				}
+				if (newUpdate.token === "") {
```
Comment (nsarrazin) [inline src/routes/conversation/[id]/+server.ts:204]:
```suggestion
				if (newUpdate.type === "stream" && newUpdate.token === "") {
```

This is needed for ensuring type checking works well, but i'm not sure what this line does anyway since this condition just returns, but we're already at the end of the `update`method ?

nsarrazin (commented):
Thanks for the contribution! Do you know of any way we could reliably reproduce this bug? (so I can approve the bugfix)

jason9693:
> Thanks for the contribution! Do you know of any way we could reliably reproduce this bug? (so I can approve the bugfix)

jason9693:
@nsarrazin I was tested on:
1. Chrome Browser
2. M1 Mac
3. No Web Search Engine.
4. Use Google login.

You can see the symptoms in here.
https://github.com/huggingface/chat-ui/issues/547

src/routes/conversation/[id]/+server.ts
Comment (jason9693) [reply src/routes/conversation/[id]/+server.ts:203]: I found the empty token ("")  will produced when a model targeted on language with using BPE tokenizer. (like Korean)
As the unicode-based language must combine two bytes, byte level bpe can displayed empty token.
(ex. token A + token B = unicode C => {'C', ''})

jason9693:
@nsarrazin can we proceed any update for merge it?

src/routes/conversation/[id]/+server.ts
Comment (jason9693) [reply src/routes/conversation/[id]/+server.ts:203]: 스크린샷 2023-11-20 오후 4 55 24

You can see this if you insert 
`console.log("update: " + JSON.stringify(newUpdate));
`
between 202 and 203 line.

src/routes/conversation/[id]/+server.ts
```diff
@@ -195,6 +195,14 @@ export async function POST({ request, locals, params, getClientAddress }) {
 					updates.push(newUpdate);
 				}
 				controller.enqueue(JSON.stringify(newUpdate) + "\n");
+
+				if (newUpdate.type === "finalAnswer") {
+					// 4096 of spaces to make sure the browser doesn't blocking buffer that holding the response
+					controller.enqueue(" ".repeat(4096));
+				}
+				if (newUpdate.type === "stream" && newUpdate.token === "") {
+					return;
```
Comment (nsarrazin) [inline src/routes/conversation/[id]/+server.ts:204]: Maybe i'm missing something but I think the function will return anyway after this condition so I still don't know what this is supposed to do. Currently the function does the same thing whether or not `newUpdate.token === ""`

Should the check be above the `controller.enqueue(JSON.stringify(newUpdate) + "\n");`line so it returns early?

Comment (jason9693) [reply src/routes/conversation/[id]/+server.ts]: As you mentioned, `newUpdate.token === "" `seems not work if it placed after `controller.enqueue(JSON.stringify(newUpdate) + "\n");`

So, I changed to above the 
```
if (newUpdate.type !== "stream") {
	updates.push(newUpdate);
}
```

nsarrazin:
Thanks for the contribution! This passes linting and doesn't break anything on my side so I'll merge it now 🚀

PR #569, opened 2023-11-20
Add ollama endpoint support

nsarrazin:


src/lib/server/models.ts
```diff
@@ -99,10 +99,12 @@ const addEndpoint = (m: Awaited<ReturnType<typeof processModel>>) => ({
 				} else if (args.type === "openai") {
 					return await endpoints.openai(args);
 				} else if (args.type === "llamacpp") {
-					return await endpoints.llamacpp(args);
+					return endpoints.llamacpp(args);
```
mishig25, line 102: no strong opinion. nit

we can also use `switch` here

```ts
switch (args.type) {
    case "tgi":
        return endpoints.tgi(args);
    case "aws":
        return await endpoints.aws(args);
    case "openai":
        return await endpoints.openai(args);
    case "llamacpp":
        return endpoints.llamacpp(args);
    case "ollama":
        return endpoints.ollama(args);
    default:
        // for legacy reason
        return endpoints.tgi(args);
}
```

nsarrazin: yeah that would be a lot cleaner, things kept growing but didn't think of refactoring haha, thanks!

julien-c:
cc @jmorganca for viz=)

PR #573, opened 2023-11-21
Add openchat to huggingchat config

nsarrazin:


gary149 (approved):

.env.template
```diff
@@ -121,7 +121,36 @@ MODELS=`[
...
       ]
-    }
+    },
+      {
+    "name": "openchat/openchat_3.5",
+    "displayName": "openchat/openchat_3.5",
+    "description": "OpenChat 3.5 is the #1 model on MT-Bench, with only 7B parameters.",
+    "websiteUrl": "https://huggingface.co/openchat/openchat_3.5",
+    "preprompt": "",
+    "chatPromptTemplate" : "<s>{{#each messages}}{{#ifUser}}GPT4 Correct User: {{#if @first}}{{#if @root.preprompt}}{{@root.preprompt}}\n{{/if}}{{/if}}{{content}}<|end_of_turn|>GPT4 Correct Assistant:{{/ifUser}}{{#ifAssistant}}{{content}}<|end_of_turn|>{{/ifAssistant}}{{/each}}",
```
gary149, line 131: does it need `<s>`?

mishig25, line 131: I don't think we need to do anything atm. openchat provides different templated for coding mode https://huggingface.co/openchat/openchat_3.5/blob/main/README.md?code=true#L103-L105

nsarrazin: it's the beginning of string token so I think we should leave it in? doesn't seem to be better or worse without it

gary149: Yes we can ignore for now.

gary149 (approved):

PR #575, opened 2023-11-22
Make chat title summarization non blocking

nsarrazin:
Before, we would generate the chat title before starting to stream tokens. 

That means that the time to first token was the full response time of the summarization task + the inital latency of the chat model.

Now they both run in parallel, which will help when the task model is slow.


nsarrazin:
Merging this, I tested extensively locally and it works without issue!

PR #578, opened 2023-11-23
fix: new chat when old one is still streaming

ftgomez:
This pull request addresses a bug that occurs when a specific chat is streaming a response to the user's answer, and before it's completed, the user clicks the `New Chat` button. In such cases, the `isAborted` variable is not set to `true`, causing the original chat to continue streaming the response even though it shouldn't. Additionally, it displays the user's question multiple times. The behavior can be observed in the following video:

https://github.com/huggingface/chat-ui/assets/54251473/313cab54-0c47-4d83-9472-aa69c7b594dc

To resolve this issue, I replaced the `isAborted` variable with a `writable` object from `svelte/store`. This allows it to be set to `true` when the user clicks the `New Chat` button, while remaining editable by the original component. The resulting behavior can be seen in the following video:

https://github.com/huggingface/chat-ui/assets/54251473/6078c083-cfda-4ed4-b594-26458637a804


nsarrazin:
Thanks for the great contribution! I've fixed linting and added a check to make sure loading is off when the request is aborted. Otherwise the loading indicator was showing up when hitting the "stop generation" button (as loading was still true)

Other than that, this works great so thanks for fixing this issue!

PR #579, opened 2023-11-23
Add support for passing an API key or any other custom token in the authorization header

galen-ft:
Dear Huggingface team, let me start by saying that I really appreciate all your work and the awesome Chat UI that you've built.

I've created this PR to help add support for passing in an API key (or other custom token) in the authorization header. 

This is needed to make the Chat UI communicate with a custom TGI endpoint that requires an API key, but does not support `Basic` or `Bearer` authentication.

To this end, I have:

- added the new CUSTOM_AUTHORIZATION_TOKEN variable to the `.env` file. The variable can be left empty;
- defined a custom wrapper in [src/lib/server/endpoints/tgi/endpointTgi.ts](https://github.com/huggingface/chat-ui/blob/010f7322eef5802cf565df89522b0c49d4d49849/src/lib/server/endpoints/tgi/endpointTgi.ts#L40) which adds the authorization header before the fetch() method is called;
- updated the README.md accordingly with a new `API Key or other` section under `Custom endpoint authorization`.

This works in the following way.

We can pass the (same or different) API key or custom token separately for each endpoint by adding the following to the `.env.local`:
```env
"endpoints": [
{
  "type": "tgi",
  "url": "https://HOST:PORT",
  "authorization": "API_KEY_OR_TOKEN",
}
]
# or "authorization": "Key API_KEY_OR_TOKEN"
```

As an alternative, if `CUSTOM_AUTHORIZATION_TOKEN` is specified in `.env.local`, the same authorization header is used across all custom TGI endpoints:
```env
CUSTOM_AUTHORIZATION_TOKEN="API_KEY_OR_TOKEN"
```

The custom wrapper is written so that if the `HF_ACCESS_TOKEN` is provided in `.env.local` (the token is added to the headers in a call to [makeRequestOptions()](https://github.com/huggingface/huggingface.js/blob/ccfb044414a235270cb82d8b586f81785ad04eba/packages/inference/src/lib/makeRequestOptions.ts#L33) from [textGenerationStream()](https://github.com/huggingface/chat-ui/blob/010f7322eef5802cf565df89522b0c49d4d49849/src/lib/server/endpoints/tgi/endpointTgi.ts#L31)), it will take precedence in either case.

### DIFF
--- a/.env
+++ b/.env
@@ -9,6 +9,7 @@ COOKIE_NAME=hf-chat
 HF_ACCESS_TOKEN=#hf_<token> from from https://huggingface.co/settings/token
 HF_API_ROOT=https://api-inference.huggingface.co/models
 OPENAI_API_KEY=#your openai api key here
+CUSTOM_AUTHORIZATION_TOKEN="" #your custom Authorization header
 
 # used to activate search with web functionality. disabled if none are defined. choose one of the following:
 YDC_API_KEY=#your docs.you.com api key here
--- a/.env
+++ b/.env
@@ -9,7 +9,6 @@ COOKIE_NAME=hf-chat
 HF_ACCESS_TOKEN=#hf_<token> from from https://huggingface.co/settings/token
 HF_API_ROOT=https://api-inference.huggingface.co/models
 OPENAI_API_KEY=#your openai api key here
-CUSTOM_AUTHORIZATION_TOKEN="" #your custom Authorization header
 
 # used to activate search with web functionality. disabled if none are defined. choose one of the following:
 YDC_API_KEY=#your docs.you.com api key here
--- a/README.md
+++ b/README.md
@@ -397,6 +397,27 @@ You can then add the generated information and the `authorization` parameter to 
 ]
 ```
 
+#### API Key or other
+
+In case your custom endpoint does not support `Basic` or `Bearer` authentication and you need to pass an `API Key` or any other custom token in the authorization header, add the following to your `.env.local`:
+```env
+"endpoints": [
+{
+  "type": "tgi",
+  "url": "https://HOST:PORT",
+  "authorization": "API_KEY_OR_TOKEN",
+}
+]
+# or "authorization": "Key API_KEY_OR_TOKEN"
+```
+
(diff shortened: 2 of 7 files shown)


Extremys:
Hello, did you try to pass "authorization": "Bearer XXXXXXXX" ?

nsarrazin (commented):
Hi! Thanks for the contribution and for fixing this regression!

 I was wondering what your use case was for the `CUSTOM_AUTHORIZATION_TOKEN` env var? I feel like it's more explicit to set it at the endpoint level, but curious if you have a use case for it. 

I'm trying to reduce the number of top level env vars we currently have, but if there's a good reason to add it, let me know, otherwise LGTM! :grin:

nsarrazin:
@galen-ft I took the liberty of pushing to your branch a few fixes to endpoints, as well as fixing the linting. I removed `CUSTOM_AUTHORIZATION_TOKEN` so the custom header is only set using the endpoint `authorization` key. 

 (It was a bit confusing to have a top level env var because the setting only works for tgi endpoints)

Let me know if it still works as you want it to! Otherwise feel free to let me know

galen-ft:
Hi @nsarrazin , and thanks a lot for looking into this!

I wanted to fix the linting earlier, but something else came up.

Your changes to the endpointTgi look very good. In particular, I like the succinct way in which you add the `Authorization` header.

I agree to the argument about `CUSTOM_AUTHORIZATION_TOKEN`.

To address your points, I pushed [a commit](https://github.com/galen-ft/chat-ui/commit/206f272cc38d3c663a0afa3f157bd62db575279d) which removes:

- the section on API key or other that I previously added to the Readme as it is no longer needed
- a comment in endpointTgi.ts which is no longer needed
- `CUSTOM_AUTHORIZATION_TOKEN` from .env

I hope the changes are all right with you. If not let me know and I'll have a look.

Otherwise authorisation works as expected!

nsarrazin:
Hi, apologies for the delay with this. Thanks for cleaning things up, I'll merge it now as it looks good to me so far and thanks for the contribution! :grin:

PR #582, opened 2023-11-25
Group conversations by date + small nav fixes

nsarrazin:
* Fixed a bug where we had a double gradient on mobile nav (one from mobile nav and one from nav menu)
* Fixed a mobile nav bug where we had rounded corners at the top of the nav on mobile (even though it was fullscreen)
* Tweaked nav a bit (tighter spacing, more contrast)
* Tweaked gradients on mobile (it was a bit strange having a left gradient on a fullscreen element)
* Added conversation grouping based on dates

| Old Nav | Nev Nav |
| image  | image  |


nsarrazin:
cc @gary149 the exact styling can still be tweaked but wdyt of grouping conversations like this?

src/lib/components/NavMenu.svelte
```diff
@@ -15,17 +15,49 @@
...
 	export let user: LayoutData["user"];
 
 	function handleNewChatClick() {
 		isAborted.set(true);
 	}
+
+	// group conversations based on conv.updatedAt property which is a date
+	// it should have the following groups: "Today", "This week", "This month", "Older"
+
+	$: groupedConversations = conversations.reduce(
```
gary149, line 35: Reduce is too complex here imo, I would just use filter, something like `.filter(isToday)` in the view

gary149 (approved):

PR #591, opened 2023-11-28
Reworked settings menu

nsarrazin:
First draft of the new settings menu.
## Main changes
The settings now uses [a store accessible with a hook](https://github.com/huggingface/chat-ui/pull/591/files#diff-5bc47293107f026168afc22ce02b09ab1c5a83d9a8b758de496a4f8e6178f48a), that automatically sends updates to the settings endpoint (with a debounce). This makes changing the settings and subscribing to changes very easy. Alternatively if we don't want the settings to autosave it would be easy to tweak it to support a confirm button

The new settings menu are also implemented using a path so for example the user settings are accessible here: http://localhost:5173/chat/settings

And model specific settings are accessible using a different URL:  http://localhost:5173/chat/settings/meta-llama/Llama-2-70b-chat-hf 

## Screenshots
image
image

## Dependencies
I added an extra dep for the gear icon, make sure to run `npm install` if testing this PR locally

## Todo
- [x] make it look good on mobile
- [x] do a second pass on how links are displayed
- [x] fix bug with empty preprompt resetting to default every time


src/lib/components/Switch.svelte
```diff
@@ -3,7 +3,13 @@
 	export let name: string;
 </script>
 
-<input bind:checked type="checkbox" {name} class="peer pointer-events-none absolute opacity-0" />
+<input
+	bind:checked
+	type="checkbox"
+	{name}
+	id={name}
```
nsarrazin, line 10: The id added here breaks the functionality of the switch for me. If it was added for accessibility, could we use `aria-label` instead? @gary149

PR #612, opened 2023-12-06
Create .env.local

Th3bull1990:
Checa


nsarrazin:
@Th3bull1990 you opened a PR with your .env.local inside, which contains your HF token and your DB secrets. 

I'm going to close it now to give it less visibility and I recommend you invalidate those credentials ASAP :sweat_smile:

PR #613, opened 2023-12-06
Add `chat-ui` and `chat-ui-db` docker images 

nsarrazin:
The goal is to end up with a docker image that we can build in CI, that will not contain any secrets. Then users will be able to just `docker run chat-ui:latest` and mount their `.env.local` either as a bind volume or as an env variable. 

## To-do
 - [x] Add an image with mongo included
 - [x] Add a github action for building & publishing the docker image
 ~~Update the readme with steps to run chat-ui using docker only~~ (I will do this in a later PR once i'm sure everything works)

### Building the image
(this step won't be needed in the future, as we will build the image in our CI pipeline)

Building the image **without db** ( `chat-ui`):
```
docker build -t chat-ui  -f ./Dockerfile.local .
```
Building the image **with db** (`chat-ui-db`):

```
docker build -t chat-ui-db --build-arg="INCLUDE_DB=true" -f ./Dockerfile.local .
```

The `chat-ui-db` image will override any `MONGODB_URL` set and will use the included DB instead.

### Running the image

Simplest possible run, DB included & default config:
```
docker run -p 3000:3000 chat-ui-db
```

#### Custom config
You can specify a DOTENV_LOCAL by passing an env variable at `docker run` time:
```
DOTENV_LOCAL=$(<.env.local)  docker run --rm -e DOTENV_LOCAL -p 3000:3000 chat-ui
```

Or you can use a bind volume pointing to `.env.local` directly
```
docker run --rm --mount type=bind,source="$(pwd)/.env.local",target=/app/.env.local -p 3000:3000 chat-ui
```


gary149:
Another question, isn't it better to have a single container for each process (and use with Docker compose)?

nsarrazin:
> Question, isn't it better to have a single container for each process (and use with Docker compose)?

That's why if `USE_LOCAL_DB` is not set to true when running, it uses the db address specified in the `.env.local`.

That way you can use another container if you want or a remote DB

nsarrazin:
> > ⚠️ this PR moves the config stuff to a subfolder config so you might need to move all your .env* to config/.env*
> 
> I think we can revert this and keep it as today.

8751c6b changes it back!

mdetournay:
@nsarrazin I was hosting this app from commit tag: 118aa57bf4f9c77ddb82880ad1cbaf453cfd9c6a (end of November) 

However with the latest main branch as of today, the app doesn't read from the .env.local file anymore that I provide (so it doesn't overwrite variables in the .env file during CI/CD). 

I'm building with Dockerfile (not Dockerfile.local) that you provided here. 

Do you know what this could be due to ? 

Thanks a lot for the help :-)

PR #636, opened 2023-12-15
First commit adding Attach file button .

riseandshine99:


mikelfried:
@riseandshine99 disable your hf token you added in .env file

gary149:
Yes you exposed your token here, you should delete it from https://huggingface.co/settings/tokens

nsarrazin:
Closing so it's less visible

PR #639, opened 2023-12-15
Assistants feature

nsarrazin:
### New assistant 
image

### Assistant (created by you)
image

### Assistant (created by another user)
image

### Conversation starting page
image
### Ongoing conversation
(note top of page indicator and avatar in side menu)
image

### Receiving a shared assistant
image

### Assistant without an avatar
image


hungryalgo:
This feature reminds me of the Agents feature that you were doing, but this seems very different without tools usage.

Why is the Agents feature dropped? Or why is this one having more precedence? 
Both seem very interesting, I am guessing that this Assistants feature is more important, but not sure why.

https://github.com/huggingface/chat-ui/pull/462

thanks!

nsarrazin:
This is ready for review @gary149, it's only missing the edit an existing assistant feature but I don't think it's a blocker and the rest is done

mishig25:
On the create `new assistant modal`, there is this cutoff text `Cli...`

image

mishig25:
Maybe in the start messages, as soon as message gets empty, delete that emoty item from the list?

image

mishig25:
For the system prompts, maybe we can create and link to an example dataset of 10 system prompts for the inspiration. Something like  https://huggingface.co/datasets/fka/awesome-chatgpt-prompts

image

mishig25:
Once assistant is created, you can't edit the system prompt. Is it expected?

image

src/routes/assistant/[assistantId]/+page.svelte
```diff
@@ -0,0 +1,97 @@
...
+	const settings = useSettingsStore();
+</script>
+
+<div
+	class="fixed inset-0 flex items-center justify-center bg-black/80 backdrop-blur-sm dark:bg-black/50"
+>
+	<dialog
+		open
+		use:clickOutside={() => {
+			if (browser) window;
```
Comment (mishig25) [inline src/routes/assistant/[assistantId]/+page.svelte:30]: what's the purpose of this line?

src/routes/assistant/[assistantId]/+page.server.ts
```diff
@@ -0,0 +1,11 @@
+import { collections } from "$lib/server/database.js";
+import { ObjectId } from "mongodb";
+
+export const load = async ({ params }) => {
+	const assistant = await collections.assistants.findOne({
```
Comment (mishig25) [inline src/routes/assistant/[assistantId]/+page.server.ts:5]: what happens if `assistant` is null. I.e. `assistant` with `params.assistantId` does not exist?

src/routes/conversation/+server.ts
```diff
@@ -49,14 +49,26 @@ export const POST: RequestHandler = async ({ locals, request }) => {
...
+		if (values.assistantId) {
+			const assistant = await collections.assistants.findOne({
+				_id: new ObjectId(values.assistantId),
+			});
+			return assistant?.preprompt;
+		} else if (values.preprompt) {
+			return values.preprompt;
+		} else {
+			return model?.preprompt;
+		}
```
mishig25, line 62:
```suggestion
		} else {
			return values?.preprompt ?? model?.preprompt;
		}
```

haven't tested. But I think we can just simplify it

src/routes/assistant/[assistantId]/+page.server.ts
Comment (nsarrazin) [reply src/routes/assistant/[assistantId]/+page.server.ts:5]: added a check ! it redirects to the home settings now

src/routes/assistant/[assistantId]/+page.svelte
Comment (nsarrazin) [reply src/routes/assistant/[assistantId]/+page.svelte]: the modal has a check to return to the last stored page if clicked outside.

but uuuuh
```
			if (browser) window;
```
i can't remember why this was in here so I got rid of it, can't click on server-side anyway 😅

src/routes/conversation/+server.ts
nsarrazin: good catch thanks!

nsarrazin:
> On the create `new assistant modal`, there is this cutoff text `Cli...`
> 
> image

I made a better looking avatar upload icon, feel free to try it now!

nsarrazin:
> Once assistant is created, you can't edit the system prompt. Is it expected?
> 
> image

Added the edit feature just now!

src/lib/types/SharedConversation.ts
```diff
@@ -10,4 +10,5 @@ export interface SharedConversation extends Timestamps {
 	title: string;
 	messages: Message[];
 	preprompt?: string;
+	assistantId?: string;
```
mishig25, line 13: this is not of type `Assistant["_id"]` ?

src/lib/components/NavConversationItem.svelte
```diff
@@ -7,15 +7,18 @@
 	import CarbonTrashCan from "~icons/carbon/trash-can";
 	import CarbonClose from "~icons/carbon/close";
 	import CarbonEdit from "~icons/carbon/edit";
+	import { useSettingsStore } from "$lib/stores/settings";
 
-	export let conv: { id: string; title: string };
+	export let conv: { id: string; title: string; assistantId?: string; avatarHash?: string };
```
mishig25, line 13: can we use the type from https://github.com/huggingface/chat-ui/blob/feature/assistants/src/lib/components/NavMenu.svelte#L11 ?

nsarrazin: good catch I made the type reusable

src/routes/assistant/[assistantId]/+page.server.ts
Comment (mishig25) [reply src/routes/assistant/[assistantId]/+page.server.ts:5]: or is it guranteed to be non-null all the time?

src/lib/types/SharedConversation.ts
nsarrazin: good catch, fixed in https://github.com/huggingface/chat-ui/pull/639/commits/41a5971f1d41e088a5a0f077c0fe9ed2b5a9ed59

src/routes/settings/assistants/new/+page.server.ts
```diff
@@ -0,0 +1,106 @@
...
+import { authCondition, requiresUser } from "$lib/server/auth";
+import { collections } from "$lib/server/database";
+import { fail, type Actions, redirect } from "@sveltejs/kit";
+import { ObjectId } from "mongodb";
+
+import { z } from "zod";
+import sizeof from "image-size";
+import { sha256 } from "$lib/utils/sha256";
+
+const newAsssistantSchema = z.object({
```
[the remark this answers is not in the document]
nsarrazin: wdyt of 

```ts
const newAsssistantSchema = z.object({
	name: z.string().min(1).max(50),
	modelId: z.string().min(1).max(100),
	preprompt: z.string().min(1).max(1024),
	description: z.string().max(1024).optional(),
	exampleInput1: z.string().max(256).optional(),
	exampleInput2: z.string().max(256).optional(),
	exampleInput3: z.string().max(256).optional(),
	exampleInput4: z.string().max(256).optional(),
	avatar: z.instanceof(File).optional(),
});

for those values ?

File: src/routes/assistant/[assistantId]/+page.server.ts (hunk 2 above)
Comment (nsarrazin) [reply src/routes/assistant/[assistantId]/+page.server.ts:5]: There is a check [here](https://github.com/huggingface/chat-ui/blob/f84c86a6bfba04137c88a6264233061d9842fad4/src/routes/settings/assistants/%5BassistantId%5D/%2Bpage.ts#L10) that redirects if it's null

File: .env [hunk 7]
```diff
@@ -109,4 +109,6 @@ LLM_SUMMERIZATION=true
 # PUBLIC_APP_COLOR=yellow
 # PUBLIC_APP_DESCRIPTION="Making the community's best AI chat models available to everyone."
 # PUBLIC_APP_DATA_SHARING=1
-# PUBLIC_APP_DISCLAIMER=1
\ No newline at end of file
+# PUBLIC_APP_DISCLAIMER=1
+
+DISABLE_ASSISTANTS=false # set to true to disable assistants feature
```
gary149, line 114: should be disabled by default?
```suggestion
ENABLE_ASSISTANTS=false # set to true to enable this feature
```

nsarrazin: The idea to make it enabled by default was that users of chat-ui would get the latest feature without having to change their `.env.local` bu it's still toggleable if users don't want it. wdyt?

gary149: I would keep it as simple as possible by default (so disabled).

nsarrazin: Handled in 31a4941adee529bfd1a9fc8747d8af21b7797a8f

src/routes/settings/assistants/new/+page.server.ts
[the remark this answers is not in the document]
mishig25: ballparking the values by testing on python
```py
a = "a"
print(a*1024) # or any other number
```

.env
```diff
@@ -99,6 +99,7 @@ PARQUET_EXPORT_SECRET=
 RATE_LIMIT= # requests per minute
 MESSAGES_BEFORE_LOGIN=# how many messages a user can send in a conversation before having to login. set to 0 to force login right away
 
+APP_BASE=""
```
mishig25, line 102: could you add a comment on what's `APP_BASE` ?

src/lib/components/DisclaimerModal.svelte
```diff
@@ -37,8 +37,9 @@
 					class:text-gray-800={$page.data.loginEnabled}
 					class:hover:bg-slate-100={$page.data.loginEnabled}
 					on:click={(e) => {
+						e.preventDefault();
+						e.stopPropagation();
```
mishig25, line 41:
```suggestion
					on:click|preventDefault|stopPropagation={(e) => {
```

haven't tested it. But is there a reason not to use [svelte event modifiers](https://svelte.dev/docs/element-directives#on-eventname)?

src/lib/components/NavMenu.svelte
```diff
@@ -7,14 +7,9 @@
...
+	import type { ConvSidebar } from "$lib/types/ConvSidebar";
 
-	interface Conv {
-		id: string;
-		title: string;
-		updatedAt: Date;
-	}
-
-	export let conversations: Array<Conv> = [];
+	export let conversations: Array<ConvSidebar> = [];
```
mishig25, line 12:
```suggestion
	export let conversations: ConvSidebar[] = [];
```
nit

src/routes/assistant/[assistantId]/thumbnail.png/ChatThumbnail.svelte
```diff
@@ -0,0 +1,21 @@
...
+</script>
+
+<div class="flex h-full w-full flex-col items-center bg-gray-900 p-4 pt-12 text-white">
+	{#if avatarUrl}
+		<img class="h-24 w-24 rounded-full" src={avatarUrl} alt="avatar" />
+	{:else}
+		<div
+			class="flex h-24 w-24 items-center justify-center rounded-full bg-gray-300 font-bold text-gray-500"
+		>
+			{name[0].toLocaleUpperCase()}
```
Comment (mishig25) [inline src/routes/assistant/[assistantId]/thumbnail.png/ChatThumbnail.svelte:14]: instead of js `toLocalUpperCase`, would it be possible to use css/tailwind [uppercase](https://tailwindcss.com/docs/text-transform)?

src/routes/assistant/[assistantId]/+page.server.ts
```diff
@@ -0,0 +1,20 @@
...
+	try {
+		const assistant = await collections.assistants.findOne({
+			_id: new ObjectId(params.assistantId),
+		});
+
+		if (!assistant) {
+			throw redirect(302, `${base}`);
+		}
+
+		return { assistant: JSON.parse(JSON.stringify(assistant)) };
```
Comment (mishig25) [inline src/routes/assistant/[assistantId]/+page.server.ts:16]:
```suggestion
		return { assistant };
```

is it not possible to do so?

src/routes/settings/assistants/[assistantId]/avatar/+server.ts
```diff
@@ -0,0 +1,47 @@
+import { collections } from "$lib/server/database";
+import { error, type RequestHandler } from "@sveltejs/kit";
+import { ObjectId } from "mongodb";
+
+export const GET: RequestHandler = async ({ params }) => {
+	const assistant = await collections.assistants.findOne({
+		_id: new ObjectId(params.assistantId),
+	});
+
+	// check user
```
Comment (mishig25) [inline src/routes/settings/assistants/[assistantId]/avatar/+server.ts:10]:
```suggestion
```

src/routes/settings/assistants/new/+page.server.ts
```diff
@@ -0,0 +1,106 @@
...
+			parse?.data?.exampleInput3 ?? "",
+			parse?.data?.exampleInput4 ?? "",
+		].filter((input) => !!input);
+
+		let hash;
+		if (parse.data.avatar && parse.data.avatar.size > 0) {
+			const dims = sizeof(Buffer.from(await parse.data.avatar.arrayBuffer()));
+
+			if ((dims.height ?? 1000) > 512 || (dims.width ?? 1000) > 512) {
+				const errors = [{ field: "avatar", message: "Avatar too big" }];
```
mishig25, line 80: we can provide more informational error msg: `Avatar is too big. Please make sure the size of your avatar is no bigger than ABC width x XYZ heigth`

src/routes/settings/assistants/[assistantId]/+page.server.ts
```diff
@@ -0,0 +1,115 @@
...
+		const fileCursor = collections.bucket.find({ filename: assistant._id.toString() });
+
+		// Step 2: Delete the existing file if it exists
+		let fileId = await fileCursor.next();
+		while (fileId) {
+			await collections.bucket.delete(fileId._id);
+			fileId = await fileCursor.next();
+		}
+
+		throw redirect(302, `${base}/settings`);
```
Comment (mishig25) [inline src/routes/settings/assistants/[assistantId]/+page.server.ts:50]: is it correct that it is `throw` rather than `return`?

src/routes/+layout.server.ts
```diff
@@ -58,24 +64,52 @@ export const load: LayoutServerLoad = async ({ locals, depends }) => {
...
+		conversations: conversations.map((conv) => ({
+			id: conv._id.toString(),
+			title: settings?.hideEmojiOnSidebar ? conv.title.replace(/\p{Emoji}/gu, "") : conv.title,
+			model: conv.model ?? defaultModel,
+			updatedAt: conv.updatedAt,
+			assistantId: conv.assistantId?.toString(),
+			avatarHash:
+				conv.assistantId &&
+				assistants.find((a) => a._id.toString() === conv.assistantId?.toString())?.avatar,
+		})) satisfies Array<ConvSidebar>,
```
mishig25, line 112:
```suggestion
		})) satisfies ConvSidebar[],
```

gururise:
With the OpenAI GPT store launching next week, this couldn't come at a better time! Looking forward to this feature!

src/lib/components/DisclaimerModal.svelte
nsarrazin: Reason was that I forgot about them haha, thanks for catching it, I'll fix it everywhere.

src/routes/assistant/[assistantId]/thumbnail.png/ChatThumbnail.svelte
Comment (nsarrazin) [reply src/routes/assistant/[assistantId]/thumbnail.png/ChatThumbnail.svelte]: Yep, good catch, fixed it everywhere

src/routes/assistant/[assistantId]/+page.server.ts
Comment (nsarrazin) [reply src/routes/assistant/[assistantId]/+page.server.ts:16]: The main reason why this is a thing is that mongodb returns SOME `ObjectId` for the IDs which are not happy being serialized by sveltekit. If I remove this line I get the following error:

```
Error: Data returned from `load` while rendering /assistant/[assistantId] is not serializable: Cannot stringify arbitrary non-POJOs (data.assistant._id)
``` 

If you know a better way to handle this let me know, I would happily get rid of this line :grin:

src/routes/settings/assistants/new/+page.server.ts
nsarrazin: Good catch, I added a clearer error message. We [resize on the client side](https://github.com/huggingface/chat-ui/blob/51b4354884f6b931b201c15449ed22daa5e29aca/src/routes/settings/assistants/new/%2Bpage.svelte#L42-L51) so this should never be seen but doesn't hurt to have a better message

src/lib/components/LoginModal.svelte
```diff
@@ -46,9 +46,8 @@
 			{:else}
 				<button
 					class="flex w-full items-center justify-center whitespace-nowrap rounded-full border-2 border-black bg-black px-5 py-2 text-lg font-semibold text-gray-100 transition-colors hover:bg-gray-900"
-					on:click={(e) => {
+					on:click|preventDefault={() => {
 						if (!cookiesAreEnabled()) {
-							e.preventDefault();
```
mishig25, line 51: this line 
```js
if (!cookiesAreEnabled()) {
   e.preventDefault();
```
was replaced to using svelte modifier `on:click|preventDefault`, which assumes that `e.preventDefault();` was outside the `if` statement like
```js
 e.preventDefault();
if (!cookiesAreEnabled()) {
  ...
```

Could you confirm that this difference in logic doesn't cause any unexpected bahviour?

src/routes/assistant/[assistantId]/+page.server.ts
Comment (mishig25) [reply src/routes/assistant/[assistantId]/+page.server.ts:16]: oh I see

src/routes/+layout.server.ts
```diff
@@ -59,24 +65,52 @@ export const load: LayoutServerLoad = async ({ locals, depends }) => {
 
 	const loginRequired = requiresUser && !locals.user && userHasExceededMessages;
 
+	const disableAssistants = !(ENABLE_ASSISTANTS === "true");
```
mishig25, line 68:
```suggestion
	const disableAssistants = ENABLE_ASSISTANTS !== "true";
```

was there reason not to use like this?

src/lib/components/LoginModal.svelte
nsarrazin: it works fine from testing!

nsarrazin: I reverted it back, just in case it doesn't fully work!

nsarrazin:
I think we need at least #678 reviewed & merged back into this branch, and then maybe #662 would also be nice!

src/routes/+layout.server.ts
mishig25: boolean logic can be tricky sometimes depending on the truthiness of different types. So I thought maybe there might be some catch 🤔

nsarrazin: i think we just went back and forth on whether the flag should enable or disable assistants so I just put a `!` in front of the whole thing without thinking 😆

src/routes/settings/+layout.svelte
```diff
@@ -66,14 +67,56 @@
 					{/if}
 				</a>
 			{/each}
+			{#if !data.disableAssistants}
```
mishig25, line 70:
```suggestion
			{#if data.enableAssistants}
```

src/routes/+layout.server.ts
```diff
@@ -59,33 +65,61 @@ export const load: LayoutServerLoad = async ({ locals, depends }) => {
 
 	const loginRequired = requiresUser && !locals.user && userHasExceededMessages;
 
+	const disableAssistants = ENABLE_ASSISTANTS !== "true";
```
mishig25, line 68: to go with [here](https://github.com/huggingface/chat-ui/pull/639/files#r1445950232)
```suggestion
	const enableAssistants = ENABLE_ASSISTANTS === "true";
```

src/routes/+layout.server.ts
```diff
@@ -120,10 +155,13 @@ export const load: LayoutServerLoad = async ({ locals, depends }) => {
 		})),
 		oldModels,
 		user: locals.user && {
+			id: locals.user._id.toString(),
 			username: locals.user.username,
 			avatarUrl: locals.user.avatarUrl,
 			email: locals.user.email,
 		},
+		assistant,
+		disableAssistants,
```
mishig25, line 164:
```suggestion
		enableAssistants,
```

nsarrazin:
here is how it looks, tried it with a lot of text and a small text

### maxed out content 
image
### short
image
### no picture
image

wdyt @gary149 ?

nsarrazin:
Merged back #662 into this branch. There's a feature flag if we think it's not a feature we want in prod yet 😄

gururise:
Testing this PR and I can't seem to create a "new assistant" as the option appears to be missing:
image

I did add the following to .env.local:
ENABLE_ASSISTANTS=true

nsarrazin:
@gururise thanks for testing! By any chance do you have any openid config ? We've configured it so that only logged-in users can create assistants if openid is configured. But it should be accessible if you have no login setup. I'll have a look as well.

gururise:
> @gururise thanks for testing! By any chance do you have any openid config ? We've configured it so that only logged-in users can create assistants if openid is configured. But it should be accessible if you have no login setup. I'll have a look as well.

I only tested without openid. I'll try with openid.

gururise:
Just tried with OpenID and it works.  One issue is that the "created by user" is always **null** when using OpenID from google. If we don't have the user's name, could we just use their email address instead?

Are the assistants shared across all logins?

nsarrazin:
Ok then I need to fix it so that it also works without OpenID! :grin: And good catch on the created by, will see if we can default to something else. 

Assistants are public but not shared by default, (you can share them using the URL in the settings)

gururise:
> Should be better @gururise 😄

Works perfectly now! Thanks!

SvCy:
Really glad to see that it's finally up! 🤗

PR #641, opened 2023-12-18
Implemend PDF-chat feature

mishig25:


nsarrazin:
:fire: So cool, will test it locally later, but just from looking at the demo, do you think there's an easy way to show an indicator of when a PDF is already uploaded and will be sent with the message ?

mishig25:
> but just from looking at the demo, do you think there's an easy way to show an indicator of when a PDF is already uploaded and will be sent with the message ?

at the moment, there is websearch-like box that indicates pdf rag was used

image

nsarrazin:
I meant more when the file is loaded and before the conversation is started, like for images:
image

I guess it would look a bit different since you can only have one PDF per conversation, but it would be nice to have an indication that a PDF will be used to answer the query :eyes:

mishig25:
I see. Let me think about it

src/routes/conversation/[id]/upload-pdf/+server.ts
```diff
@@ -0,0 +1,41 @@
...
+		throw error(404, "Conversation not found");
+	}
+
+	const formData = await request.formData();
+	const file = formData.get("pdf"); // 'pdf' is the name used in FormData on the frontend
+	const data = new Uint8Array(await file.arrayBuffer());
+	const loadingTask = pdfjsLib.getDocument({ data });
+	const pdf = await loadingTask.promise;
+
+	const N_MAX_PAGES = 20;
```
Comment (gary149) [inline src/routes/conversation/[id]/upload-pdf/+server.ts:27]: This seems a bit low, 100 or 200 pages maybe?

src/lib/buildPrompt.ts
```diff
@@ -19,6 +21,7 @@ export async function buildPrompt({
 	messages,
 	model,
 	webSearch,
+	pdfSearch,
```
gary149, line 24: pdfSearch should be enabled/disabled from an env var.

src/lib/components/UploadBtn.svelte
```diff
@@ -1,23 +1,72 @@
...
+		on:change={onChange}
 		class="absolute w-full cursor-pointer opacity-0"
 		type="file"
-		accept="image/*"
+		{accept}
+		disabled={uploading}
 	/>
-	<CarbonUpload class="mr-2 text-xs " /> Upload image
+	<CarbonUpload class="mr-2 text-xs " />
+	{#if uploadPdfStatus === PdfUploadStatus.Uploaded}
```
gary149, line 65: Yes as already said, you probably want to display uploaded files somewhere in the UI.

src/routes/conversation/[id]/upload-pdf/+server.ts
Comment (mishig25) [reply src/routes/conversation/[id]/upload-pdf/+server.ts:27]: I see. 100 or 200 would be bit slow to create embeddings on CPU using the current solution of transformers.js (I will provide actual benchmark numbers). 

In this case, should I review/push for the community PR https://github.com/huggingface/chat-ui/pull/646 that makes it possible to have an embedding endpoint for faster embeddings creation. We can still use tfjs embeddings for websearch, and use TEI-powered embeddings endpoint for PDF & possible for assistants #639 (if users can upload documents). Wdyt ?

src/routes/+page.svelte
```diff
@@ -51,14 +45,34 @@
...
 </script>
 
 <svelte:head>
 	<title>{PUBLIC_APP_NAME}</title>
 </svelte:head>
 
 <ChatWindow
-	on:message={(ev) => createConversation(ev.detail)}
+	on:message={(ev) => createConversationWithMsg(ev.detail)}
+	on:uploadpdf={(ev) => createConversationWithPdf(ev.detail)}
```
nsarrazin, line 75: I'm not sure why we create an empty chat with a document, seems to me like it should be done like for images, where the files are "stored" in the front-end and the conversation created along when the first message is sent ?

src/routes/conversation/[id]/upload-pdf/+server.ts
```diff
@@ -0,0 +1,41 @@
...
+		...authCondition(locals),
+	});
+
+	if (!conversation) {
+		throw error(404, "Conversation not found");
+	}
+
+	const formData = await request.formData();
+	const file = formData.get("pdf"); // 'pdf' is the name used in FormData on the frontend
+	const data = new Uint8Array(await file.arrayBuffer());
```
Comment (nsarrazin) [inline src/routes/conversation/[id]/upload-pdf/+server.ts:23]:
```suggestion
	if (!file || typeof file === "string") {
		throw error(400, "No file provided");
	}
	const data = new Uint8Array(await file.arrayBuffer());
```

Check that pdf is indeed a file (for type checks)

src/routes/conversation/[id]/upload-pdf/+server.ts
```diff
@@ -0,0 +1,41 @@
...
+	const loadingTask = pdfjsLib.getDocument({ data });
+	const pdf = await loadingTask.promise;
+
+	const N_MAX_PAGES = 20;
+	let text = "";
+	for (let i = 1; i <= Math.min(pdf.numPages, N_MAX_PAGES); i++) {
+		const page = await pdf.getPage(i);
+		const content = await page.getTextContent();
+		text += content.items.map((item) => item.str).join(" ");
```
Comment (nsarrazin) [inline src/routes/conversation/[id]/upload-pdf/+server.ts:32]:
```suggestion
		text += content.items.map((item) => (item as { str?: string }).str ?? "").join(" ");
```

just type checks

src/lib/components/chat/ChatWindow.svelte
```diff
@@ -15,24 +15,26 @@
...
 	export let loading = false;
 	export let pending = false;
 	export let shared = false;
 	export let currentModel: Model;
 	export let models: Model[];
-	export let webSearchMessages: WebSearchUpdate[] = [];
+	export let RAGMessages: RAGUpdate[] = [];
 	export let preprompt: string | undefined = undefined;
 	export let files: File[] = [];
+	export let uploadPdfStatus: PdfUploadStatus;
```
nsarrazin, line 37:
```suggestion
	export let uploadPdfStatus: PdfUploadStatus | undefined = undefined;
```

fix the missing uploadPdfStatus in type checks

src/lib/components/UploadBtn.svelte
```diff
@@ -1,23 +1,72 @@
 <script lang="ts">
+	import { PdfUploadStatus } from "$lib/types/PdfChat";
+	import { createEventDispatcher, onDestroy } from "svelte";
 	import CarbonUpload from "~icons/carbon/upload";
 
 	export let classNames = "";
+	export let multimodal = false;
 	export let files: File[];
-	let filelist: FileList;
+	export let uploadPdfStatus: PdfUploadStatus;
```
nsarrazin, line 10:
```suggestion
	export let uploadPdfStatus: PdfUploadStatus | undefined = undefined;
```

fix the missing uploadPdfStatus in type checks

nsarrazin (commented):
The feature itself works super well, I'm super impressed by what it can do tbh :fire: 

There's just a couple things that are not super clear to me around the user experience that I think could be improved:

* Why does it create an empty conversation on file upload ? Seems to me like it should show that a PDF has been uploaded client-side, and create the conversation with the PDF only when the first message is sent, the way we deal with images 
* Would be cool to have an indicator in a conversation that shows that a PDF is available in a specific conversation. Like something at the top of the conversation that says `"${filename}.pdf is available in this conversation"` or something 
* It's not super clear from the UI perspective what happens when you upload a PDF to a conversation that already has one. I think it silently replaces the old PDF by the new one, but it's a bit confusing, I think we could just make it so that PDFs can only be uploaded at the beginning of a conversation, wdyt?

I also left a couple of fixes for type checking as comment/suggestions in the PR

nsarrazin:
Small note, if I drag and drop a non-pdf file I get the following weird output
Screenshot from 2023-12-21 13-40-59

src/routes/conversation/[id]/upload-pdf/+server.ts
```diff
@@ -0,0 +1,41 @@
...
+	});
+
+	if (!conversation) {
+		throw error(404, "Conversation not found");
+	}
+
+	const formData = await request.formData();
+	const file = formData.get("pdf"); // 'pdf' is the name used in FormData on the frontend
+	const data = new Uint8Array(await file.arrayBuffer());
+	const loadingTask = pdfjsLib.getDocument({ data });
```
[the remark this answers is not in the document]
Comment (mishig25) [reply src/routes/conversation/[id]/upload-pdf/+server.ts]: indeed https://github.com/huggingface/chat-ui/pull/641/commits/5d4939c02fa29f0e35fb8526c412aca3d8dc0cf3

mishig25:
Updates:
1. Fixed this upload non-img file bug [here](https://github.com/huggingface/chat-ui/pull/641#issuecomment-1866372292)
2. Provided better UI/UX for uploaded file (see attached video). Specifically: 1. name if the uploaded pdf with pdf icon appears; 2. this pdf file name and icon does "blinking" animation while the pdf is being uploaded & embeddings are being created; 3. on hover `x` btn appears, that let's you delete the uploaded PDF file
3. env var (config) for enabling pdf-chat feature as [here](https://github.com/huggingface/chat-ui/pull/641/commits/e9ffdab5b982fa21b0064e0e635f071ee84d4725)


https://github.com/huggingface/chat-ui/assets/11827707/723cd355-edea-4204-98db-0632546b3cf4

wdhorton:
Thanks for working on this! One question I had is: what was your thought process for adding a new upload button for PDFs, versus using the existing drag-and-drop functionality that already exists for images?

mishig25:
@wdhorton currently the UI might still evolve. For now, the reason for resuing the same upload btn (instead of adding a new upload btn) is: having two different upload btns makes the UI look cluttered, especially on smaller screens

itaybar:
I have few questions:
* Why you are limiting this feature to PDF and not csv, txt, etc?
* I don't sure you have to use embedding for PDF (or at least make it optional, and not mandatory)
*  Why using Mongo as vectorDB and calculate the vector similarity in client side instead of using some real vectorDB that can to it whey more efficient and fast than running it in JS, this can make the code cleaner and remove any limitation of content size
* Does storing all the embedding is good idea? this can make the DB blow relatively fast.

src/lib/components/UploadBtn.svelte
mishig25: handled in https://github.com/huggingface/chat-ui/pull/641#issuecomment-1883017855

mishig25:
Update: this PR is getting big. Unfortunately, there is no other option (I think). The specific points are:
1. In #689, I've generalized RAG. What does it mean? It means two things: 1. server side. 2. frontend side. In terms of `1. server side`, RAG applications have to implement a RAG interface for consistency & better organization of codebase (you can checkout directory `src/lib/servrer/rag`). In terms of `2. server side`, OpenWebSearchResults svelte component is generalized into OpenRAGResults svelte component that will show up on RAG augmented messages (as suggested [here](https://github.com/huggingface/chat-ui/pull/641#discussion_r1434012545)).
2. Besides creating embeddings through [TEI for pdf-chat](https://github.com/huggingface/chat-ui/pull/646#issuecomment-1882871832), we would need vectorDB support for multiple reasons:
   1. without vectorDB, the pdf-chat session will lose the pdf embeddings (for instance, when you close your browser & re-open the same chat-ui conversation, the pdf embeddings will no longer be available)
   2. storing PDF embeddings on [mongo gridFS](https://www.mongodb.com/docs/manual/core/gridfs/) would slow-down performance (as questioned [here](https://github.com/huggingface/chat-ui/pull/641#issuecomment-1884693746)) & large number of embeddings can cause a lot of latency on the server since [findSimilarSentences](https://github.com/huggingface/chat-ui/blob/chatPDF/src/lib/server/embeddings.ts#L31-L32) runs locally on the server. Therefore, we would need support for vectorDB.
   3. We would need vectorDB for other features as well. For instance, we would need vecorDB + pdf RAG for #639. There was also internal slack discussion [here](https://huggingface.slack.com/archives/C055NNJF98Q/p1704892458918669?thread_ts=1704811522.875959&cid=C055NNJF98Q).
   4. for the info, checking [openai/chatgpt-retrieval-plugin](https://github.com/openai/chatgpt-retrieval-plugin) to see if we would need to follow commonly used API for vectorDBs 

wdyt @nsarrazin @gary149

itaybar:
> @itaybar, thanks a lot [for your questions](https://github.com/huggingface/chat-ui/pull/641#issuecomment-1884693746)
> 
> > Why you are limiting this feature to PDF and not csv, txt, etc?
> 
> Yes, we will add support to other text files. Once this PR is done, supporting other text files would be trivial. (Might even include as part of this PR)
> 
> > I don't sure you have to use embedding for PDF (or at least make it optional, and not mandatory)
> 
> Could you elaborate on it. And what would be the alternatives?
> 
> > Why using Mongo as vectorDB and calculate the vector similarity in client side instead of using some real vectorDB that can to it whey more efficient and fast than running it in JS, this can make the code cleaner and remove any limitation of content size. Does storing all the embedding is good idea? this can make the DB blow relatively fast.
> 
> Indeed this is a good point. I/we will add support for vectorDB (likely part of this PR)

Thanks for the quit response guys!
About the optional embedding for the pdf, correct me if I wrong, but you can just read the texts from the pdf without using embedding for the images,  etc and by that removing the hard limits for content file

itaybar:
@mishig25 What is your estimation for merging this in your opinion?

mishig25:
Can't give exact date. But will do my best to merge it soon :)

src/routes/conversation/[id]/+server.ts
```diff
@@ -232,13 +233,25 @@ export async function POST({ request, locals, params, getClientAddress }) {
...
-			messages[messages.length - 1].webSearch = webSearchResults;
+			messages[messages.length - 1].ragContext = webSearchResults;
+
+			let pdfSearchResults: RagContext | undefined;
+			const pdfSearch = await collections.files.findOne({ filename: `${convId.toString()}-pdf` });
+			if (pdfSearch) {
+				pdfSearchResults = await RAGs["pdfChat"].retrieveRagContext(conv, newPrompt, update);
+			}
+
+			messages[messages.length - 1].ragContext = pdfSearchResults;
```
Comment (zzen0008) [inline src/routes/conversation/[id]/+server.ts:254]: This causes a bug with web search.
messages[messages.length - 1].ragContext = webSearchResults is being overridden with  undefined if no pdf document has been uploaded by the user.

Comment (zzen0008) [reply src/routes/conversation/[id]/+server.ts:254]: Submitted a PR to fix the issue #745

johndpope:
Merge. hire 10 more people to help.

mishig25:
> Merge

Will merge soon

> hire 10 more people to help.

Hiring 10 people rarely results in 10x productivity

flexchar:
It's a tough one to implement. I appreciate the work on this one.

I had a couple of thoughts on this I'd like to share.

First, I thought of a plugin-like system that is registered based on the file type, has `handler for processing/storing` and `handler for retrieval`. That would allow community to scale while not choking the HF developers which I'm beyond impressed being able to deliver such variety of products.

Alternatively, it could also be a third party API - much like OpenAI function calling works - so that the responsibility is NOT on you but on the end user who chooses to deploy. It's great for developers but would probably be a pain for the those who just one to feel the power of deploying and has no use beyond (inspired by the shutdown story of banana.dev).

I believe these would inherently fit better as the nature of open-source deployments is to customize. As such, there is an infinite number of use cases and solutions... _PDFs, images, audio files, web search_ as `input`; _ChromaDB, Pinecone, Qdrant, PGVector, Meilisearch_ as `storage/retrieval` to name a few.

zubu007:
Hello. I have cloned the "chatPDF" branch to use the pdf upload feature. It works locally on my machine when I run `npm run dev`. However, when I perform the same thing on my apache2 server, it shows an uploading PDF error. I attaching the screenshot. 
image

I have tried copying the exact .env.local file from my local directory to my server's directory but it still shows the same error. Should I open a new issue about it? What else can I provide to help you understand the error?

flexchar:
Can you ideally skip apache2? Or check error logs from it to see if anything is being logged? Could be that some headers are lost or file being too big and blocked. What does your network devtools say on this upload response?

@zubu007

zubu007:
The error on the console shows 403 Forbidden error meaning it has something to do with authentication. I am using custom models with separate endpoint rather than the default. However, the chat function with the custom model works as expected. When the pdf is being uploaded, it throws the 403. My question is, the pdf fetch function is using a different authentication? 

Let me add further console logs in both my local machine and server to see the difference. I will let you know the results here.

windprak:
I have the same problem as https://github.com/huggingface/chat-ui/issues/693  It retrieves correctly when looking at the prompt and parameters, but the model answers as if it has gotten only the question.

lmaosweqf1:
hey, whats the current status of the PR?

C-Loftus:
Also very interested in this feature. Is there help needed for this? Wasn't sure the blocker or how to help.

mansur-abdirimov:
@mishig25 Thanks for working on this! Very interested in this feature as well thus curious if there are any updates it?

ruizcrp:
Hi @mishig25 and all, from my side I would also offer to help in my freetime if I can, as this is quite an important feature. Reading the above I think that an overview of the required steps is important or someone that coordinates tasks. E.g. is something missing in PR https://github.com/huggingface/chat-ui/pull/745 ? Or is a redesign of the entire RAG-logic including externalization of the API required? Or is there maybe a meeting needed with some overarching architecture etc.?

antonkulaga:
I am really confused, current version of the chat doesn't seem to have the file upload despute this PR being merged. Was it deliberately deleted?

johndpope:
it's better to just have 1 employee solving the issue - especially when the company is valued at $4,500 million dollars...

bpawnzZ:
why has this not be added?

DeanLemans:
will this ever be added?

kmk142789 (approved):

Issue #637, opened 2023-12-15
Building Dockerfile failing after PR#613


galen-ft:
Hello,

Building an image from [Dockerfile](https://github.com/huggingface/chat-ui/blob/ad481ce628998761215f569c462af92b0c579040/Dockerfile#L1) is failing. The breaking change appears to have been introduced by [commit 73b316cac5a154bee40504f19e64d59eef6aa49a](https://github.com/huggingface/chat-ui/commit/73b316cac5a154bee40504f19e64d59eef6aa49a) which in turn was merged with [PR#613](https://github.com/huggingface/chat-ui/pull/613).

I use the following command to perform the build:
```bash
DOCKER_BUILDKIT=1 docker build -t $TAG -f Dockerfile .
```

It appears that there is a problem with the way `.env.local` is being read. The error message is related to `MONGODB_URL` and it is:

```bash
[...]
> [builder 3/3] RUN --mount=type=secret,id=DOTENV_LOCAL,dst=.env.local     npm run build:
#20 0.742 
#20 0.742 > chat-ui@0.6.0 build
#20 0.742 > vite build
#20 0.742 
#20 0.908 ▲ [WARNING] Cannot find base config file "./.svelte-kit/tsconfig.json" [tsconfig.json]
#20 0.908 
#20 0.908     tsconfig.json:2:12:
#20 0.908       2 │   "extends": "./.svelte-kit/tsconfig.json",
#20 0.908         ╵              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
#20 0.908 
#20 1.660 
#20 1.660 vite v4.3.9 building SSR bundle for production...
#20 1.692 transforming...
#20 4.208 "BaseClient" and "TokenSet" are imported from external module "openid-client" but never used in "src/lib/server/auth.ts".
#20 4.209 ✓ 215 modules transformed.
#20 4.292 rendering chunks...
#20 5.214 
#20 5.214 node:internal/event_target:1088
#20 5.214   process.nextTick(() => { throw err; });
#20 5.214                            ^
#20 5.216 Error: Please specify the MONGODB_URL environment variable inside .env.local. Set it to mongodb://localhost:27017 if you are running MongoDB locally, or to a MongoDB Atlas free instance for example.
```

[...]

PR #642, opened 2023-12-18
Add negating pattern to not ignore .env.local when running docker build

gterziysky:
This PR aims to fix the cause of the "Building Dockerfile failing" problem explained in [issue#637](https://github.com/huggingface/chat-ui/issues/637).

You can see the fix itself in my [comment](https://github.com/huggingface/chat-ui/issues/637#issuecomment-1857673358) to the aforementioned issue.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -8,3 +8,4 @@ node_modules/
 .svelte-kit/
 .env*
 !.env
+!.env.local


nsarrazin (approved):

nsarrazin:
Thanks for the fix!

PR #645, opened 2023-12-19
Update share behaviour

mishig25:
### TLDR: Change the current share conversation behaviour for desktop:

[internal slack link](https://huggingface.slack.com/archives/C052EKHD1U0/p1697718356502049?thread_ts=1697717286.065219&cid=C052EKHD1U0)

Currently: when you click `share conversation`, chat-ui opens [window.prompt](https://developer.mozilla.org/en-US/docs/Web/API/Window/prompt)

With this PR: just like [hf.co collection](https://huggingface.co/docs/hub/collections) sharing 
Screenshot 2023-12-19 at 3 11 59 PM

when you click `share conversation`, chat-ui saves the share url in user's clipboard and the share btn text becomes `Copied to clipboard` for 2 seconds
Screenshot 2023-12-19 at 3 11 36 PM

#### For mobile, nothing changes:

It shows the mobile share window as before

### DIFF
--- a/src/lib/components/chat/ChatWindow.svelte
+++ b/src/lib/components/chat/ChatWindow.svelte
@@ -24,7 +24,6 @@
 	import UploadBtn from "../UploadBtn.svelte";
 	import file2base64 from "$lib/utils/file2base64";
 	import { useSettingsStore } from "$lib/stores/settings";
-	import { isConversationShared } from "$lib/stores/shareConversation";
 
 	export let messages: Message[] = [];
 	export let loading = false;
@@ -41,6 +40,8 @@
 	let loginModalOpen = false;
 	let message: string;
 	let timeout: ReturnType<typeof setTimeout>;
+	let isSharedRecently = false;  
+    $: $page.params.id && (isSharedRecently = false);
 
 	const dispatch = createEventDispatcher<{
 		message: string;
@@ -77,15 +78,15 @@
 
 	const settings = useSettingsStore();
 
-	$: {
-		if ($isConversationShared) {
-			if (timeout) {
+	function onShare(){
+		dispatch("share");  
+		isSharedRecently = true;
+		if (timeout) {
 				clearTimeout(timeout);
 			}
 			timeout = setTimeout(() => {
-				$isConversationShared = false;
+				isSharedRecently = false;
 			}, 2000);
-		}
 	}
 
 	onDestroy(() => {
@@ -248,11 +249,11 @@
 					<button
 						class="flex flex-none items-center hover:text-gray-400 max-sm:rounded-lg max-sm:bg-gray-50 max-sm:px-2.5 dark:max-sm:bg-gray-800"
 						type="button"
-						class:hover:underline={!$isConversationShared}
-						on:click={() => dispatch("share")}
-						disabled={$isConversationShared}
+						class:hover:underline={!isSharedRecently}
+						on:click={onShare}
+						disabled={isSharedRecently}
 					>
-						{#if $isConversationShared}
+						{#if isSharedRecently}
 							<CarbonCheckmark class="text-[.6rem] sm:mr-1.5 sm:text-green-500" />
 							<div class="text-green-600 max-sm:hidden">Copied to clipboard</div>
 						{:else}
--- a/src/lib/components/chat/ChatWindow.svelte
+++ b/src/lib/components/chat/ChatWindow.svelte
@@ -40,8 +40,8 @@
 	let loginModalOpen = false;
 	let message: string;
 	let timeout: ReturnType<typeof setTimeout>;
-	let isSharedRecently = false;  
-    $: $page.params.id && (isSharedRecently = false);
+	let isSharedRecently = false;
+	$: $page.params.id && (isSharedRecently = false);
 
 	const dispatch = createEventDispatcher<{
 		message: string;
@@ -78,15 +78,15 @@
 
 	const settings = useSettingsStore();
 
-	function onShare(){
-		dispatch("share");  
+	function onShare() {
+		dispatch("share");
 		isSharedRecently = true;
 		if (timeout) {
-				clearTimeout(timeout);
-			}
-			timeout = setTimeout(() => {
-				isSharedRecently = false;
-			}, 2000);
+			clearTimeout(timeout);
+		}
+		timeout = setTimeout(() => {
+			isSharedRecently = false;
+		}, 2000);
 	}
 
 	onDestroy(() => {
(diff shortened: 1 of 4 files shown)


src/lib/components/chat/ChatWindow.svelte
```diff
@@ -38,6 +40,7 @@
 
 	let loginModalOpen = false;
 	let message: string;
+	let timeout: ReturnType<typeof setTimeout>;
```
nsarrazin, line 44:
```suggestion
	let isSharedRecently = false;
	$: $page.params.id && (isSharedRecently = false);
```

src/lib/components/chat/ChatWindow.svelte
```diff
@@ -226,12 +246,19 @@
 				</p>
 				{#if messages.length}
 					<button
-						class="flex flex-none items-center hover:text-gray-400 hover:underline max-sm:rounded-lg max-sm:bg-gray-50 max-sm:px-2.5 dark:max-sm:bg-gray-800"
+						class="flex flex-none items-center hover:text-gray-400 max-sm:rounded-lg max-sm:bg-gray-50 max-sm:px-2.5 dark:max-sm:bg-gray-800"
 						type="button"
+						class:hover:underline={!$isConversationShared}
```
nsarrazin, line 251:
```suggestion
						class:hover:underline={!isSharedRecently}
```

src/lib/components/chat/ChatWindow.svelte
```diff
@@ -226,12 +246,19 @@
...
 						on:click={() => dispatch("share")}
+						disabled={$isConversationShared}
```
nsarrazin, line 253:
```suggestion
						disabled={isSharedRecently}
```

src/lib/components/chat/ChatWindow.svelte
```diff
@@ -226,12 +246,19 @@
 				</p>
 				{#if messages.length}
 					<button
-						class="flex flex-none items-center hover:text-gray-400 hover:underline max-sm:rounded-lg max-sm:bg-gray-50 max-sm:px-2.5 dark:max-sm:bg-gray-800"
+						class="flex flex-none items-center hover:text-gray-400 max-sm:rounded-lg max-sm:bg-gray-50 max-sm:px-2.5 dark:max-sm:bg-gray-800"
 						type="button"
+						class:hover:underline={!$isConversationShared}
 						on:click={() => dispatch("share")}
```
nsarrazin, line 252:
```suggestion
						on:click={() => {
							dispatch("share");
							isSharedRecently = true;
							setTimeout(() => {
								isSharedRecently = false;
							}, 2000);
						}}
```

src/lib/components/chat/ChatWindow.svelte
```diff
@@ -226,12 +246,19 @@
...
+						disabled={$isConversationShared}
 					>
-						<CarbonExport class="text-[.6rem] sm:mr-1.5 sm:text-primary-500" />
-						<div class="max-sm:hidden">Share this conversation</div>
+						{#if $isConversationShared}
```
nsarrazin, line 255:
```suggestion
						{#if isSharedRecently}
```

nsarrazin (commented):
I like the new behaviour, looks a lot better :grin: 

I think you can simplify it quite a bit and do it entirely in the `ChatWindow.svelte` without needing a store, see the suggestions, but maybe there's a reason to do it the other way, feel free to accept the changes or not!

src/lib/components/chat/ChatWindow.svelte
nsarrazin: this is the main piece, we dispatch the event and set the timeout here, shouldn't need much else

src/lib/components/chat/ChatWindow.svelte
mishig25: handled in https://github.com/huggingface/chat-ui/pull/645/commits/fdbb4cafbf846c95557cb7e2fe0ce45effcd4c76

src/lib/components/chat/ChatWindow.svelte
mishig25: handled in https://github.com/huggingface/chat-ui/pull/645/commits/fdbb4cafbf846c95557cb7e2fe0ce45effcd4c76

src/lib/components/chat/ChatWindow.svelte
mishig25: handled in https://github.com/huggingface/chat-ui/pull/645/commits/fdbb4cafbf846c95557cb7e2fe0ce45effcd4c76

src/lib/components/chat/ChatWindow.svelte
mishig25: handled in https://github.com/huggingface/chat-ui/pull/645/commits/fdbb4cafbf846c95557cb7e2fe0ce45effcd4c76

src/lib/components/chat/ChatWindow.svelte
mishig25: handled in https://github.com/huggingface/chat-ui/pull/645/commits/fdbb4cafbf846c95557cb7e2fe0ce45effcd4c76

mishig25:
> I think you can simplify it quite a bit and do it entirely in the ChatWindow.svelte without needing a store, see the suggestions, but maybe there's a reason to do it the other way, feel free to accept the changes or not!

initially, I did the store thing to make sure to only show `Copied to clipboard` msg if [share function](https://github.com/huggingface/chat-ui/blob/update_share_btn/src/lib/utils/share.ts#L1-L2) succeeded. But I see that it overcomplicates the flow a bit and if there is an error in  [share function](https://github.com/huggingface/chat-ui/blob/update_share_btn/src/lib/utils/share.ts#L1-L2), it will in console.error anyways

gary149 (approved):

src/lib/components/chat/ChatWindow.svelte
```diff
@@ -38,6 +39,9 @@
 
 	let loginModalOpen = false;
 	let message: string;
+	let timeout: ReturnType<typeof setTimeout>;
```
nsarrazin, line 42: Kind of a nit or maybe I'm missing something: does the timeout needs to be referenced and handled in the `onDestroy` and can't we just do it like this?
```
on:click={() => {
	dispatch("share");
	isSharedRecently = true;
	setTimeout(() => {
		isSharedRecently = false;
	}, 2000);
}}
```

just curious, other than that lgtm

mishig25: > does the timeout needs to be referenced and handled in the onDestroy and can't we just do it like this?

yes, for the edge case.

If a user clicks `share btn` & closes the window within less than 2 seconds, it would end up creating memory leak

PR #652, opened 2023-12-22
Use a low temperature mistral for tasks

nsarrazin:
Decouple the task model settings from the normal endpoint settings.

Use a lower temp for the tasks, as they don't benefit from increased randomness in the tokens


nsarrazin:
I think this can help avoid [this kind of issues](https://huggingface.co/spaces/huggingchat/chat-ui/discussions/336#6585b97f98e0c1f98458733e), @gary149 wdyt?

gary149 (approved):
👍

PR #662, opened 2023-12-26
[Assistants] Use textToImage task for avatar generation

nsarrazin:
This PR adds a togglable feature for generating the assistants avatars using a customizable text to image model.


nsarrazin:
Need to put a front-facing checkbox for the user to use this feature or not when creating an avatar, but as a poc it works quite well. for huggingchat in prod we may need to pin a model for this if possible, to reduce generation latency

src/routes/settings/assistants/new/+page.server.ts
```diff
@@ -37,6 +40,19 @@ const uploadAvatar = async (avatar: File, assistantId: ObjectId): Promise<string
...
+const generateAvatar = (description?: string, name?: string): Promise<string> => {
+	const textPrompt = `Generate a prompt for an image-generation model for the following: 
+Name: ${name}
+Description: ${description}
+`;
+
+	return generateFromDefaultEndpoint({
+		messages: [{ from: "user", content: textPrompt }],
+		preprompt:
+			"You are an assistant tasked with generating simple image descriptions. The user will ask you for an image, baed on the name and a description of what they want, and you should reply with a short, concise, safe, descriptive sentence.",
```
mikelfried, line 52:
```suggestion
			"You are an assistant tasked with generating simple image descriptions. The user will ask you for an image, based on the name and a description of what they want, and you should reply with a short, concise, safe, descriptive sentence.",
```

nsarrazin:
I added a user control in the assistant creation form.
image

There's also a 30s timeout on the image generation.

mishig25:
I'll review it soon. Could you resolve the conflicts in the meantime?

nsarrazin:
Should be good @mishig25 !

mishig25:
From UX-wise, if you click "generate avatar", you shouldn't be able to select local img. Otherwise, it is confusing, which image will be used

https://github.com/huggingface/chat-ui/assets/11827707/92b45bb5-77cf-4e59-a5ca-302eabcf6371

.env
```diff
@@ -128,4 +128,6 @@ LLM_SUMMERIZATION=true
 # PUBLIC_APP_DATA_SHARING=1
 # PUBLIC_APP_DISCLAIMER=1
 
-ENABLE_ASSISTANTS=false #set to true to enable assistants feature
\ No newline at end of file
+ENABLE_ASSISTANTS=false #set to true to enable assistants feature
+ASSISTANTS_GENERATE_AVATAR=true #requires an hf token, uses the model description and name to generate an avatar using a text to image model
+TEXT_TO_IMAGE_MODEL="runwayml/stable-diffusion-v1-5"
```
mishig25, line 133: nit: but we can use newer models like https://huggingface.co/latent-consistency/lcm-lora-ssd-1b (assuming that they are hosted on inference API) that produce higher quality images faster

1. It is higher quality because, it is based on SD XL
2. It produces images faster, because it is smaller model (ssd) & uses fewer sampling (lcm-lora)

src/routes/+layout.server.ts
```diff
@@ -161,6 +163,7 @@ export const load: LayoutServerLoad = async ({ locals, depends }) => {
 			email: locals.user.email,
 		},
 		assistant,
+		avatarGeneration: ASSISTANTS_GENERATE_AVATAR === "true" && TEXT_TO_IMAGE_MODEL !== "",
```
mishig25, line 166:
```suggestion
		avatarGeneration: ASSISTANTS_GENERATE_AVATAR === "true" && !TEXT_TO_IMAGE_MODEL,
```

I think ts string truthiness would work fine here

.env
nsarrazin: Ah that's fair! I picked sd1.5 over sdxl because I assumed it'd be faster but if we have faster models available by all mean let's change it :grin: 

I think we should prio speed>image quality for this feature, most of the time this will be seen as only a small thumbnail, so if you have good model recommendations, feel free!

src/routes/settings/assistants/new/+page.server.ts
```diff
@@ -18,6 +21,10 @@ const newAsssistantSchema = z.object({
 	exampleInput3: z.string().optional(),
 	exampleInput4: z.string().optional(),
 	avatar: z.instanceof(File).optional(),
+	generateAvatar: z
```
mishig25, line 24: do those fields have to be mutually exclusive or not? Basically, both of them can't be active right?

src/lib/utils/generateAvatar.ts
```diff
@@ -0,0 +1,24 @@
+import { HF_TOKEN, TEXT_TO_IMAGE_MODEL } from "$env/static/private";
+import { generateFromDefaultEndpoint } from "$lib/server/generateFromDefaultEndpoint";
+import { HfInference } from "@huggingface/inference";
+
+export async function generateAvatar(description?: string, name?: string): Promise<File> {
+	const queryPrompt = `Generate a prompt for an image-generation model for the following: 
+Name: ${name}
+Description: ${description}
+`;
+	const imagePrompt = await generateFromDefaultEndpoint({
```
mishig25, line 10: is the small model being used here or not?

src/lib/components/AssistantSettings.svelte
```diff
@@ -125,6 +135,12 @@
 					<span class="text-xs text-gray-500 hover:underline">Click to upload</span>
 				{/if}
 				<p class="text-xs text-red-500">{getError("avatar", form)}</p>
+				{#if (!files || !files[0]) && $page.data.avatarGeneration && !assistant?.avatar}
```
mishig25, line 138:
```suggestion
				{#if !files?.[0] && $page.data.avatarGeneration && !assistant?.avatar}
```

I think `!files || !files[0])`, can be just `!files?.[0]`. I haven't tested it

src/lib/components/AssistantSettings.svelte
```diff
@@ -125,6 +135,12 @@
...
+					<label class="text-xs text-gray-500">
+						<input type="checkbox" name="generateAvatar" class="text-xs text-gray-500" />
+						Generate avatar using a text-to-image model.
```
mishig25, line 141:
```suggestion
						Generate avatar
```

maybe just this?

.env
mishig25: for example, https://huggingface.co/latent-consistency/lcm-lora-sdv1-5 would be faster than `runway/sd-1.5`. https://huggingface.co/latent-consistency/lcm-lora-sdxl might still be faster than `runway/sd-1.5`. Maybe @patil-suraj @sayakpaul can confirm?

mishig25: https://huggingface.co/blog/lcm_lora

sayakpaul: Since it's SDXL, it will not be SD speed. You can try out Segmind's SSD-1B.

mishig25: So https://huggingface.co/latent-consistency/lcm-lora-ssd-1b  would be the best, in terms of speed/qualiy tradeoff?

sayakpaul: I would say so. But, would also consider playing SD Turbo.

mishig25: What's the license of SD Turbo ?

src/routes/settings/assistants/new/+page.server.ts
nsarrazin: Yeah pretty much. I do the check a bit later down but do you think there's a nice way to do it with zod directly?

src/lib/utils/generateAvatar.ts
nsarrazin: Yes indeed, like for websearch & title summarization

src/lib/components/AssistantSettings.svelte
nsarrazin: maybe `Generate avatar from description` to make it clear it's text to image and where the definition is coming from ? Though not sure about this, lmk what you think

nsarrazin:
> From UX-wise, if you click "generate avatar", you shouldn't be able to select local img (i.e, click "click to upload", it should be disabled). Otherwise, it is confusing, which image will be used
> Screen.Recording.2024-01-12.at.5.46.41.PM.mov

The intent was that if an image is uploaded, it takes precedence over the checkbox (the checkbox disappears and the image preview is shown instead) but if you think it's clearer to also disable the upload button when the checkbox is checked then I can add that for sure. Do you think it's better if the button is disabled or completely hidden ?

src/lib/components/AssistantSettings.svelte
nsarrazin: :bow:  it works pretty well

src/routes/+layout.server.ts
nsarrazin: I think it should have been `!!`  but it works :grin:

mishig25:
Do you think it's better if the button is disabled or completely hidden ?

> disabled

src/routes/settings/assistants/new/+page.server.ts
mishig25: > I do the check a bit later

oh I see & that's what I meant as well.

> do you think there's a nice way to do it with zod directly

I don't think so

src/lib/components/AssistantSettings.svelte
nsarrazin: done!

nsarrazin:
> > Do you think it's better if the button is disabled or completely hidden ?
> 
> disabled

done!

mishig25 (approved):
lgtm from my side!

.env
nsarrazin: I tried `latent-consistency/lcm-lora-ssd-1bz` but it doesn't seem to load, I think it would need to be pinned in the API on our side if we go for it, just a note not necessarily a blocker 😁

nsarrazin:
There's a feature flag in front of it so I think we can merge this back into the assistants PR and then enable the flag or not in the prod config if we decide to go for it

PR #663, opened 2023-12-26
Fix MODELS env

doracpphp:
I copied the MODELS env and it did not work, so I modified the README.


nsarrazin (approved):
Good catch, thanks for the fix!

PR #668, opened 2023-12-27
[Assistants] Popular assistants page

nsarrazin:
image

Very WIP, idea is to have a page that shows the most popular assistants and allow users to grab them easily.

There should also be a link back to the "new assistant page", to create visibility for the feature inside the app ?

### Random notes
* Only show assistants that have > 1 users?
* Allow filtering per model for now, any other filters ?
* Should we show number of users for a model?
* Design needs work for sure lol

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -8,3 +8,4 @@ node_modules/
 .svelte-kit/
 .env*
 !.env
+!.env.local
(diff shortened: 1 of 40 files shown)


gururise:
**Filter Suggestions:**

- All
- New
- Popular
- Recently Used 

Perhaps the top 8 models would be shown in each category, except (All) which would show all the models.

Might also need a **sort** dropdown as well, so one can sort by Popularity, or Date.

nsarrazin:
I'm continuing work on this over at #740 since the branches diverged pretty heavily. But thanks for the feedback @gururise !

PR #669, opened 2023-12-28
Add support for SerpStack API

gururise:
This PR adds support for [serpstack](https://serpstack.com) following a similar format to the other SERP providers. 

The `SERPSTACK_API_KEY` is added to the .env and when provided with a valid API key, web search should use Serp Stack.


nsarrazin (approved):
We'll probably want to refactor how new webserch modules are added but that's out of scope, LGTM!

Issue #670, opened 2023-12-28
Expected double-quoted property name in JSON


flexchar:
Hi! I've tried this end of October and now giving another shot. Both times I end up facing issues with getting the config right as I want because of various JSON parsing issues...

This is what I'm stuck with at the time of writing this issue and same I had in the past
```bash
2:45:31 PM [vite] .env.local changed, restarting server...

2:45:31 PM [vite] server restarted.
2:45:33 PM [vite] Error when evaluating SSR module /src/lib/server/models.ts:
|- SyntaxError: Expected double-quoted property name in JSON at position 191 (line 10 column 5)
    at JSON.parse (<anonymous>)
    at /Users/luke/pill/chat-ui/src/lib/server/models.ts:71:51
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async instantiateModule (file:///Users/luke/pill/chat-ui/node_modules/vite/dist/node/chunks/dep-e8f070e8.js:54405:9)

2:45:33 PM [vite] Error when evaluating SSR module /src/routes/+layout.server.ts: failed to import "/src/lib/server/models.ts"
|- SyntaxError: Expected double-quoted property name in JSON at position 191 (line 10 column 5)
    at JSON.parse (<anonymous>)
    at /Users/luke/pill/chat-ui/src/lib/server/models.ts:71:51
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async instantiateModule (file:///Users/luke/pill/chat-ui/node_modules/vite/dist/node/chunks/dep-e8f070e8.js:54405:9)

SyntaxError: Expected double-quoted property name in JSON at position 191 (line 10 column 5)
    at JSON.parse (<anonymous>)
    at /Users/luke/pill/chat-ui/src/lib/server/models.ts:71:51
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async instantiateModule (file:///Users/luke/pill/chat-ui/node_modules/vite/dist/node/chunks/dep-e8f070e8.js:54405:9)
SyntaxError: Expected double-quoted property name in JSON at position 191 (line 10 column 5)
    at JSON.parse (<anonymous>)
    at /Users/luke/pill/chat-ui/src/lib/server/models.ts:71:51
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
[...]

PR #671, opened 2023-12-29
Use JSON5 for parsing all config vars

nsarrazin:
This should help with issues like #670 and a lot of others we've seen in the past where users were getting confused about their configs not parsing properly.

Most of the times it's due do trailing commas, comments in the JSON or similar. Using [JSON5](https://json5.org/) solves all of these.

I tested it with the huggingchat prod config and it works fine.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,6 +20,7 @@
 				"highlight.js": "^11.7.0",
 				"image-size": "^1.0.2",
 				"jsdom": "^22.0.0",
+				"json5": "^2.2.3",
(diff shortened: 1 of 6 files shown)


nsarrazin:
In the long run I'd like to do a breaking change and replace the `.env.local` by a `config.yaml` or `config.toml` and refactor large parts of the config file so the vars make more sense, but for now I think this already helps 😁

julien-c (commented):
this PR sounds good to me

Issue #656, opened 2023-12-22
Web Search failed with "Invalid URL"


gururise:
![image](https://github.com/huggingface/chat-ui/assets/4380009/229430b6-6d10-495f-be66-c5bc54f6061d)

Why is this happening?  It seems to happen regardless of whether I have USE_LOCAL_WEBSEARCH set to true or false.
```
SERPAPI_KEY=<my key>
USE_LOCAL_WEBSEARCH=true

MODELS=`[
    {
      "name": "mistralai/Mixtral-8x7b-Instruct-v0.1",
      "displayName": "mistralai/Mixtral-8x7b-Instruct-v0.1",
      "description": "Mixtral-8x7b-Instruct-v0.1 is a state of the art language model, based on a mixture of experts, that outperforms ChatGPT.",
      "websiteUrl": "https://www.aaprintsupplyco.com",
      "preprompt": "",
      "chatPromptTemplate" : "<s>{{#each messages}}{{#ifUser}}[INST] {{#if @first}}{{#if @root.preprompt}}{{@root.preprompt}}\n{{/if}}{{/if}}{{content}} [/INST]{{/ifUser}}{{#ifAssistant}}{{content}}</s>{{/ifAssistant}}{{/each}}",
      "parameters": {
        "temperature": 0.4,
        "top_p": 0.95,
        "top_k": 50,
        "truncate": 31768,
        "max_new_tokens": 2048,
        "stop": ["[INST]","</s>"]
      },
      "endpoints" : [{
        "type": "openai",
        "baseURL": "https://api.together.xyz/v1"
      }],
      "promptExamples": [
        {
          "title": "Write a blog post",
          "prompt": "Your goal is to help me create a compelling blog post about a topic.\nYou will follow the following process:\n\n1. Ask me for the topic of the blog post.\n2. After I provide my answer you will need to collect some additional information by going through the next steps:\na) Questions (ask any relevant questions pertaining to what additional information is needed from me to write a good blog post).\n\nOnce you have enough information, or once I say I am done, you will write the blog post."
        }, {
          "title": "Improve my English",
          "prompt": "I want you to act as an English grammar and spelling corrector and improver. I will speak to you and you will answer in the corrected and improved version of my text, in English. I want you to replace my simplified A0-level words and sentences with improved, higher level English words and sentences. Keep the meaning same, but make them sound better. I want you to only reply the correction, the improvements and nothing else, do not write explanations. If there is nothing to improve, just reply with the original text."
        }, {
[...]

PR #675, opened 2024-01-02
Fixes "invalid url" error in websearch 

muhammadbaqirjafari:
I have fixed the issue #656 

This issue was occuring because local web search sometimes extract invalid url.
So this line causes error:
`const { hostname } = new URL(link);` 

Now I have added exception handling around it.


nsarrazin (approved):
Nice, thanks for the great contribution! Looks good to me, will merge after the tests pass

PR #678, opened 2024-01-04
Reuse Assisntants settings component

mishig25:
This PR:
1. Created `AssistantSettings.svelte` component that is re-used between assistant creation & assistant edit pages. Otherwise, assistant creation & assistant edit pages almost had identical htmls
2. Fixes some issues
    1. padding was missing here Screenshot 2024-01-04 at 6 42 56 PM
    2. Avatar reset btn was weridly placed before. Now, it is placed directly under the avatar Screenshot 2024-01-04 at 6 45 09 PM
    3. Fixed errors in choose avatar btn: 1. When file picker opens & you don't select a file, it was removing the existing avatar. 2. When you click `Reset`, it resets & was opening the file picker afterwards


nsarrazin:
Going to let @gary149 review this one as I think he was working on the design but LGTM, and thanks for making the component reusable!

PR #683, opened 2024-01-09
Add doc section about updating HuggingChat

nsarrazin:
This has been working for a while but wasn't properly documented, now we have it in the docs!


nsarrazin:
Might be relevant for you at some point! cc @gary149 @mishig25

mishig25 (approved):

PR #698, opened 2024-01-16
Add conversation API endpoint to get conversation

alak:
- Sorting conversations for the `/api/conversation` endpoint
- Addind `/api/conversation/[id]` endpoint to get one conversation from its `id`


alak:
I will handle the share part later

PR #704, opened 2024-01-17
[Small model] Explicitly close AsyncGenerator

### LINKED ISSUE #691: Error "fetch failed" sometimes during web search
Sometimes websearch gives you an error "fetch failed". The error is happening when websearch was succesfull, but the RAG augmented message generation failed

And error log in server shows
>  cause: SocketError: other side closed

I wasn't able to reproduce the error locally. However, will spend more time & fix this error

mishig25:
Explicitly close AsyncGenerator when using small model (for generating webquery or sumarization purposes). I couldn't replicate the #691. By explicitly closing the connection of a small model, this might fix #691 if the issue was rooted in some kind of rate limiting around outgoing network requests from Spaces to TGI/inference endpoint


nsarrazin (approved):
Makes sense, let's see if it helps in prod!

PR #707, opened 2024-01-18
Continue generation feature

### LINKED ISSUE #280: Button to continue generation
Sometimes the model stops early when generating. It should be fairly easy to add a "continue generation" button to help unstuck the answer when that happens.

It's been requested quite a bit on the [space](https://huggingface.co/spaces/huggingchat/chat-ui/discussions/179#647ec5c532c471a7faa593c6).

nsarrazin:
close #280 

Added a button to continue generation on the last message.

### How it works

We detect if the last token emitted from the inference API is a special token. If it's **not** a special token, then we assume that the request was cut short. We set a flag `interrupted: true` on the message. If the flag is true and only **if it's the last message** we show the continue button. 

When a message requests passes `is_continue=True`, we build the conv prompt and then strip the stop tokens from the end of the last assistant message, that way the model continues the answer without thinking it's a new message. 

### Demo
https://github.com/huggingface/chat-ui/assets/25119303/937ba9ca-7e75-47fc-9ea4-7aef0228184d

### Notes

* ~~Currently doesn't work with the websearch because the search does some prompt manipulation to inject context, I need to find a workaround for this.~~ Should be fixed
* Support is limited to TGI atm, I can implem some others but it will always be limited to endpoint types that let us manipulate the prompt as a string. (no chat completions API from openAI for example)
* ~~I don't really like having the button in the corner and hidden until hover, I think it should be in the answer bubble centered below the message, or something like that, wdyt @gary149 ?~~ Moved the button 
* We could even have a "limitless generation" mode where we chain calls to the API until the model answers with an end token but I think this could impact us if the model is stuck in a loop for example.

### DIFF
--- a/src/lib/server/endpoints/tgi/endpointTgi.ts
+++ b/src/lib/server/endpoints/tgi/endpointTgi.ts
@@ -15,14 +15,24 @@ export const endpointTgiParametersSchema = z.object({
 
 export function endpointTgi(input: z.input<typeof endpointTgiParametersSchema>): Endpoint {
 	const { url, accessToken, model, authorization } = endpointTgiParametersSchema.parse(input);
-	return async ({ conversation }) => {
-		const prompt = await buildPrompt({
+
+	return async ({ conversation, continue: messageContinue }) => {
+		let prompt = await buildPrompt({
 			messages: conversation.messages,
 			webSearch: conversation.messages[conversation.messages.length - 1].webSearch,
 			preprompt: conversation.preprompt,
 			model,
 			id: conversation._id,
 		});
+
+		if (messageContinue) {
+			prompt = model.parameters.stop.reduce((acc: string, curr: string) => {
+				if (acc.endsWith(curr)) {
+					return acc.slice(0, acc.length - curr.length);
+				}
+				return acc;
+			}, prompt.trimEnd());
+		}
 
 		return textGenerationStream(
 			{
--- a/src/lib/server/endpoints/tgi/endpointTgi.ts
+++ b/src/lib/server/endpoints/tgi/endpointTgi.ts
@@ -26,6 +26,7 @@ export function endpointTgi(input: z.input<typeof endpointTgiParametersSchema>):
 		});
 
 		if (messageContinue) {
+			// start with the full prompt, and for each stop token, try to remove it from the end of the prompt
 			prompt = model.parameters.stop.reduce((acc: string, curr: string) => {
 				if (acc.endsWith(curr)) {
 					return acc.slice(0, acc.length - curr.length);
--- a/src/routes/conversation/[id]/+page.svelte
+++ b/src/routes/conversation/[id]/+page.svelte
@@ -317,27 +317,36 @@
 
 	async function onMessage(event: CustomEvent<string>) {
 		if (!data.shared) {
-			writeMessage({ prompt: event.detail });
+			await writeMessage({ prompt: event.detail });
 		} else {
-			convFromShared()
+			await convFromShared()
 				.then(async (convId) => {
 					await goto(`${base}/conversation/${convId}`, { invalidateAll: true });
 				})
-				.then(() => writeMessage({ prompt: event.detail }))
+				.then(async () => await writeMessage({ prompt: event.detail }))
 				.finally(() => (loading = false));
 		}
 	}
 
 	async function onRetry(event: CustomEvent<{ id: Message["id"]; content: string }>) {
 		if (!data.shared) {
-			writeMessage({ prompt: event.detail.content, messageId: event.detail.id, isRetry: true });
+			await writeMessage({
+				prompt: event.detail.content,
+				messageId: event.detail.id,
+				isRetry: true,
+			});
 		} else {
-			convFromShared()
+			await convFromShared()
 				.then(async (convId) => {
 					await goto(`${base}/conversation/${convId}`, { invalidateAll: true });
 				})
-				.then(() =>
-					writeMessage({ prompt: event.detail.content, messageId: event.detail.id, isRetry: true })
+				.then(
+					async () =>
+						await writeMessage({
+							prompt: event.detail.content,
+							messageId: event.detail.id,
+							isRetry: true,
+						})
 				)
 				.finally(() => (loading = false));
 		}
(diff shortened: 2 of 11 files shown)


nsarrazin:
Might be worth having a look @mishig25 I tweaked the web search section of `buildPrompt` a bit. Basically before we were removing the last message and replacing it with the modified websearch injection prompt, because we assumed that the last message was always a user message. 

Because of the continue feature, the last message can be an assistant message that we expand. So now we find the last user message by index, instead of assuming it's the last message. Should be identical behaviour, but just FYI

nsarrazin:
Here's how it looks now

image
image

src/routes/conversation/[id]/+server.ts
```diff
@@ -136,38 +138,53 @@ export async function POST({ request, locals, params, getClientAddress }) {
 		hashes = await Promise.all(files.map(async (file) => await uploadFile(file, conv)));
 	}
 
+	// can only call isContinue on the last message id
+	if (
+		isContinue &&
+		conv.messages.findIndex((message) => message.id === messageId) !== conv.messages.length - 1
```
Comment (mishig25) [inline src/routes/conversation/[id]/+server.ts:144]:
```suggestion
		conv.messages[conv.messages.length - 1].id !=== messageId
```

haven't tested it but wouldn't this simpler boolean statement work ?

src/routes/conversation/[id]/+server.ts
```diff
@@ -136,38 +138,53 @@ export async function POST({ request, locals, params, getClientAddress }) {
...
+					content: conv.messages[retryMessageIdx]?.content,
 					from: "user",
 					id: messageId as Message["id"],
 					updatedAt: new Date(),
 					files: conv.messages[retryMessageIdx]?.files,
 				},
 			];
+		} else if (isContinue && messageId) {
+			// for continue we do nothing and expand the last assistant message
+			return [...conv.messages];
```
Comment (mishig25) [inline src/routes/conversation/[id]/+server.ts:173]:
```suggestion
			return conv.messages;
```

why not just this ? do we need to clone it [... xyz] ?

src/routes/conversation/[id]/+page.svelte
```diff
@@ -275,37 +310,45 @@
...
+				.then(() =>
+					writeMessage({ prompt: event.detail.content, messageId: event.detail.id, isRetry: true })
+				)
 				.finally(() => (loading = false));
 		}
 	}
 
+	async function onContinue(event: CustomEvent<{ id: Message["id"] }>) {
+		if (!data.shared) {
+			writeMessage({ messageId: event.detail.id, isContinue: true });
```
Comment (mishig25) [inline src/routes/conversation/[id]/+page.svelte:348]:
```suggestion
			await writeMessage({ messageId: event.detail.id, isContinue: true });
```

src/routes/conversation/[id]/+page.svelte
```diff
@@ -64,23 +64,39 @@
...
 	}
 	// this function is used to send new message to the backends
-	async function writeMessage(message: string, messageId = randomUUID()) {
-		if (!message.trim()) return;
-
+	async function writeMessage({
+		prompt,
+		messageId = randomUUID(),
+		isRetry = false,
+		isContinue = false,
```
Comment (mishig25) [inline src/routes/conversation/[id]/+page.svelte:71]: this is a stylistic comment. In hf/moon-landing, we usually group those kind of arguments into one argument called `opts`. So it would be:
```ts
opts:{
    isRetry?: boolean,
    isContinue?: boolean
}
```

src/routes/conversation/[id]/+page.svelte
```diff
@@ -99,25 +115,44 @@
...
+					...messages.slice(0, msgIndex),
+					{
+						from: "user",
+						content: messages[msgIndex].content,
+						id: messageId,
+						files: messages[msgIndex].files,
+					},
+				];
+			} else if (isContinue) {
+				console.log("isContinue");
```
Comment (mishig25) [inline src/routes/conversation/[id]/+page.svelte:130]: unneded ?

src/routes/conversation/[id]/+server.ts
```diff
@@ -292,8 +318,9 @@ export async function POST({ request, locals, params, getClientAddress }) {
 							...messages.slice(0, -1),
 							{
 								...messages[messages.length - 1],
-								content: output.generated_text,
+								content: previousContent + output.generated_text,
```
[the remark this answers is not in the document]
Comment (nsarrazin) [reply src/routes/conversation/[id]/+server.ts:321]: Yep shouldn't add an extra space, in case a word is composed of multiple tokens.

src/routes/conversation/[id]/+page.svelte
Comment (nsarrazin) [reply src/routes/conversation/[id]/+page.svelte:130]: oh yeah good catch idk why i did that

src/lib/server/endpoints/tgi/endpointTgi.ts
```diff
@@ -15,15 +15,25 @@ export const endpointTgiParametersSchema = z.object({
...
+		let prompt = await buildPrompt({
 			messages: conversation.messages,
 			webSearch: conversation.messages[conversation.messages.length - 1].webSearch,
 			preprompt: conversation.preprompt,
 			model,
 			id: conversation._id,
 		});
 
+		if (messageContinue) {
+			prompt = model.parameters.stop.reduce((acc: string, curr: string) => {
```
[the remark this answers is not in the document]
nsarrazin: On it!

src/routes/conversation/[id]/+server.ts
Comment (nsarrazin) [reply src/routes/conversation/[id]/+server.ts]: fixed it!

src/lib/server/endpoints/tgi/endpointTgi.ts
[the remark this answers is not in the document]
nsarrazin: left
```// start with the full prompt, and for each stop token, try to remove it from the end of the prompt```

src/routes/conversation/[id]/+page.svelte
Comment (nsarrazin) [reply src/routes/conversation/[id]/+page.svelte:71]: Would you also change it at the API level or just for this function ?

gary149 (approved):

src/routes/conversation/[id]/+page.svelte
Comment (nsarrazin) [reply src/routes/conversation/[id]/+page.svelte:71]: Let's tackle it in a refacto PR, i think i might need some extra opts for conversation trees anyway :grin:

PR #711, opened 2024-01-19
[JS] use shorthand notation

mishig25:
Using `shorthand notation`. Just a stylistic difference https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -34,6 +34,7 @@ module.exports = {
 				argsIgnorePattern: "^_",
 			},
 		],
+		"object-shorthand": ["error", "always"],
 	},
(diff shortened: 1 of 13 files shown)


nsarrazin:
Agree with this, should we specify it in our [eslint config directly](https://eslint.org/docs/latest/rules/object-shorthand) then to make it easier to catch in the future ? :grin:

mishig25:
added the rule https://github.com/huggingface/chat-ui/pull/711/commits/34da3273253f3d849a679f613440d3036c95e343

nsarrazin (approved):
Nice, thanks!

PR #718, opened 2024-01-22
Bring back mistral task model

nsarrazin:
Lower temp config specifically for tasks


mishig25:
@nsarrazin does it solve the summarization hallucination issues ?

PR #719, opened 2024-01-22
Use JSON5 for embeddings

### LINKED ISSUE #716: JSON.parse
Have already looked into solving by using JSON5, it's still not working!

SyntaxError: Unexpected token # in JSON at position 418
    at JSON.parse (<anonymous>)
    at C:\Users\Parth\Desktop\chat-ui-main\src\lib\server\embeddingModels.ts:41:60
    at async instantiateModule (file:///C:/Users/Parth/Desktop/chat-ui-main/node_modules/vite/dist/node/chunks/dep-9A4-l-43.js:50861:9)

nsarrazin:
We were parsing with built-in JSON but we use JSON5 everywhere else so just a quickfix here

Should fix #716


mishig25 (approved):
lgtm !

PR #720, opened 2024-01-23
Mobile navs: bigger buttons and hide new?

gary149:
image

### DIFF
--- a/src/lib/components/MobileNav.svelte
+++ b/src/lib/components/MobileNav.svelte
@@ -3,6 +3,7 @@
 	import { createEventDispatcher } from "svelte";
 	import { browser } from "$app/environment";
 	import { base } from "$app/paths";
+	import { page } from "$app/stores";
 
 	import CarbonClose from "~icons/carbon/close";
 	import CarbonAdd from "~icons/carbon/add";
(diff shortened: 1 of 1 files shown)


mishig25:
checking

src/lib/components/MobileNav.svelte
```diff
@@ -30,18 +31,20 @@
...
 		on:click={() => dispatch("toggle", true)}
 		aria-label="Open menu"
 		bind:this={openEl}><CarbonTextAlignJustify /></button
 	>
 	<span class="truncate px-4">{title}</span>
-	<a href={`${base}/`} class="-mr-3 flex h-9 w-9 shrink-0 items-center justify-center"
-		><CarbonAdd /></a
+	<a
+		class:invisible={!$page.params.id}
+		href="{base}/"
```
mishig25, line 46:
```suggestion
		href={base}
```

src/lib/components/MobileNav.svelte
```diff
@@ -30,18 +31,20 @@
 </script>
 
 <nav
```
mishig25, line 33:
```suggestion

<svelte:window on:resize={() => dispatch("toggle", false)} />

<nav
```

you can add this line. Otherwise, when mobilenav is open and user makes the window, the mobile nav is still open

gary149: >  when mobilenav is open and user makes the window

? it's on mobile the window doesn't resize? or I'm missing something?

PR #721, opened 2024-01-23
[Assistants] Generate avatar button + endpoint

nsarrazin:
https://github.com/huggingface/chat-ui/assets/25119303/e551ce9d-425a-41b6-982d-30b72aaebba1

Replace the checkbox by an avatar button, also cleaned up the upload button design. The button calls an endpoint that is rate-limited to avoid abuse.


src/lib/components/AssistantSettings.svelte
```diff
@@ -53,8 +56,48 @@
...
 
 	let loading = false;
+	let generatingAvatar = false;
 
-	let generateAvatar = false;
+	async function onGenerate() {
+		generatingAvatar = true;
+		form = null;
+
+		const generatedAvatar = await fetch(base + "/generate", {
```
gary149, line 65: `/generate` can be a confusing route (for a text-genreation app)? maybe `generate-avatar` instead?

nsarrazin: fixed!

PR #722, opened 2024-01-23
close the stream after the final answer

alak:
If the stream is not closed it will timeout and create an error client side.

### DIFF
--- a/src/routes/conversation/[id]/+server.ts
+++ b/src/routes/conversation/[id]/+server.ts
@@ -349,6 +349,7 @@ export async function POST({ request, locals, params, getClientAddress }) {
 			});
 
 			await summarizeIfNeeded;
+			controller.close();
 			return;
 		},
 		async cancel() {


alak:
Make sense. Done!

nsarrazin:
Going to test it a bit but should be good :grin:

.env.template
```diff
@@ -254,4 +254,4 @@ PUBLIC_GOOGLE_ANALYTICS_ID=G-8Q63TH4CSL
 # ADDRESS_HEADER=X-Forwarded-For
 # XFF_DEPTH=2
 
-EXPOSE_API=false
\ No newline at end of file
+EXPOSE_API=true
```
nsarrazin, line 257: Want to switch it on for prod ? :eyes:

alak: oupsy

alak: fixed

nsarrazin: no problem, thanks!

nsarrazin (approved):
Also looks good to me!

mishig25 (approved):
lgtm !

mishig25:
@nsarrazin do we have to do something similar for the small model as well? OR is the logic/usage different there?

nsarrazin:
small model doesnt stream directly to the user, so should be all good there :grin:

PR #725, opened 2024-01-24
[Assistant] Delete avatar button instead of reset

nsarrazin:
This PR changes the "reset" behaviour to a "delete" behaviour


nsarrazin:
This PR merges against #639 and will remove the existing AI generation feature there @gary149

PR #727, opened 2024-01-24
[Fix] small fixes in websearch generateQuery

mishig25:
small fixes in websearch generateQuery found while digging deeper

### DIFF
--- a/src/lib/server/websearch/generateQuery.ts
+++ b/src/lib/server/websearch/generateQuery.ts
@@ -40,7 +40,7 @@ Current Question: What about Mexico?
 			content: `Previous questions: 
 - When is the next formula 1 grand prix?
 
-Current Question: Where is it being hosted ?`,
+Current Question: Where is it being hosted?`,
 		},
 		{
 			from: "assistant",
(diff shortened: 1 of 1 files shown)


src/lib/server/websearch/generateQuery.ts
```diff
@@ -40,7 +40,7 @@ Current Question: What about Mexico?
 			content: `Previous questions: 
 - When is the next formula 1 grand prix?
 
-Current Question: Where is it being hosted ?`,
+Current Question: Where is it being hosted?`,
```
mishig25, line 43: because it is not 🇫🇷 : we don't need space before question mark

src/lib/server/websearch/generateQuery.ts
```diff
@@ -69,7 +69,7 @@ Current Question: Where is it being hosted ?`,
 							.map(({ content }) => `- ${content}`)
 							.join("\n")}`
 					: "") +
-				"\n\nCurrent Question:" +
+				"\n\nCurrent Question: " +
```
mishig25, line 72: we need space here. Otherwise it was becoming something like `Current Question:who is Elon Musk`

src/lib/server/websearch/generateQuery.ts
nsarrazin: damn, getting called out :laughing: good catch though!

PR #729, opened 2024-01-24
[SmallModel] TASK_MODEL='mistralai/Mistral-7B-Instruct-v0.1'

mishig25:
According to my experiments https://github.com/huggingface/chat-ui-experiments/tree/main/1-task-model-hallucination

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -220,18 +220,7 @@ OLD_MODELS=`[
   {"name": "tiiuae/falcon-180B-chat"}
 ]`
 
-TASK_MODEL=`{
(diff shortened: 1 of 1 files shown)


nsarrazin:
Is the model still pinned ? Otherwise this could break the tasks if a cold start occurs, maybe cc @Narsil

mishig25:
> Is the model still pinned ? Otherwise this could break the tasks if a cold start occurs, maybe cc @Narsil

from my side, no idea if it is pinned or not

mishig25:
it is pinned

PR #739, opened 2024-01-26
switching api to public

alak:
Hey,

In order to release the first internal beta of the iOS app I will need the api to go public.


nsarrazin (approved):
Checked the API routes one last time, I think we should be all good here! Let me know when you want this deployed @alak

alak:
@nsarrazin if you can deploy it today it would be great. Thanks

Issue #730, opened 2024-01-24
Use jinja for chat templates


nsarrazin:
We can use [@huggingface/jinja](https://www.npmjs.com/package/@huggingface/jinja) for parsing chat templates and fetch chat templates from the hub directly. This would allow for a single source of truth for model templates.

julien-c:
(cc @xenova for viz)

xenova:
And luckily, we already have `@xenova/transformers` as a dependency, so we can upgrade to >=2.14, and then use it as follows:

```js
import { AutoTokenizer } from "@xenova/transformers";

// Load tokenizer from the Hugging Face Hub
const tokenizer = await AutoTokenizer.from_pretrained("mistralai/Mistral-7B-Instruct-v0.1");

// Define chat messages
const chat = [
	{ role: "user", content: "Hello, how are you?" },
	{ role: "assistant", content: "I'm doing great. How can I help you today?" },
	{ role: "user", content: "I'd like to show off how chat templating works!" },
];

const text = tokenizer.apply_chat_template(chat, { tokenize: false });
// "<s>[INST] Hello, how are you? [/INST]I'm doing great. How can I help you today?</s> [INST] I'd like to show off how chat templating works! [/INST]"

const input_ids = tokenizer.apply_chat_template(chat, { tokenize: true, return_tensor: false });
// [1, 733, 16289, 28793, 22557, 28725, 910, 460, 368, 28804, 733, 28748, 16289, 28793, 28737, 28742, 28719, 2548, 1598, 28723, 1602, 541, 315, 1316, 368, 3154, 28804, 2, 28705, 733, 16289, 28793, 315, 28742, 28715, 737, 298, 1347, 805, 910, 10706, 5752, 1077, 3791, 28808, 733, 28748, 16289, 28793]
```

mishig25:
Thanks to https://github.com/huggingface/chat-ui/pull/850, there is a new field in model info called `tokenizer` which can be used for calling 

```ts
const tokenizer = await AutoTokenizer.from_pretrained(model.tokenizer);
```

which is already the case here

https://github.com/huggingface/chat-ui/blob/df27d779deb319d0f7cb50ab3c799666e12149fb/src/lib/components/TokensCounter.svelte#L12-L15

PR #744, opened 2024-01-26
Use jinja template for chat formatting (#730)

nsarrazin:
* Upgrade transformers version and fix the types in transformersjs.ts

* Got rid of legacy parameters `userMessageToken`, `userMessageEndToken`, `assistantMessageToken`, `assistantMessageEndToken` so this is a breaking change but we've been pushing for using `chatPromptTemplate` for a while so I feel like it makes sense.

We support both `chatPromptTemplate` and just specifying the `tokenizer`. Priority goes `chatPromptTemplate > tokenizer` (more specific)

Did not change the templates in prod yet because they don't work well with system prompts

(#730)


.env.template
```diff
@@ -33,11 +32,7 @@ MODELS=`[
       "name": "meta-llama/Llama-2-70b-chat-hf",
       "description": "The latest and biggest model from Meta, fine-tuned for chat.",
       "websiteUrl": "https://ai.meta.com/llama/",
-      "userMessageToken": "",
-      "userMessageEndToken": " [/INST] ",
-      "assistantMessageToken": "",
-      "assistantMessageEndToken": " </s><s>[INST] ",
-      "preprompt": " ",
+      "preprompt": "",
       "chatPromptTemplate" : "<s>[INST] <<SYS>>\n{{preprompt}}\n<</SYS>>\n\n{{#each messages}}{{#ifUser}}{{content}} [/INST] {{/ifUser}}{{#ifAssistant}}{{content}} </s><s>[INST] {{/ifAssistant}}{{/each}}",
```
nsarrazin, line 36: can't remove it for llama-2-70b-chat-hf because the model is gated and so is the chat template, too bad haha

nsarrazin:
@mishig25 this should be good to go! am not using it in the prod config because most of the templates fail, but tested locally with some other models and it should work well.

src/lib/server/embeddingEndpoints/transformersjs/embeddingEndpoints.ts
```diff
@@ -1,6 +1,6 @@
 import { z } from "zod";
 import type { EmbeddingEndpoint } from "../embeddingEndpoints";
-import type { Tensor, Pipeline } from "@xenova/transformers";
+import type { Tensor, FeatureExtractionPipeline } from "@xenova/transformers";
```
mishig25, line 3: double checking: this change is unrelated to tokenization (this PR) right? I think we can still keep it

nsarrazin: yes I was having type related issues with the update to transformers js so unrelated but needed

src/lib/server/models.ts
```diff
@@ -84,11 +73,65 @@ const modelConfig = z.object({
 
 const modelsRaw = z.array(modelConfig).parse(JSON5.parse(MODELS));
 
+async function getChatPromptRender(
```
mishig25, line 76: this is singleton right? This only gets called once at init. Is this correct assumption?

nsarrazin: yes correct! it's called where the template compilation used to take place

src/lib/server/models.ts
```diff
@@ -84,11 +73,65 @@ const modelConfig = z.object({
...
+	m: z.infer<typeof modelConfig>
+): Promise<ReturnType<typeof compileTemplate<ChatTemplateInput>>> {
+	if (m.chatPromptTemplate) {
```
mishig25, line 79: shouldnt the order of fallback be the other way around? First, try to use transfomersjs, then try to use chatPromptTemplate

src/lib/server/models.ts
```diff
@@ -84,11 +73,65 @@ const modelConfig = z.object({
...
+		return compileTemplate<ChatTemplateInput>(m.chatPromptTemplate, m);
+	} else {
```
mishig25, line 81: because of `return` the line above, you can delete `else` and uniindent the lines inside else

src/lib/server/models.ts
nsarrazin: The downside of doing it this way is that for example in prod, we specify the tokenizer for token counting but we actually want to override the chat template with our own template. 

I think in terms of specificity it makes sense to be `custom chat template > tokenizer "default" template` wdyt?

src/lib/server/models.ts
nsarrazin: done!

src/lib/server/models.ts
mishig25: yep sounds good. And if we want to change the order, let's do it in follow-up PR. Also, if we want to change the order, we would need to handle https://github.com/huggingface/chat-ui/pull/744#discussion_r1551321918 first

mishig25 (approved):
lgtm !

src/lib/server/models.ts
nsarrazin: yes agreed, overall we should push towards using tokenizers rather than chat prompt templates, but we will need to handle the edge cases before indeed 😁

src/lib/server/models.ts
```diff
@@ -84,11 +73,65 @@ const modelConfig = z.object({
...
+				formattedMessages = [
+					{
+						role: "system",
+						content: preprompt,
+					},
+					...formattedMessages,
+				];
+			}
+
+			const output = tokenizer.apply_chat_template(formattedMessages, {
```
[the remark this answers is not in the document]
nsarrazin: yep, I tried looking into it, but because the error messages are [hardcoded in the jinja template](https://huggingface.co/mistralai/Mistral-7B-Instruct-v0.2/blob/main/tokenizer_config.json#L42), it's not obvious to me how to determine if the template is failing bc of a lack of system prompt or for some other reason

I guess you could retry without the `"system"` message and see if it builds the prompt then? we could do it in a second PR if that sounds good to you ?

flexchar:
So I pulled the latest changes and was met with the exciting greeting `No tokenizer specified and no chat prompt template specified for model gpt-3.5-turbo-0125`. Not so nice :)

I debugged my way to this PR introducing the error https://github.com/huggingface/chat-ui/blame/a9c711026b13672328ce93abc3865f1823524b0d/src/lib/server/models.ts#L133.

What is the migration path for this breaking change? 

It would be nice to have such ones documented in a dedicated file. It would save a lot of time us all down the road. ✌️

zacps:
I just ran into this as well, I'm not sure I understand why the default prompt template was removed?

It seems like keeping it would alleviate the need for a breaking change.

nsarrazin:
Hey, apologies for the breaking change, fixed it in #985 

My reasoning was that we want people to use either a `tokenizer` or `chatPromptTemplate` to ensure the chat is formatted correctly but of course if you use an endpoint that uses chat completion API, you don't need a chat template 🤦 So you would get an error complaining that something was missing even though you didn't need it.

I've set a default that should be sane (ChatML) which should fix the issue, and in the future I want to add a `raw: true/false` flag at the model level, so we can check if we need a prompt template at all and use the right API routes. 

Seems like most people are moving away from text completion APIs and moving towards chat completion anyway 👀

iChristGit:
Hey @nsarrazin 
Ive been following for a while!
I had some issues that some models don't respond correctly to making up the "Title" and web search.
Does using the tokenizer instead fix the issue with the chatPromptTemplate being hard to set up for local models that are not on the prompt list?
https://github.com/huggingface/chat-ui/blob/main/PROMPTS.md

nsarrazin:
Yes it should help! As long as your model has a `chat_template` in the `tokenizer_config.json` on the hub ([example](https://huggingface.co/google/gemma-1.1-7b-it/blob/main/tokenizer_config.json#L1747) for `google/gemma-1.1-7b-it`) , you can just use the model name in the `tokenizer` field. (for example `tokenizer: "google/gemma-1.1-7b-it"` in this case)

PR #745, opened 2024-01-28
Feature/multi rag

zzen0008:
## Extend RAG Contexts to Support PDF and WebSearch Simultaneously

Enhanced the RAG context to support both PDF and web search in parallel. Here are the main changes made:

1. **Refactor `buildPrompt` function:** Refactored the `buildPrompt` function and its implementations to support multiple RAG contexts. Specifically adjusted for web search and PDF chat. Accommodates more than one RAG context.

2. **Update conversation endpoints:** Updated the conversation endpoints in the `server` to match the refactored `buildPrompt` function, replacing `ragContext` with `ragContexts`. This ensures that the conversation can handle these two types of contexts.

3. **Update `rag.ts`:** Updated `rag.ts` to export webSearch and pdfChat separately from the `RAGs` namespace. This makes their usage in the code more explicit.

4. **Alter `Message` type:** Altered the `Message` type to include `ragContexts` which will hold the `webSearch` and `pdfChat` contexts.

5. **Adapt POST method:** Adapted the POST method in `routes/conversation/[id]/+server.ts` to make use of the new `ragContexts`. Consequently, whichever context is retrieved, it is set in the relevant field of the last message in the conversation.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -2,7 +2,7 @@
 	"editor.formatOnSave": true,
 	"editor.defaultFormatter": "esbenp.prettier-vscode",
 	"editor.codeActionsOnSave": {
-		"source.fixAll": true
+		"source.fixAll": "explicit"
 	},
 	"eslint.validate": ["javascript", "svelte"]
 }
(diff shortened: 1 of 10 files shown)


itaybar:
I realy appreciate this PR, I think this is a great one. I Wrote one comment that I think will be realy major to make this feature WAY more generic and scaleable

PR #749, opened 2024-01-29
Web search: UI update

gary149:
Try to remove the need to see any details.

https://github.com/huggingface/chat-ui/assets/3841370/5b4a3689-1c25-415b-b84b-264be9c8446b


src/lib/components/OpenWebSearchResults.svelte
```diff
@@ -11,31 +8,69 @@
...
-		<span class="px-2 font-medium" class:text-red-700={error} class:dark:text-red-500={error}>
-			Web search
-		</span>
-		<div class="my-auto transition-all" class:rotate-90={detailsOpen}>
-			<CarbonCaretRight />
+	<summary class="grid min-w-72 select-none grid-cols-[40px,1fr] items-center gap-2.5 p-2">
+		<div
+			class="relative grid aspect-square place-content-center overflow-hidden rounded-lg bg-gray-100 dark:bg-gray-800"
+		>
+			<svg
```
mishig25, line 22: they can probably be placed under `src/lib/components/icons`

mishig25:
Just tried it locally. It is 🔥

src/lib/components/OpenWebSearchResults.svelte
```diff
@@ -11,31 +8,69 @@
...
+				class="absolute inset-0 text-gray-300 transition-opacity dark:text-gray-700 {!sources &&
+				!error
+					? 'opacity-100'
+					: 'opacity-0'}"
+				width="40"
+				height="40"
```
mishig25, line 28: once they are placed as icons, we should probably use the units `rem`

src/lib/components/OpenWebSearchResults.svelte
```diff
@@ -88,6 +123,16 @@
 </details>
 
 <style>
+	details[open] .content {
+		animation-name: grow;
+		animation-duration: 300ms;
+		animation-delay: 0ms;
```
mishig25, line 129: the animation only works for the first time, And there is no animation when you close it or open it for the 2+ time

src/lib/components/OpenWebSearchResults.svelte
```diff
@@ -11,31 +8,69 @@
...
-			<CarbonCaretRight />
+	<summary class="grid min-w-72 select-none grid-cols-[40px,1fr] items-center gap-2.5 p-2">
+		<div
+			class="relative grid aspect-square place-content-center overflow-hidden rounded-lg bg-gray-100 dark:bg-gray-800"
+		>
+			<svg
+				class="absolute inset-0 text-gray-300 transition-opacity dark:text-gray-700 {!sources &&
+				!error
+					? 'opacity-100'
+					: 'opacity-0'}"
```
mishig25, line 26:
```suggestion
				class="absolute inset-0 text-gray-300 transition-opacity dark:text-gray-700 opacity-0"
				class:opacity-100={!sources && !error}
```

haven't tested but I think it should work

mishig25:
[export let loading = false;](https://github.com/huggingface/chat-ui/blob/web-search-ui-update/src/lib/components/OpenWebSearchResults.svelte#L7) is no longer being used? If so, we can remove it

src/lib/components/OpenWebSearchResults.svelte
gary149: Here it's not really an icon and some processing happens on the <path>, so it makes sense to keep it here. (But I did move the globe icon)

src/lib/components/OpenWebSearchResults.svelte
gary149: See my above comment. it makes sense to user pixels in this context.

src/lib/components/OpenWebSearchResults.svelte
gary149: the element would have `opacity-0` and `opacity-100` set at the same time in classes, not ideal.

src/lib/components/OpenWebSearchResults.svelte
```diff
@@ -1,15 +1,16 @@
 <script lang="ts">
 	import type { WebSearchUpdate } from "$lib/types/MessageUpdate";
-	import CarbonError from "~icons/carbon/error-filled";
 
+	import CarbonError from "~icons/carbon/error-filled";
 	import EosIconsLoading from "~icons/eos-icons/loading";
+	import IconInternet from "./icons/IconInternet.svelte";
```
mishig25, line 6: on the above lines, you can see that we use the icons directly from libs when possible
```html
import CarbonError from "~icons/carbon/error-filled";
import EosIconsLoading from "~icons/eos-icons/loading";
```

Similarly, would it be possible not to manually define `IconInternet`. Instead use it like from "~icons/xyz/internet" ?

src/lib/components/OpenWebSearchResults.svelte
gary149: I removed it completly.

src/lib/components/OpenWebSearchResults.svelte
gary149: It's a custom made icon (not from iconify)

mishig25 (commented):
lgtm !

PR #752, opened 2024-01-30
Add CodeLlama-70b-Instruct-hf

nsarrazin:
This PR adds CodeLlama-70b-Instruct-hf to HuggingChat. I had to make sure we were properly stripping the end tokens, but here we go, it should work!


mishig25 (approved):
lgtm!

mishig25 (approved):

PR #754, opened 2024-01-30
Remove tabs in websearch prompt

nsarrazin:
The websearch prompt injected extra tabs because of the multi-line string :sweat_smile:


mishig25 (approved):
lgtm !

PR #760, opened 2024-02-01
Several QoL contributions

flexchar:
Hi! 

First of all, many times thank you for wonderful product.

This is enormous amount of work that has beyond value said in words. I've personally found it to help me in my daily tasks using private API keys as well as testing variety of model providers through `openai`-compatible endpoints.

I have end up customising several parts of the software to make it a tiny bit more flexible & friendly for my use-cases. I felt this would be valuable to share outside the local computer folder. 

List of changes - I've made a dedicated commit for each for better visibility:
- customize disclaimer message shown on login popup;
- support `defaultHeaders` when using `openai` endpoints;
- add several examples to README for `openai` endpoints including PortKey's recently open sourced gateway for Claude and Azure OpenAI, also Mistral through Deepinfra; (I'm not affiliated with any but I use them extensively);
- fix streaming issue when I deploy behind cloudflare network (messages would appear as if `stream: false` has been used);
- support rendering new lines in model descriptions;
- hide automatic urls to huggingface unless `modelUrl` explicitly set - I shared with non technical friends and they found themselves in broken links faster than one can blink :);

Let me know if you'd like me to change something! Personally I would be very glad to have these merged as it'd allow me to `git pull` from the `main` and stay upstream with no merge conflicts. Cheers.


julien-c (commented):
wow, neat PR @flexchar!

nsarrazin:
🔥 Thanks for the PR, will do a deeper review but overall I think these are all good improvements we could definitely use!

flexchar:
I pushed couple changes to have the workflows pass :)

gary149 (approved):
LGTM, wdyt @nsarrazin ?

nsarrazin (approved):
Also looks good to me! thanks for contributing @flexchar !

I like the added `defaultHeaders` and I think we should also make it available to other endpoint types but that can be in another PR. For now let's get this merged :grin:

PR #763, opened 2024-02-02
add an api endpoint to get current user's informations

alak:


nsarrazin (approved):
Looks good to me!

We could maybe call the route `/api/whoami` to be consistent with [the hub API](https://huggingface.co/api/whoami-v2) ? not sure  :grin:

PR #764, opened 2024-02-02
Assistants: UI tweaks for release

gary149:


nsarrazin (approved):
Overall looks good to me, maybe we should put a min width on the AssistantIntroduction component ? Since otherwise with a short name/description it looks a bit strange

Screenshot from 2024-02-02 12-57-31

Other than that LGTM

PR #775, opened 2024-02-05
Make the assistant & system prompt button floating on top of screen

nsarrazin:
image

It was confusing on longer conversations, as you couldn't easily go back to the assistant without scrolling all the way up.


gary149:
Not sure about it but going to test it.
Another solution is showing the assistant name here instead of model:
image

PR #778, opened 2024-02-05
Disable websearch on assistants, server & client side

nsarrazin:
This makes sure the websearch feature is always disabled on assistants

### DIFF
--- a/src/routes/conversation/[id]/+page.svelte
+++ b/src/routes/conversation/[id]/+page.svelte
@@ -154,7 +154,7 @@
 					id: messageId,
 					is_retry: isRetry,
 					is_continue: isContinue,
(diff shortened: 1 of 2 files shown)


src/routes/conversation/[id]/+page.svelte
```diff
@@ -151,7 +154,7 @@
 					id: messageId,
 					is_retry: isRetry,
 					is_continue: isContinue,
-					web_search: $webSearchParameters.useSearch,
+					web_search: hasAssistant ? false : $webSearchParameters.useSearch,
```
Comment (gary149) [inline src/routes/conversation/[id]/+page.svelte:157]:
```suggestion
					web_search: !hasAssistant && $webSearchParameters.useSearch
```

PR #779, opened 2024-02-05
Add Plausible analytics

nsarrazin:
This PR adds plausible analytics, by setting a new env var `PUBLIC_PLAUSIBLE_DOMAIN`

In prod we proxy the plausible script to `/js/script.js` so the script doesn't need to be in huggingchat directly.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -235,6 +235,7 @@ RATE_LIMIT=16
 MESSAGES_BEFORE_LOGIN=5# how many messages a user can send in a conversation before having to login. set to 0 to force login right away
 
 PUBLIC_GOOGLE_ANALYTICS_ID=G-8Q63TH4CSL
+PUBLIC_PLAUSIBLE_DOMAIN=huggingface.co
 
 # Not part of the .env but set as other variables in the space
 # ADDRESS_HEADER=X-Forwarded-For
--- a/.env.template
+++ b/.env.template
@@ -235,7 +235,7 @@ RATE_LIMIT=16
 MESSAGES_BEFORE_LOGIN=5# how many messages a user can send in a conversation before having to login. set to 0 to force login right away
 
 PUBLIC_GOOGLE_ANALYTICS_ID=G-8Q63TH4CSL
-PUBLIC_PLAUSIBLE_SCRIPT_URL="/js/script.js"
+PUBLIC_PLAUSIBLE_SCRIPT_URL="https://huggingface.co/js/script.js"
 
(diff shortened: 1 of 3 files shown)


julien-c (approved):
lgtm

cc @coyotte508 @SBrandeis @Kakulukian for super quick visibility but i think it's corect

src/routes/+layout.svelte
```diff
@@ -152,6 +156,10 @@
 		rel="manifest"
 		href="{PUBLIC_ORIGIN || $page.url.origin}{base}/{PUBLIC_APP_ASSETS}/manifest.json"
 	/>
+
+	{#if PUBLIC_PLAUSIBLE_DOMAIN}
+		<script defer data-domain={PUBLIC_PLAUSIBLE_DOMAIN} src="/js/script.js"></script>
+	{/if}
```
coyotte508, line 162:
```suggestion
	{#if PUBLIC_PLAUSIBLE_SCRIPT_URL}
		<script defer data-domain={new URL(PUBLIC_ORIGIN).hostname} src={PUBLIC_PLAUSIBLE_SCRIPT_URL}></script>
	{/if}
```

something like this IMO, so other users of HF chat don't have to have a proxy

nsarrazin: Ah that's fair thanks, will update the variables for this

.env.template
```diff
@@ -235,6 +235,7 @@ RATE_LIMIT=16
 MESSAGES_BEFORE_LOGIN=5# how many messages a user can send in a conversation before having to login. set to 0 to force login right away
 
 PUBLIC_GOOGLE_ANALYTICS_ID=G-8Q63TH4CSL
+PUBLIC_PLAUSIBLE_SCRIPT_URL="/js/script.js"
```
coyotte508, line 238:
```suggestion
PUBLIC_PLAUSIBLE_SCRIPT_URL="https://huggingface.co/js/script.js"
```

Maybe?

Because otherwise running inside the space (https://huggingface.co/spaces/huggingchat/chat-ui) it won't work (maybe not important)

coyotte508 (approved):

.env.template
nsarrazin: I think most usage goes through the proxied version but doesn't hurt to have it :)

julien-c: well inside the standard Space URL it's already tracked at the hub page level anyways

So we'll have twice the same script (inside and outside the iframe)

PR #783, opened 2024-02-05
[Assistants] Fix/clear searchParams while changing models

mishig25:
Fixes a bug: pagination query needs to be removed when changing a model. Otherwise, it can result in a query "?modelId={someModel}&p={someNumber}" & depending on that `someNumber` it can result in empty models page since the backend makes mongo cursor skips based on `someNumber`

https://github.com/huggingface/chat-ui/assets/11827707/44d5ee3f-1126-4e81-837b-ec21243ba42b


gary149 (approved):

mishig25:
due to partial reload https://github.com/huggingface/chat-ui/pull/783/commits/8cd7b98d4fa4ceb16f649714fc8d56d811a9c39d is also needed

Issue #770, opened 2024-02-03
[Assistants] Start chatting button freezes app


gururise:
#### Local Install is unable to start chatting with assistants

Steps:
1. Create a new assistant (and manually set users to 2 and make it featured)
2. Browse to Popular 'Assistants' page.
3. Select newly created assistant.
4. Click "Start chatting"

Once you click the "Start chatting" button, the page freezes and clicking the "Cancel' button no longer works.  If you click the cancel button before you click the "Start chatting" button, you can dismiss the dialog. The only way to get out of this page is to change the URL once it freezes.

Here is a new assistant I created:
![image](https://github.com/huggingface/chat-ui/assets/4380009/afc41046-8a1c-4ce2-a9e0-ecc97a44af0e)

At this dialog, clicking "Start chatting" will freeze the app, making the cancel button not work anymore:
![image](https://github.com/huggingface/chat-ui/assets/4380009/27bc07c8-1bdd-42c5-9790-32f91e6c2715)

**NOTE: Chatting with an assistant by using the settings menu and making the assistant ACTIVE works fine**:
![image](https://github.com/huggingface/chat-ui/assets/4380009/835f7706-6366-46fc-bf3b-fdaf2d0be8d2)

johndpope:
my chat is (partially) working with mistralai/Mistral-7B-Instruct-v0.1 • 
![Screenshot from 2024-02-05 06-22-39](https://github.com/huggingface/chat-ui/assets/289994/716be72b-0551-4e63-b080-74ef24a7dcb3)
though I don't see assistants on side / (have to go directly to /assistants) and I lose them after I close them.
![Screenshot from 2024-02-05 06-21-49](https://github.com/huggingface/chat-ui/assets/289994/906fc2c5-67b3-4ee1-b21b-7323bc06abf2)

gururise:
Pulled Latest changes and this bug still exists.

PR #785, opened 2024-02-06
[fix] If APP_BASE is empty, go to '/' instead of an empty string.

gururise:
If the APP_BASE env is an empty string, you won't be able to launch an assistant from the "Popular Assistants" page by clicking 'Start chatting'. This fixes that problem.

Fixes issue #770


minh-v (approved):

nsarrazin (approved):
Thanks for the contrib & sorry for the delay in reviewing this. :hugs:

PR #790, opened 2024-02-06
[Chat input] Blur on submit

### LINKED ISSUE #362: Mobile: close keyboard on submit
It can probably work by just using [blur](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/blur) on submit (only on mobile).

mishig25:
Closes #362

### DIFF
--- a/src/lib/components/chat/ChatInput.svelte
+++ b/src/lib/components/chat/ChatInput.svelte
@@ -21,6 +21,7 @@
 		// submit on enter
 		if (event.key === "Enter" && !event.shiftKey) {
 			event.preventDefault();
+			textareaElement.blur();
 			dispatch("submit"); // use a custom event instead of `event.target.form.requestSubmit()` as it does not work on Safari 14
 		}
 	}
--- a/src/lib/components/chat/ChatInput.svelte
+++ b/src/lib/components/chat/ChatInput.svelte
@@ -21,7 +21,7 @@
 		// submit on enter
 		if (event.key === "Enter" && !event.shiftKey) {
 			event.preventDefault();
-			// call blur to close keyboard on mobile
+			// blur to close keyboard on mobile
 			textareaElement.blur();
 			// refocus so that user on desktop can start typing without needing to reclick on textarea
 			textareaElement.focus();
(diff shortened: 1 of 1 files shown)


nsarrazin (approved):

mishig25:
dont merge yet

mishig25:
done. wdyt of https://github.com/huggingface/chat-ui/pull/790/commits/29f10020e36465ef8a1f0cfefaf623b884f06551 ?

nsarrazin (approved):
Tested locally, works well like this I think on mobile & desktop.

mishig25:
@gary149 added this line https://github.com/huggingface/chat-ui/pull/790/commits/ddb4fe090c805da8a1de65521fadd717adbf760c based on our slack discussion

src/lib/components/chat/ChatInput.svelte
```diff
@@ -21,6 +21,12 @@
 		// submit on enter
 		if (event.key === "Enter" && !event.shiftKey) {
 			event.preventDefault();
+			// blur to close keyboard on mobile
+			textareaElement.blur();
+			// refocus so that user on desktop can start typing without needing to reclick on textarea
+			if (innerWidth > TABLET_VIEWPORT_WIDTH) {
```
gary149, line 27: What if we do a touch device check instead of screen width?  👀

nsarrazin: I think with `navigator.maxTouchPoints`  you'll have false positives on laptops with touch screens for example. But with the viewport solution you'll also have false positives on desktop browsers if they're sized super small. 

I think we would need the [VirtualKeyboard API](https://developer.mozilla.org/en-US/docs/Web/API/VirtualKeyboard_API) which is [not supported very well](https://developer.mozilla.org/en-US/docs/Web/API/VirtualKeyboard_API#browser_compatibility), or tracking changes in the [availHeight](https://developer.mozilla.org/en-US/docs/Web/API/Screen/availHeight) which feels overkill. So it's just which compromise we want to pick, I think viewport width is fine

gary149:
Mhh on mobile the keyboard is already closing in prod? hf.co/chat - or am I missing something?

gary149 (approved):

PR #794, opened 2024-02-06
Feature checkbox

minh-v:
Added checkbox button to create assistant page to set `featured` flag.
Also sets `userCount` to 2 if featured flag was set.

image


gary149:
Thanks for your PR, but for now we are manually checking/adding the feature flag to make sure featured assistants are always respectful/safe (we will automate this pipeline, but it will take some time). Assistants with a higher user count are reviewed first btw.

PR #526, opened 2023-10-20
Bring back disclaimer modal & updates to guest mode

nsarrazin:
## Changes

New modals flow:
1. First modal shows up on first page load
  a. If login required but no guest messages allowed, show only the "sign in" button
  b. If login required with guest messages, show "sign in" AND "try as guest" buttons
  c. if no login required, show  "start chatting" button

2. If guest mode is enabled, show second modal when message limit is reached, asking to log-in 

Guest mode limit is now calculated based on TOTAL number of AI responses across all conversations. 

### Fixes
* Flipped gradient orientation according to figma
* Got rid of version number on modal according to figma

## Screenshots
First modal
image
Second modal
image

## To test
1. pull branch
2. `npm run updateLocalEnv`
3. `npm run dev`

PR #796, opened 2024-02-07
[Login Modal] Fix not showing properly on mobile

mishig25:
Logic from https://github.com/huggingface/chat-ui/pull/526 was good/working as expected. However, on mobile, the login modal was not showing up after user has reached the limit of messages.

| before | with this pr |
|---------|---------|
|   |   |

### DIFF
--- a/src/lib/components/chat/ChatWindow.svelte
+++ b/src/lib/components/chat/ChatWindow.svelte
@@ -212,7 +212,7 @@
 								placeholder="Ask anything"
 								bind:value={message}
 								on:submit={handleSubmit}
-								on:keypress={(ev) => {
+								on:beforeinput={(ev) => {
 									if ($page.data.loginRequired) {
 										ev.preventDefault();
 										loginModalOpen = true;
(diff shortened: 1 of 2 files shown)


src/lib/components/chat/ChatWindow.svelte
```diff
@@ -212,7 +212,7 @@
 								placeholder="Ask anything"
 								bind:value={message}
 								on:submit={handleSubmit}
-								on:keypress={(ev) => {
```
mishig25, line 215: `keypress` was not firing on mobile. We would need more general event (i.e. `beforeinput`). It also means that on desktop: user can paste a text & would not receive LoginModal since paste event would not fire keypress. `beforeinput` would fire on all forms of input: including keypress & paste

mishig25: https://caniuse.com/?search=beforeinput

image

gary149: Yes, my bade, I misread the change.

gary149 (approved):
ok

nsarrazin (approved):
Looks good to me, we could do it everywhere maybe

mishig25:
Merged this one. Let's check the other instances of keypress in the codebase

Issue #338, opened 2023-07-10
Searxng as search provider [free, open source]


flozi00:
https://github.com/searxng/searxng
https://searx.space/

Searx would be an free alternative for personal use to the serp providers.
What do you think about integrating it ?
I could start an PR for that if you are open for

nsarrazin:
Hey! I think it would be nice to support an open source search tool for sure. Feel free to open a PR with an integration and let me know if you need help.

flozi00:
Great, I will add it tomorrow

It's only one file, correct ? https://github.com/huggingface/chat-ui/blob/main/src/lib/server/websearch/searchWeb.ts

nsarrazin:
Thanks! Yes it's this file, you can also add an empty entry to the `.env` so people know it's a field that can be filled.

RachelShalom:
Hey after adding the serper api key I'm trying to run the app locally "nmp run dev" and I get an issue related to websearch:

[vite]: Rollup failed to resolve import "@xenova/transformers" from "C:/Users/username/chat-ui/src/lib/server/websearch/sentenceSimilarity.ts".
This is most likely unintended because it can break your application at runtime.
If you do want to externalize this module explicitly add it to
`build.rollupOptions.external`
error during build:
Error: [vite]: Rollup failed to resolve import "@xenova/transformers" from "C:/Users/username/chat-ui/src/lib/server/websearch/sentenceSimilarity.ts".
This is most likely unintended because it can break your application at runtime.
If you do want to externalize this module explicitly add it to
`build.rollupOptions.external`
    at viteWarn (file:///C:/Users/username/chat-ui/node_modules/vite/dist/node/chunks/dep-df561101.js:48142:27)   
    at onRollupWarning (file:///C:/Users/username/chat-ui/node_modules/vite/dist/node/chunks/dep-df561101.js:48174:9)
    at onwarn (file:///C:/Users/username/chat-ui/node_modules/vite/dist/node/chunks/dep-df561101.js:47902:13)     
    at file:///C:/Users/username/chat-ui/node_modules/rollup/dist/es/shared/node-entry.js:24152:13
    at Object.logger [as onLog] (file:///C:/Users/username/chat-ui/node_modules/rollup/dist/es/shared/node-entry.js:25825:9)
    at ModuleLoader.handleInvalidResolvedId (file:///C:/Users/rachel_shalom/chat-ui/node_modules/rollup/dist/es/shared/node-entry.js:24738:26)
[...]

PR #805, opened 2024-02-09
Added support for SearXNG as a websearch source

Karlostavitch1:
[SearXNG](https://github.com/searxng/searxng) is an open-source, privacy-centric search aggregator. With this single integration, it provides access to over 50 search engines with results in JSON format (if enabled on the instance). There are public instances of SearXNG and it is easily self-hosted.

I note that there has been some [discussion already about this feature](https://github.com/huggingface/chat-ui/issues/338).

This is a feature that's been very helpful for me and I hope you find it worthwhile for the wider project.

Created searchSearxng.ts for querying and parsing searxng instance.
Added SearXNG to the searchWeb.ts to ensure inclusion in the search initialisation script. 
Added SearXNG to +layout.server.ts to ensure the search button is enabled if the URL provided in .env.local assignment 
Updated .env to include SearXNG configuration
Updated README.md under Web Search Config to mention the addition of SearXNG URL use in .env

### DIFF
--- a/.env
+++ b/.env
@@ -18,6 +18,11 @@ SERPER_API_KEY=#your serper.dev api key here
 SERPAPI_KEY=#your serpapi key here
 SERPSTACK_API_KEY=#your serpstack api key here
 USE_LOCAL_WEBSEARCH=#set to true to parse google results yourself, overrides other API keys
+USE_LOCAL_SEARXNG_URL=#your SearXNG URL eg https://etsi.me (ensure json enabled on instance settings.yml)
+
+#SearXNG configuration for public instances (or use default settings in self hosted instances).
+SEARXNG_CATEGORIES=#comma seperated general,images,videos,news,map,music,it,science,files,social_media (or undefined for instance defaults).
+SEARXNG_ENGINES=#define specific engines or undefined for instance defaults (https://docs.searxng.org/user/configured_engines.html#configured-engines).
 
 WEBSEARCH_ALLOWLIST=`[]` # if it's defined, allow websites from only this list.
 WEBSEARCH_BLOCKLIST=`[]` # if it's defined, block websites from this list.
--- a/.env
+++ b/.env
@@ -18,11 +18,7 @@ SERPER_API_KEY=#your serper.dev api key here
 SERPAPI_KEY=#your serpapi key here
 SERPSTACK_API_KEY=#your serpstack api key here
 USE_LOCAL_WEBSEARCH=#set to true to parse google results yourself, overrides other API keys
-USE_LOCAL_SEARXNG_URL=#your SearXNG URL eg https://etsi.me (ensure json enabled on instance settings.yml)
-
-#SearXNG configuration for public instances (or use default settings in self hosted instances).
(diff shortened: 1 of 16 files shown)


.env
```diff
@@ -18,6 +18,11 @@ SERPER_API_KEY=#your serper.dev api key here
 SERPAPI_KEY=#your serpapi key here
 SERPSTACK_API_KEY=#your serpstack api key here
 USE_LOCAL_WEBSEARCH=#set to true to parse google results yourself, overrides other API keys
+USE_LOCAL_SEARXNG_URL=#your SearXNG URL eg https://etsi.me (ensure json enabled on instance settings.yml)
```
nsarrazin, line 21:
```suggestion
SEARXNG_URL=#your SearXNG URL eg https://etsi.me (ensure json enabled on instance settings.yml)
```

maybe ?

.env
```diff
@@ -18,6 +18,11 @@ SERPER_API_KEY=#your serper.dev api key here
...
+
+#SearXNG configuration for public instances (or use default settings in self hosted instances).
+SEARXNG_CATEGORIES=#comma seperated general,images,videos,news,map,music,it,science,files,social_media (or undefined for instance defaults).
```
nsarrazin, line 24: Is the customization needed or can we have some sane defaults that are used across the app ?

I feel like it's a lot of new env variables that most people won't need. Same for `SEARXNG_ENGINES` but i'm not super familiar with the project so please tell me if they're needed.

nsarrazin: I feel like given the parameters are just injected into the URL as query parameters, could we rewrite this so that `SEARXNG_URL` already contains the query parameters for engines, categories, etc. if people want to set them ?

Would require a bit of URL parsing logic to correctly append the query, but it would make it possible to have only one env variable for this. what do you think?

nsarrazin (commented):
Left a few comments, looks good to me otherwise! Let me know what you think.

.env
Karlostavitch1: I'll try to get these into a single string.

.env
Karlostavitch1: Looks good, I'll update.

.env
Karlostavitch1: > Is the customization needed or can we have some sane defaults that are used across the app ?
> 
> I feel like it's a lot of new env variables that most people won't need. Same for `SEARXNG_ENGINES` but i'm not super familiar with the project so please tell me if they're needed.

I think engines is important but could do away with categories (as its  engine grouping).

Karlostavitch1:
Hi nsarrazin,

I've attempted to accommodate your requests, please make any minor edits you wish or let me know if there are any issues.

Thanks,

nsarrazin:
Hey, thank you for the changes, I think it's very nice like this! I'll fix the linting and expand a bit the readme section to explain the `<query>`, but this is great!

nsarrazin:
Feel free to push to this branch for SearXNG! For Serper and Serpapi, I think it was on purpose since these only search google. For SearXNG it's different I think so it makes sense in that case.

Karlostavitch1:
was this a transient issue (docker auth) or do you need me to debug it?

nsarrazin (approved):
Looks very nice, thanks for the contribution!

PR #806, opened 2024-02-09
Fix delete assistant query performance

nsarrazin:
It was running on all users settings which was slow in prod


nsarrazin:
Did you test it only in prod? The current query in prod takes very long because it has to pull from every user settings, since there was no filter, so it likely times out before it redirects you. (this is what this PR fixes)

The form action should [redirect you to the settings already](https://github.com/huggingface/chat-ui/blob/9f7119c358f1c24b2c2235544e352a13a0edf3f0/src/routes/settings/assistants/%5BassistantId%5D/%2Bpage.server.ts#L54C1-L54C43) though.

mishig25:
Yes. I tested it in prod. I see

mishig25 (approved):
lgtm !

nsarrazin:
Looks like it worked! We should still add some kind of feedback while it loads but it no longer times out at all :grin:

mishig25:
Yep, just tested as well

PR #807, opened 2024-02-09
Conversation trees (#223)

### LINKED ISSUE #223: Adding support for conversation threads
Might be valuable to allow the user to branch out into separate slack-like threads. This lets you keep the main convo rolling while allowing you to dig deeper into a specific answer.

image

nsarrazin:
https://github.com/huggingface/chat-ui/assets/25119303/6f778bbb-2094-4cb1-865c-a16fdfc67605

## Conversation trees

It's a big refactoring because we have to structure messages like a tree instead of a list, which had implications across the codebase. The way I built it is to add the following two fields on Message:

```diff
export type Message = Partial<Timestamps> & {
	from: "user" | "assistant";
	id: ReturnType<typeof crypto.randomUUID>;
	content: string;
	updates?: MessageUpdate[];
	webSearchId?: WebSearch["_id"]; // legacy version
	webSearch?: WebSearch;
	score?: -1 | 0 | 1;
	files?: string[]; // can contain either the hash of the file or the b64 encoded image data on the client side when uploading
	interrupted?: boolean;

+	ancestors?: Message["id"][];
+	children?: Message["id"][];
};
```

`children` are used to display the alternative choices in the UI, and `ancestors` are used to build the entire subtree for building the prompt.

#### Helpers
I added a lot of helpers functions under `utils/tree` which have associated unit tests. Test conciseness could be improved but they cover my use cases so that's good enough for me for now 😁  

#### Legacy conversations
I didn't want to run a one-off migration for the whole DB because I wanted the change to be compatible with every chat-ui setup out there without manual intervention, so what I ended up doing is converting legacy conversations on the fly when they are being loaded. I have a helper function with tests for it, seems to work well. 

#### Misc changes
I refactored `buildPrompt` and broke it down into `preprocessMessages` (inject websearch results and images for multimodal) and `buildPrompt`  (turn an array of messages into a string prompt).

I moved the prompt building out of the inference-specific code and into the POST endpoint for new messages. (less code duplication) 

### To-Do

- [x] Fix client-side updates when generating a new answer (messages render well when the webpage is loaded  but the logic for optimistic updates hasn't been added yet so it doesn't load while streaming)
- [x] Keep track of current tree when load function invalidates data (currently you get reset to every first children every time data is invalidated)
- [x] Fix websearch updates disappearing
- [x] Double loading icon when waiting for message
- [x] Streaming doesn't work for first conversation message
- [x] Fix shared conversations (they display fine but posting a new message or retrying doesn't work)
- [x] Replace the buttons for each children (`1`, `2`, `3`) by arrows (`<`/ `>`), or not ? cc @gary149 
- [x] Make sure you see the new branch when retrying/editing

Closes #223


gary149:
Nice! maybe @coyotte508 can take a quick look at the data model?

coyotte508:
Looks nice / well-thought-out to me!

---

Something we may want to consider, in the future, is if we don't want to load the full tree of messages every time / move the `Message` into their own collection.

In that case we could have add `Conversation` just a single array of message ids for the default list of messages displayed, and other paths loaded async.

(migration from current PR to this model would be straightforward)

nsarrazin:
Yeah I think we could do a refactoring pass where we fetch only the current subtree directly in mongo by using a query w/ filters to reduce work on the node backend & the client.

Could be done by moving `Message` into their own collections (or just querying against `"conv.messages"` directly in mongo? not sure if that would be bad for performance, maybe not with the right index?)

Would require quite a bit of front-end work to handle the streaming of new paths, so i guess let's do it if we notice performance is bad on big conversations. 👀

nsarrazin:
Replaced the buttons by arrows 
image

gary149 (approved):
🚀

src/lib/server/database.ts
```diff
@@ -62,6 +62,12 @@ client.on("open", () => {
 			{ partialFilterExpression: { userId: { $exists: true } } }
 		)
 		.catch(console.error);
+	conversations
+		.createIndex(
+			{ "message.id": 1, "message.ancestors": 1 },
+			{ partialFilterExpression: { userId: { $exists: true } } }
+		)
+		.catch(console.error);
```
coyotte508, line 70: Where is this index used? :thinking:

PR #808, opened 2024-02-09
[Privacy page] Skip disclaimer modal

mishig25:


src/routes/+layout.svelte
```diff
@@ -166,7 +166,7 @@
 	{/if}
 </svelte:head>
 
-{#if !$settings.ethicsModalAccepted}
+{#if !$settings.ethicsModalAccepted && $page.url.pathname !== "/privacy"}
```
mishig25, line 169: we can generalize the solution by creating a whitelist array if there are other pages the disclaimer modal should not appear. For now, this is good enough imo

alak (approved):

src/routes/+layout.svelte
nsarrazin: I think [layout groups](https://kit.svelte.dev/docs/advanced-routing#advanced-layouts-group) are meant for use cases like this if we want to handle it in the future, but indeed no need to overbuild for now 😄

mishig25: At the moment, I can't think of any other page that should not show disclaimer. Can you ?

nsarrazin: I think that's the only one for now 👀

julien-c:
i don't think it's fixed:

image

PR #809, opened 2024-02-09
/assistants: add user count

gary149:
image

### DIFF
--- a/src/lib/utils/formatUserCount.ts
+++ b/src/lib/utils/formatUserCount.ts
@@ -1,5 +1,6 @@
-export function formatUserCount(userCount: number): string | null {
+export function formatUserCount(userCount: number): string {
 	const userCountRanges: Array<{ min: number; max: number; label: string }> = [
+		{ min: 0, max: 1, label: "1" },
 		{ min: 2, max: 9, label: "1-10" },
 		{ min: 10, max: 99, label: "+10" },
 		{ min: 100, max: 299, label: "+100" },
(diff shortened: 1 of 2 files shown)


julien-c:
love those ranges

src/lib/utils/formatUserCount.ts
```diff
@@ -0,0 +1,18 @@
+export function formatUserCount(userCount: number): string {
+	const userCountRanges: Array<{ min: number; max: number; label: string }> = [
```
mishig25, line 2:
```suggestion
	const userCountRanges: { min: number; max: number; label: string }[] = [
```

src/lib/utils/formatUserCount.ts
```diff
@@ -0,0 +1,18 @@
...
+		{ min: 100, max: 299, label: "+100" },
+		{ min: 300, max: 499, label: "+300" },
+		{ min: 500, max: 999, label: "+500" },
+		{ min: 1_000, max: 2_999, label: "+1k" },
+		{ min: 3_000, max: 4_999, label: "+3k" },
+		{ min: 5_000, max: 9_999, label: "+5k" },
+		{ min: 10_000, max: Infinity, label: "+10k" },
+	];
+
+	const range = userCountRanges.find(({ min, max }) => userCount >= min && userCount <= max);
```
mishig25, line 16:
```suggestion
	const range = userCountRanges.find(({ max }) => userCount <= max);
```

up to you whether to accept this suggestion, which is more implicit. We can just ignore `min`

mishig25 (approved):

PR #810, opened 2024-02-09
[Assistants] Add new indices for efficient querying

mishig25:
Read[ this comment](https://github.com/huggingface/chat-ui/pull/810/files#r1484441671)

### DIFF
--- a/src/lib/server/database.ts
+++ b/src/lib/server/database.ts
@@ -75,10 +75,9 @@ client.on("open", () => {
 	messageEvents.createIndex({ createdAt: 1 }, { expireAfterSeconds: 60 }).catch(console.error);
 	sessions.createIndex({ expiresAt: 1 }, { expireAfterSeconds: 0 }).catch(console.error);
 	sessions.createIndex({ sessionId: 1 }, { unique: true }).catch(console.error);
-	assistants.createIndex({ createdBy: 1 }).catch(console.error);
+	assistants.createIndex({ createdById: 1, userCount: -1 }).catch(console.error);
 	assistants.createIndex({ userCount: 1 }).catch(console.error);
-	assistants.createIndex({ featured: 1 }).catch(console.error);
-	assistants.createIndex({ modelId: 1, userCount: -1, createdByName: 1 }).catch(console.error);
-	assistants.createIndex({ modelId: 1, userCount: -1, featured: 1 }).catch(console.error);
+	assistants.createIndex({ featured: 1, userCount: -1 }).catch(console.error);
+	assistants.createIndex({ modelId: 1, userCount: -1 }).catch(console.error);
 	reports.createIndex({ assistantId: 1 }).catch(console.error);
 });
--- a/src/lib/server/database.ts
+++ b/src/lib/server/database.ts
@@ -75,5 +75,7 @@ client.on("open", () => {
 	assistants.createIndex({ createdBy: 1 }).catch(console.error);
 	assistants.createIndex({ userCount: 1 }).catch(console.error);
 	assistants.createIndex({ featured: 1 }).catch(console.error);
+	assistants.createIndex({ modelId: 1, userCount: -1, createdByName: 1 }).catch(console.error);
+	assistants.createIndex({ modelId: 1, userCount: -1, featured: 1 }).catch(console.error);
 	reports.createIndex({ assistantId: 1 }).catch(console.error);
 });
--- a/src/routes/assistants/+page.server.ts
+++ b/src/routes/assistants/+page.server.ts
@@ -18,9 +18,12 @@ export const load = async ({ url, locals }) => {
 	const username = url.searchParams.get("user");
 	const createdByCurrentUser = locals.user?.username && locals.user.username === username;
 
-	let user: User | null = null;
+	let user: Pick<User, "_id"> | null = null;
 	if (username) {
-		user = await collections.users.findOne({ username }, { projection: { _id: 1 } });
+		user = await collections.users.findOne<Pick<User, "_id">>(
+			{ username },
+			{ projection: { _id: 1 } }
+		);
 		if (!user) {
 			throw error(404, `User "${username}" doesn't exist`);
 		}
--- a/src/routes/assistants/+page.server.ts
+++ b/src/routes/assistants/+page.server.ts
@@ -2,6 +2,7 @@ import { base } from "$app/paths";
 import { ENABLE_ASSISTANTS } from "$env/static/private";
 import { collections } from "$lib/server/database.js";
 import type { Assistant } from "$lib/types/Assistant";
+import type { User } from "$lib/types/User";
 import { error, redirect } from "@sveltejs/kit";
 import type { Filter } from "mongodb";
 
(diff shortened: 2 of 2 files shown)


src/lib/server/database.ts
```diff
@@ -75,5 +75,7 @@ client.on("open", () => {
 	assistants.createIndex({ createdBy: 1 }).catch(console.error);
 	assistants.createIndex({ userCount: 1 }).catch(console.error);
 	assistants.createIndex({ featured: 1 }).catch(console.error);
+	assistants.createIndex({ modelId: 1, userCount: -1, createdByName: 1 }).catch(console.error);
+	assistants.createIndex({ modelId: 1, userCount: -1, featured: 1 }).catch(console.error);
 	reports.createIndex({ assistantId: 1 }).catch(console.error);
```
mishig25, line 80: we have [this filter](https://github.com/huggingface/chat-ui/blob/assistans_index/src/routes/assistants/%2Bpage.server.ts#L27-L32), which is
```ts
const filter: Filter<Assistant> = {
	modelId: modelId ?? { $exists: true },
	...(!createdByCurrentUser && { userCount: { $gt: 1 } }),
	...(createdByName ? { createdByName } : { featured: true }),
};
```

coyotte508: Looking at `Assistant.ts`, `modelId` is always defined.

So the filter should be like this:

```ts
const filter: Filter<Assistant> = {
	...(modelId && {modelId}),
	...(!createdByCurrentUser && { userCount: { $gt: 1 } }),
	...(createdByName ? { createdByName } : { featured: true }),
};
...
.sort({ userCount: -1 })
```

There is already an index on `createdBy`, but `createdBy` doesn't exist in `Assistant.ts`, there is instead `createdByName` and `createdById`.

IMO:

- Change the `createdBy` index to `{createdById: 1, userCount: -1}`
- Change the filter to use `createdById` if possible (by doing a first query to find user matching name). Otherwise add an index `{createdByName: 1, userCount: -1}`
- Change the `featured` index to `{featured: 1, userCount: -1}`
- Create an index on `{modelId: 1, userCount: -1}`. No need for a more specific index at the moment, if we add a more specific index the other key should be added _before_ `userCount`, eg `{modelId: 1, featured: 1, userCount: -1`). It's important that the "sort" key is last.

mishig25: great suggestions. Will apply those.

> if we add a more specific index the other key should be added before userCount, eg {modelId: 1, featured: 1, userCount: -1). It's important that the "sort" key is last.

I don't see any information "the "sort" key is last" on the docs [here](https://www.mongodb.com/docs/manual/tutorial/sort-results-with-indexes/#sort-on-multiple-fields). Why so ?

coyotte508: Check out this video in their resources: https://learn.mongodb.com/learn/course/mongodb-indexes/lesson-4-working-with-compound-indexes-in-mongodb/learn?page=1 

Basically in a compound index you want to:

- have an exact match on key1, key2, ... keyN
- sort on keyN+1

For example, if you define an index `{featured: 1, userCount: -1}`, and you have those documents:

```ts
{_id: "id0", featured: true, userCount: 1},
{_id: "id1", featured: true, userCount: 1},
{_id: "id2", featured: true, userCount: 2},
{_id: "id3", featured: true, userCount: 3},
{_id: "id4", featured: false, userCount: 1},
{_id: "id5", featured: false, userCount: 2},
{_id: "id6", featured: false, userCount: 3},
```

The index will have this internal structure:

```
{
  [featured=true]: {
    [userCount=3]: ["id3"],
    [userCount=2]: ["id2"],
    [userCount=1]: ["id1", "id0"]
  },
 [featured=false]: {
    [userCount=3]: ["id6"],
    [userCount=2]: ["id5"],
    [userCount=1]: ["id4"]
  },
}
```

So if you do the query `.find({featured: true}).sort({userCount: -1})`, mongodb will be able to look up the desired results very efficiently: `index[featured=true].entries().flatMap(([, v] => v)`.

If the index was defined `({userCount: -1, featured: 1})`  the internal index structure would look like this:

```
{
  [userCount=3]: {
    [featured=true]: ["id3"],
    [featured=false]: ["id6"],
  },
  [userCount=2]: {
    [featured=true]: ["id2"],
    [featured=false]: ["id5"],
  },
  [userCount=1]: {
    [featured=true]: ["id1", "id0"],
    [featured=false]: ["id4"],
  },
}
```

and MongoDB would have to examine all index keys to return the results: 

```
index.entries().flatMap(([, v]) => v.entries().filter(k => k === "featured=true").flatMap([, v] => v);
```

It will examine all 7 indexed values instead of only the 4 returned.

This rule - putting "sort" last - is only when we have an _exact_ match on the previous keys. It's more complicated otherwise.

coyotte508: In general https://learn.mongodb.com/ is an awesome resource to learn from - but the format is videos

mishig25: @coyotte508 wdyt https://github.com/huggingface/chat-ui/pull/810/commits/2e87ab517c70c3697d4905ff23f370a252305725 ?

src/lib/server/database.ts
```diff
@@ -75,8 +75,9 @@ client.on("open", () => {
 	messageEvents.createIndex({ createdAt: 1 }, { expireAfterSeconds: 60 }).catch(console.error);
 	sessions.createIndex({ expiresAt: 1 }, { expireAfterSeconds: 0 }).catch(console.error);
 	sessions.createIndex({ sessionId: 1 }, { unique: true }).catch(console.error);
-	assistants.createIndex({ createdBy: 1 }).catch(console.error);
+	assistants.createIndex({ createdById: 1, userCount: -1 }).catch(console.error);
 	assistants.createIndex({ userCount: 1 }).catch(console.error);
```
mishig25, line 79: with the newly added indices, I don't think we need this index. Do we? And even if we need it, we should change it to `-1` ? cc: @coyotte508 @nsarrazin

mishig25: I'm talking about this one
```ts
assistants.createIndex({ userCount: 1 }).catch(console.error);
```

coyotte508: If the query can be just `{userCount: {$gt: 1}}`, then we should keep it.

`-1` or `1` doesn't have a difference in indexes with only one key (mongo can always "reverse" the index). In compound indexes, it makes a difference

coyotte508 (approved):

src/lib/server/database.ts
mishig25: yes, query can be just `{userCount: {$gt: 1}}`. Therefore, keeping it

nsarrazin:
I don't have enough mongo knowledge to chime in so if it's good for @coyotte508 then it's good for me :grin: 

Do we need to manually build the indices in atlas or will they be picked up on space rebuild ?

coyotte508:
It will automatically build

We should delete the removed indicies manually though

mishig25:
these indices were deleted:
```ts
assistants.createIndex({ createdBy: 1 }).catch(console.error);
assistants.createIndex({ featured: 1 }).catch(console.error);
```

Issue #425, opened 2023-08-31
Is it possible to modify it so that .env.local environment variables are set at runtime?


martinkozle:
Currently for every different deployment of Chat-UI it is required to rebuild the Docker image with different .env.local environment variables. Is it theoretically possible to have it so that 1 image can be used for all deployments, but with different secrets passed at runtime? What environment variables and for what reason are truly needed at build time for Chat-UI to function? In #204 it says `HF_ACCESS_TOKEN` is needed at build time, but what if we use `OPENID` authentication instead? Is there anything else blocking this type of use case?

nsarrazin:
Yeah it's kind of annoying, I agree. I think you would need to replace the secrets from `$env/static/private` to `$env/dynamic/private`. Not sure what the side effects would be, but I can't see why this should change anything. It could be worth trying!

haje01:
I agree too. Baking environment variables directly into a Docker image is bad practice:
- Lack of Flexibility
- Security Risks
- Portability Issues

zacps:
Opened a PR for this, needs a rebase but I'm waiting on someone to have time to review it.

PR #813, opened 2024-02-12
Load configuration at runtime instead of build time

zacps:
This will make it possible to publish self contained docker images which can then be used in simple docker/kubernetes deployments.

This is a *breaking change*, ~~in particular private environment variables must now have the `VITE_` prefix, and~~ users will need to make sure environment variables are available at runtime.

Fixes https://github.com/huggingface/chat-ui/issues/425


anly2:
Instead of renaming the variables with `VITE_` wouldn't it be better to improve the usages to follow https://kit.svelte.dev/docs/modules#$env-dynamic-private

There will be import adjustments, but it will be more intuitive.
(And no variable renames expected)

If secrets are made dynamic, then the docker images will stay "env agnostic" and can be built once.
(I know feature flags are also different across environments, but those are not really a problem like secrets)

anly2:
I am no expert at all either, but both the documentation and this tutorial don't have anything particularly special: https://learn.svelte.dev/tutorial/env-dynamic-private

So wouldn't expect renames, but changing the usage places and imports.

zacps:
You are correct, I must have had something else miss-configured when I thought the prefix was required. 

Fixed!

anly2:
Admittedly, I couldn't quite get it to work myself either, so I feel your pain.

(I tried updating just the mongodb vars, and with the dynamic vars it builds and runs locally just fine.
But when I build a docker image and provide the `.env.local` outside, I fail. I suspect it is something to do with how I mount the env file, but haven't gotten to the bottom of it)

anly2:
TL;DR: You only need to fix the `MONGODB_URL`

----
As a follow up to my previous comment, I actually made it work with relatively few changes.

The only problem really was the bloody `MONGODB_URL` check that throws!
I defaulted that in the `.env` file and everything else was fine...


Only annoying thing was (which could just be me doing things wrong) that I expected the `$env/static/private` and the `$env/dynamic/private` to overlap but they don't seem to?!
I expect if I provide it in the `.env.local` file and use `/*dynamic*/  env.MONGODB_URL` that I would get the value. But I didn't.
So I worked around this with these ugly fallbacks: `const mongoUrl = env.MONGODB_URL || MONGODB_URL;` (which, again, I expected to already be the case)

zacps:
Just to clarify, did you need to change that in this PR? Or does it work for you as is?

anly2:
I had to do the following changes:

In `.env` give `MONGO_URL` an actual default value:
```
MONGODB_URL=mongodb://localhost:27017 # your mongodb URL here
```

And then I updated the `database.ts` file because I was actually providing the url as an actual env var (not as a line in a dotenv file):
```typescript
import { MONGODB_URL, MONGODB_DB_NAME, MONGODB_DIRECT_CONNECTION } from "$env/static/private";
import { env } from "$env/dynamic/private";

// ...

const mongoUrl = env.MONGODB_URL || MONGODB_URL; // I expected vars to work like this already!
if (!mongoUrl)  // throw ...

const client = new MongoClient(mongoUrl, {
	directConnection: (env.MONGODB_DIRECT_CONNECTION ?? MONGODB_DIRECT_CONNECTION) === "true",
});

// ...

const db = client.db(
	(env.MONGODB_DB_NAME ?? MONGODB_DB_NAME) + (import.meta.env.MODE === "test" ? "-test" : "")
);

// ...
```


First part made it build ok. (I really don't understand why the code is actually executed at build time)
Second part allowed me to override to the correct value at runtime.

zacps:
> import { MONGODB_URL, MONGODB_DB_NAME, MONGODB_DIRECT_CONNECTION } from "$env/static/private";

This line isn't in this PR. 

> const mongoUrl = env.MONGODB_URL || MONGODB_URL

This is unnecessary with the `$env/dynamic` imports.

zacps:
Moved the database check to runtime only.

The one outstanding issue I've encountered with this PR is that it doesn't currently make it possible to use a dynamic `APP_BASE`; I think this should be possible but probably requires a sveltekit bump (at least) to get access to the `reroute` hook.

zacps:
I've abandoned my attempt to make `APP_BASE` relative, it is exceptionally difficult to make work.

Happy to provide my notes if anyone else wants to attempt but for now this PR just leaves `APP_BASE` static.

nsarrazin:
Hey! Will close this PR as we handled it with #1092. Sorry for the double work, we needed to figure it out for our infra on HuggingChat. :sweat_smile: Appreciate the work on this

PR #820, opened 2024-02-13
Fix model identification when creating conversations

zacps:
PR #181 introduced a model ID field so that models could be switched out without breaking existing conversations. However, the new conversation endpoint incorrectly uses the model name instead of ID, so without this fix it was not possible to have a model with a different ID and name.

### DIFF
--- a/src/routes/conversation/+server.ts
+++ b/src/routes/conversation/+server.ts
@@ -39,7 +39,7 @@ export const POST: RequestHandler = async ({ locals, request }) => {
 		);
 	}
 
-	const model = models.find((m) => m.name === values.model);
+	const model = models.find((m) => (m.id || m.name) === values.model);
 
 	if (!model) {
 		throw error(400, "Invalid model");


nsarrazin:
Hi, thanks for the contribution! I'm not sure I fully understand what the code is supposed to do here. Could you give an example of an issue you were having that this solves?

I could imagine scenarios where `(m.id ?? m.name) === values.model);
` or ` m.id === values.model || m.name === values.model` would make sense but wouldn't `m.id || m.name === values.model` just return the first model with non-null ID?

zacps:
Yes! Thanks for catching that, I made mistake with precedence; fixed :).

I think `(m.id ?? m.name) === values.model);` and `(m.id || m.name) === values.model);` are functionally equivalent here but happy to use one over the other if you prefer.

Currently if you have a model configured with a different id and name, for example something like:

```
   {
      "id": "gpt-4-1106-preview",
      "name": "gpt-4",
      "displayName": "GPT 4 Turbo",
      "endpoints" : [{
         "type": "openai" ,
      }]
   },
```

You won't be able to create any conversations because the client sends the model id, `gpt-4-1106-preview`, and the server only searches the model names `["gpt-4"]`.

zacps:
@nsarrazin do you think you'll have a chance to review this soon? This is the last PR keeping us on a fork.

nsarrazin (approved):

nsarrazin:
Finally got around to reviewing this sorry! Will merge it now, thanks for the contrib @zacps :rocket:

PR #822, opened 2024-02-13
Allow customising defaultQuery for native Azure OpenAI support

zacps:
The api-version is a required query parameter for Azure OpenAI, by supporting this we can support it without a gateway like portkey.

---

P.S. I apologise for the flurry of PRs, please let me know if you'd like me to roll them up.


nsarrazin (approved):
Thank you for the contribution! This looks good to me

PR #825, opened 2024-02-14
Ask for a reason when reporting assistants

nsarrazin:
| Empty  | Filled reason |
| image|Screenshot from 2024-02-14 17-54-41 | 

Maxed reason to 128 chars, should work with the slack webhook integration.

### DIFF
--- a/src/lib/components/Modal.svelte
+++ b/src/lib/components/Modal.svelte
@@ -46,7 +46,7 @@
 		role="presentation"
 		tabindex="-1"
 		bind:this={backdropEl}
-		on:click={handleBackdropClick}
+		on:click|stopPropagation={handleBackdropClick}
 		transition:fade|global={{ easing: cubicOut, duration: 300 }}
 		class="fixed inset-0 z-40 flex items-center justify-center bg-black/80 p-8 backdrop-blur-sm dark:bg-black/50"
 	>
--- a/src/routes/settings/assistants/[assistantId]/+page.server.ts
+++ b/src/routes/settings/assistants/[assistantId]/+page.server.ts
@@ -53,7 +53,7 @@ export const actions: Actions = {
 
 		throw redirect(302, `${base}/settings`);
 	},
-	report: async ({ params, locals, url }) => {
+	report: async ({ request, params, locals, url }) => {
 		// is there already a report from this user for this model ?
 		const report = await collections.reports.findOne({
 			assistantId: new ObjectId(params.assistantId),
@@ -64,12 +64,19 @@ export const actions: Actions = {
 			return fail(400, { error: true, message: "Already reported" });
 		}
 
+		const reason = (await request.formData()).get("reportReason"); // get the reportReason and validate it
+
+		if (!reason || typeof reason !== "string" || reason.length > 128 || reason.length < 1) {
+			return fail(400, { error: true, message: "Invalid report reason" });
+		}
+
 		const { acknowledged } = await collections.reports.insertOne({
 			_id: new ObjectId(),
 			assistantId: new ObjectId(params.assistantId),
 			createdBy: locals.user?._id ?? locals.sessionId,
 			createdAt: new Date(),
 			updatedAt: new Date(),
+			reason,
 		});
 
 		if (!acknowledged) {
@@ -91,7 +98,7 @@ export const actions: Actions = {
 					"Content-type": "application/json",
 				},
 				body: JSON.stringify({
-					text: `Assistant <${assistantUrl}|${assistant?.name}> reported by <http://hf.co/${locals.user?.username}|${locals.user?.username}>`,
+					text: `Assistant <${assistantUrl}|${assistant?.name}> reported by <http://hf.co/${locals.user?.username}|${locals.user?.username}>. The following reason was given \n\n> ${reason}`,
 				}),
 			});
 
--- a/src/routes/settings/assistants/[assistantId]/+page.server.ts
+++ b/src/routes/settings/assistants/[assistantId]/+page.server.ts
@@ -5,7 +5,7 @@ import { authCondition } from "$lib/server/auth";
 import { base } from "$app/paths";
 import { PUBLIC_ORIGIN, PUBLIC_SHARE_PREFIX } from "$env/static/public";
 import { WEBHOOK_URL_REPORT_ASSISTANT } from "$env/static/private";
-
+import { z } from "zod";
 async function assistantOnlyIfAuthor(locals: App.Locals, assistantId?: string) {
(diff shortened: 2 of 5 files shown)


src/lib/components/Modal.svelte
```diff
@@ -46,7 +46,7 @@
 		role="presentation"
 		tabindex="-1"
 		bind:this={backdropEl}
-		on:click={handleBackdropClick}
+		on:click|stopPropagation={handleBackdropClick}
```
mishig25, line 49: just to be sure, this doesn't affect other use cases of `Modal`. Right ?

src/routes/settings/assistants/[assistantId]/ReportModal.svelte
```diff
@@ -0,0 +1,57 @@
+<script lang="ts">
```
Comment (mishig25) [inline src/routes/settings/assistants/[assistantId]/ReportModal.svelte:1]: up to you. Instead of creating a new component, we can just use [prompt](https://developer.mozilla.org/en-US/docs/Web/API/Window/prompt)

src/routes/settings/assistants/[assistantId]/ReportModal.svelte
```diff
@@ -0,0 +1,57 @@
...
+		}}
+		class="w-full min-w-64 p-4"
+	>
+		<span class="mb-1 text-sm font-semibold">Report an assistant</span>
+
+		<p class="text-sm text-gray-500">
+			Please provide a brief description of why you are reporting this assistant.
+		</p>
+
+		<input
```
Comment (mishig25) [inline src/routes/settings/assistants/[assistantId]/ReportModal.svelte:31]: maybe `textarea` would be a better choice since with `input`, you can't write multi-line texts. In this case, `textarea` should have max-height

src/routes/settings/assistants/[assistantId]/+page.server.ts
```diff
@@ -64,12 +64,19 @@ export const actions: Actions = {
 			return fail(400, { error: true, message: "Already reported" });
 		}
 
+		const reason = (await request.formData()).get("reportReason"); // get the reportReason and validate it
+
+		if (!reason || typeof reason !== "string" || reason.length > 128 || reason.length < 1) {
```
Comment (mishig25) [inline src/routes/settings/assistants/[assistantId]/+page.server.ts:69]: we can't do this validation with `zod`?

src/lib/components/Modal.svelte
nsarrazin: I tested our other modals and it doesn't, but I think we needed this line anyway: otherwise when you have two modals, clicking the backdrop would close all of them instead of the topmost one

src/routes/settings/assistants/[assistantId]/ReportModal.svelte
Comment (nsarrazin) [reply src/routes/settings/assistants/[assistantId]/ReportModal.svelte:1]: Ha, had I known about this, I think we could have gone for `prompt`, but now that the component is here, might as well use it I guess ? :laughing:

src/routes/settings/assistants/[assistantId]/ReportModal.svelte
Comment (nsarrazin) [reply src/routes/settings/assistants/[assistantId]/ReportModal.svelte:31]: I'd like to encourage people to keep it short, since it's 128 chars max but maybe a small textarea is still better indeed

src/routes/settings/assistants/[assistantId]/+page.server.ts
Comment (nsarrazin) [reply src/routes/settings/assistants/[assistantId]/+page.server.ts:69]: Fixed!

src/routes/settings/assistants/[assistantId]/ReportModal.svelte
Comment (julien-c) [reply src/routes/settings/assistants/[assistantId]/ReportModal.svelte:1]: (fwiw on the Hub we use window.prompt)

Comment (julien-c) [reply src/routes/settings/assistants/[assistantId]/ReportModal.svelte:1]: (but up to you)

Comment (nsarrazin) [reply src/routes/settings/assistants/[assistantId]/ReportModal.svelte:1]: then let's do it for consistency, will update :+1:

src/routes/settings/assistants/[assistantId]/+page.server.ts
```diff
@@ -64,9 +64,10 @@ export const actions: Actions = {
 			return fail(400, { error: true, message: "Already reported" });
 		}
 
-		const reason = (await request.formData()).get("reportReason"); // get the reportReason and validate it
+		const formData = await request.formData();
+		const result = z.string().min(1).max(128).safeParse(formData?.get("reportReason"));
 
-		if (!reason || typeof reason !== "string" || reason.length > 128 || reason.length < 1) {
+		if (!result.success) {
```
Comment (mishig25) [inline src/routes/settings/assistants/[assistantId]/+page.server.ts:70]: is there something like `result.message` that can be used in the line below rather than the generic "Invalid report reason"?

src/routes/settings/assistants/[assistantId]/ReportModal.svelte
Comment (julien-c) [reply src/routes/settings/assistants/[assistantId]/ReportModal.svelte:1]: honestly your modal looks cool, let's use it IMO!

Comment (mishig25) [reply src/routes/settings/assistants/[assistantId]/ReportModal.svelte:1]: especially if we are not using`textarea` & want to [keep the text short](https://github.com/huggingface/chat-ui/pull/825#discussion_r1490655236), prompt could be a good choice since we don't have to develop/maintain a component that is similar to native one

Comment (mishig25) [reply src/routes/settings/assistants/[assistantId]/ReportModal.svelte:1]: conflicting opinions. So yeah, Nathan, up to you :)

src/routes/settings/assistants/[assistantId]/ReportModal.svelte
```diff
@@ -0,0 +1,57 @@
...
+			name="reportReason"
+			class="mt-6 w-full rounded-lg border-2 border-gray-200 bg-gray-100 p-2 text-smd"
+			placeholder="This assistant should be reported because..."
```
Comment (gary149) [inline src/routes/settings/assistants/[assistantId]/ReportModal.svelte:34]:
```suggestion
			placeholder="Reason(s) for the report"
```

gary149 (approved):

src/routes/settings/assistants/[assistantId]/ReportModal.svelte
Comment (nsarrazin) [reply src/routes/settings/assistants/[assistantId]/ReportModal.svelte:31]: done!

mishig25 (approved):
lgtm !

src/routes/settings/assistants/[assistantId]/+page.server.ts
Comment (mishig25) [reply src/routes/settings/assistants/[assistantId]/+page.server.ts:70]: closing it since in the UI, we already have `maxlength="128"`

Comment (nsarrazin) [reply src/routes/settings/assistants/[assistantId]/+page.server.ts:70]: sorry I was taking a look, but we don't show error messages so I think the client side checks we have are currently enough (max length & disabled button on empty)

nsarrazin:
Alright, going to check the slack integration in prod then!

PR #828, opened 2024-02-15
✨ Add stats on conversations

coyotte508:
This PR adds an api call on `/admin/stats/compute`:

```console
curl -X POST -H "Authorization: Bearer {ADMIN_API_SECRET}" https://huggingface.co/chat/admin/stats/compute
```

This will return a 202 and launch the stats update in the background. The first run will be slow (all the stats to compute), and the next will be faster.

Once the first run successfully completes, the next runs should be much faster to execute. The last day of stats in DB will always be recomputed.

The stats are stored using this schema:

```ts
interface ConversationStats {
	createdAt: Date;
	updatedAt: Date;
	date: {
		at: Date;
		span: "day" | "weak" | "month";
		field: "updatedAt" | "createdAt";
	};
	type: "conversation" | "message";
	/**  _id => number of conversations/messages in the month */
	distinct: "sessionId" | "userId" | "userOrSessionId" | "_id";
	count: number;
}
```

There are 12 documents for each day, 12 for each week, 12 for each month:

- one on updatedAt / sessionId
- one on updatedAt / userId
- one on updatedAt / userOrSessionId
- one on updatedAt / _id (number of conversations updated that day)
- same 4 for createdAt
- same 4 for createdAt on messages

This allows to compute DAU as well as some stats on conversations

---

After merging, this will need a call on `/admin/stats/compute` to update the stats every few hours

### DIFF
--- a/.env
+++ b/.env
@@ -109,9 +109,9 @@ PUBLIC_ANNOUNCEMENT_BANNERS=`[
 
 PARQUET_EXPORT_DATASET=
 PARQUET_EXPORT_HF_TOKEN=
-PARQUET_EXPORT_SECRET=
+ADMIN_API_SECRET=# secret to admin API calls, like computing usage stats or exporting parquet data
 
-ADMIN_API_SECRET=# secret to admin API calls, like computing usage stats
+PARQUET_EXPORT_SECRET=#DEPRECATED, use ADMIN_API_SECRET instead
 
 RATE_LIMIT= # requests per minute
 MESSAGES_BEFORE_LOGIN=# how many messages a user can send in a conversation before having to login. set to 0 to force login right away
--- a/.env
+++ b/.env
@@ -111,6 +111,8 @@ PARQUET_EXPORT_DATASET=
 PARQUET_EXPORT_HF_TOKEN=
 PARQUET_EXPORT_SECRET=
 
+ADMIN_API_SECRET=# secret to admin API calls, like computing usage stats
+
 RATE_LIMIT= # requests per minute
 MESSAGES_BEFORE_LOGIN=# how many messages a user can send in a conversation before having to login. set to 0 to force login right away
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -4093,10 +4093,17 @@
 			"resolved": "https://registry.npmjs.org/int53/-/int53-0.2.4.tgz",
 			"integrity": "sha512-a5jlKftS7HUOhkUyYD7j2sJ/ZnvWiNlZS1ldR+g1ifQ+/UuZXIE+YTc/lK1qGj/GwAU5F8Z0e1eVq2t1J5Ob2g=="
 		},
-		"node_modules/ip": {
-			"version": "2.0.0",
-			"resolved": "https://registry.npmjs.org/ip/-/ip-2.0.0.tgz",
-			"integrity": "sha512-WKa+XuLG1A1R0UWhl2+1XQSi+fZWMsYKffMZTTYsiZaUD8k2yDAj5atimTUD2TZkyCkNEeYE5NhFZmupOGtjYQ=="
+		"node_modules/ip-address": {
+			"version": "9.0.5",
+			"resolved": "https://registry.npmjs.org/ip-address/-/ip-address-9.0.5.tgz",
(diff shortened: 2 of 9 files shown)


package-lock.json
coyotte508, line 1: Fix 2 high vulns running `npm audit fix`, no relation to the rest of the PR

.env
```diff
@@ -111,6 +111,8 @@ PARQUET_EXPORT_DATASET=
 PARQUET_EXPORT_HF_TOKEN=
 PARQUET_EXPORT_SECRET=
 
+ADMIN_API_SECRET=# secret to admin API calls, like computing usage stats
```
coyotte508, line 114: `PARQUET_EXPORT_SECRET` could be folded into `ADMIN_API_SECRET`

julien-c: yes agree with that suggestion^

julien-c (commented):
do you think the same system could be used for WAU/MAU? i.e. add a period to `date` or something?

coyotte508:
> do you think the same system could be used for WAU/MAU? i.e. add a period to `date` or something?

Oh no need to do anything extra for WAU/MAU, there's few enough documents

julien-c:
hmm i don't understand how you compute WAU/MAU – which data are you basing you on?

coyotte508:
yea my bad :face_with_spiral_eyes: 

will add two fields then - one timespan and one for conversation/message

nsarrazin (commented):
Left a small comment regarding the auth check, other than that I think this is a great addition! 

Would love to combine  `ADMIN_API_SECRET` and `PARQUET_EXPORT_SECRET` (maybe `ADMIN_API_SECRET ?? PARQUET_EXPORT_SECRET` for legacy configs) and then I think we can try it out!

coyotte508:
by the way @alak noticed that we have this in `handle.ts`:

```
	function errorResponse(status: number, message: string) {
		const sendJson =
			event.request.headers.get("accept")?.includes("application/json") ||
			event.request.headers.get("content-type")?.includes("application/json");
		return new Response(sendJson ? JSON.stringify({ error: message }) : message, {
			status,
			headers: {
				"content-type": sendJson ? "application/json" : "text/plain",
			},
		});
	}
```

Which returns `{error: string}` instead of `{message: string}` :facepalm: 

We should harmonize (maybe throw the api errors with `throw error` and handle them in `handleError` in `hooks.server.ts`)

nsarrazin (approved):
Looks good to me now! I'll add the `ADMIN_API_SECRET` to the prod deployment workflow in a separate PR

nsarrazin:
We were thinking about batching some stuff in cronjobs for chat-ui, for example computing trending assistants based on the number of recent conversations using them.

I looked at [croner](https://www.npmjs.com/package/croner) for this, and I feel like a scheduled job would also make sense for computing the stats, WDYT @coyotte508 ?

Doesn't have to be in this PR at all, just wanted your opinion on the topic 😄

julien-c:
@nsarrazin There's also a feature request for a simple way to define cron jobs in Spaces (directly in README.md for instance)

image

Please advocate for it in #hf-spaces-support if it'd make sense

coyotte508:
Funny I just made https://huggingface.co/spaces/huggingchat/cron-update-stats :) (private but could be public :shrug: ) 

You just need to set `ADMIN_API_SECRET` in that space for it to work.

(a scheduled github action would work too :shrug:)

---

Anyway the concern is the multiple processes running, and we only want to execute the cron job once. We could use DB locks (with a dedicated collection & unique index). But I opted for a separate space.

Up to you, feel free to pause the space and use another solution :)

nsarrazin:
I think it'd be super nice to have for sure! but because a lot of people self-host chat-ui on their machine I don't think we should rely on it for chat-ui

julien-c:
well on their machine they have cron anyways no?

nsarrazin:
My reasoning was that it would be nice if scheduled jobs could be run without any extra setup from the user side. 

We have a pretty streamlined local deployment process for chat-ui which imo we should strive to keep, and having to setup extra services/cronjobs on your machine manually creates a lot of extra friction for new users. 

This isn't a blocker for this PR tho, just something worth thinking about for the future 👀

nsarrazin:
@coyotte508 I added the secret to the prod workflow in this commit ( [ab85214](https://github.com/huggingface/chat-ui/pull/828/commits/ab85214b41217d615e80e0e409877fb8eca0039d) ), I'll add it to github secrets and I think we're good to go with this pr

nsarrazin:
@coyotte508 I opened [this issue](https://github.com/huggingface/chat-ui/issues/846) to keep track of the current conversation 😁

Issue #818, opened 2024-02-13
Settings Page Freezes


lordsoffallen:
When I go to settings to change model (after I ran a convo with a model), the UI settings page can't be closed. It freezes. Right now I have to keep reloading the page to use it

julien-c:
i think i've seen this too

nsarrazin:
Hi, is this is related to HuggingChat or a local chat-ui setup? @lordsoffallen

lordsoffallen:
It is related to HuggingChat. @nsarrazin

PR #830, opened 2024-02-16
[Assistants settings] Fix freezing

mishig25:
Fixes https://github.com/huggingface/chat-ui/issues/818

[Read this comment](https://github.com/huggingface/chat-ui/pull/830/files#r1491813449)


src/routes/settings/assistants/[assistantId]/+page.ts
```diff
@@ -1,14 +0,0 @@
-import { base } from "$app/paths";
```
Comment (mishig25) [inline src/routes/settings/assistants/[assistantId]/+page.ts:1]: I deleted `+page.ts` that uses `const data = await parent();`, which I believe was causing the freezing.

As a replacement, I've put the `load` in `+page.server.ts` that uses mongodb rather than `const data = await parent();` to get the assistant info.

Was there a specific reason to prefer using `load` in `+page.ts` over  `+page.server.ts`? (cc: @nsarrazin )

Comment (mishig25) [reply src/routes/settings/assistants/[assistantId]/+page.ts:1]: if it looks good, I can also update the model settings equavalent https://github.com/huggingface/chat-ui/blob/fix_settings_freeze/src/routes/settings/%5B...model%5D/%2Bpage.ts

Comment (nsarrazin) [reply src/routes/settings/assistants/[assistantId]/+page.ts:1]: I think the reasoning was to filter the correct assistant from parent data to reduce the number of DB calls. 

Obviously that doesn't seem to have helped with performance :laughing: But this looks good!

nsarrazin (approved):

gary149 (approved):
nice

src/routes/settings/assistants/[assistantId]/+page.ts
Comment (mishig25) [reply src/routes/settings/assistants/[assistantId]/+page.ts:1]: > if this PR looks good, I can also update the model settings equavalent https://github.com/huggingface/chat-ui/blob/fix_settings_freeze/src/routes/settings/%5B...model%5D/%2Bpage.ts

actually, this is not possible atm because models are not in db. They are rather in memory. Therefore, the current approach is sufficient

mishig25:
https://github.com/huggingface/chat-ui/pull/832 solves the issue in a different way

PR #831, opened 2024-02-16
Revert "[Assistants settings] Fix freezing"

mishig25:
Reverts huggingface/chat-ui#830 since it broke "Edit" feature. Will re-submit this PR with correct fix


mishig25:
needed to handle the dynamic fields

https://github.com/huggingface/chat-ui/blob/main/src/routes/settings/%2Blayout.server.ts#L19-L27

mishig25:
https://github.com/huggingface/chat-ui/pull/832 solves the issue in a different way

PR #832, opened 2024-02-16
[Mongo] Optimize `reports` collection query

mishig25:
As the title says, optimize `reports` collection query

### DIFF
--- a/src/lib/server/database.ts
+++ b/src/lib/server/database.ts
@@ -85,5 +85,5 @@ client.on("open", () => {
 	assistants.createIndex({ userCount: 1 }).catch(console.error);
 	assistants.createIndex({ featured: 1, userCount: -1 }).catch(console.error);
 	assistants.createIndex({ modelId: 1, userCount: -1 }).catch(console.error);
-	reports.createIndex({ assistantId: 1, createdBy: 1 }).catch(console.error);
+	reports.createIndex({ createdBy: 1, assistantId: 1 }).catch(console.error);
 });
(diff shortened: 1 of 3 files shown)


src/routes/settings/+layout.server.ts
```diff
@@ -21,10 +21,10 @@ export const load = (async ({ locals, parent }) => {
 				createdByMe:
 					el.createdById.toString() === (locals.user?._id ?? locals.sessionId).toString(),
 				reported:
-					(await collections.reports.countDocuments({
+					(await collections.reports.findOne({
```
mishig25, line 24: this should be faster than aggregation of `countDocuments`

src/lib/server/database.ts
```diff
@@ -85,5 +85,5 @@ client.on("open", () => {
 	assistants.createIndex({ userCount: 1 }).catch(console.error);
 	assistants.createIndex({ featured: 1, userCount: -1 }).catch(console.error);
 	assistants.createIndex({ modelId: 1, userCount: -1 }).catch(console.error);
-	reports.createIndex({ assistantId: 1 }).catch(console.error);
+	reports.createIndex({ createdBy: 1, assistantId: 1 }).catch(console.error);
```
coyotte508, line 88: I think it's nice to keep both indexes - that way we can easily see how many times a particular assistant has been reported, or which assistant is the most reported

mishig25: wdyt 2750cbcb322e04cd1d85f5fc969c6af486e7e976 ?

coyotte508: nice!

PR #841, opened 2024-02-19
[Assistants] Filter on names

mishig25:
### TLDR: Filter on assistant's names

https://github.com/huggingface/chat-ui/assets/11827707/bca0b256-d6e0-4157-82aa-88c676334156

Tried to make a bit of a nicer UX on mobile using tailwind transitions

https://github.com/huggingface/chat-ui/assets/11827707/93fb3a62-0bc0-4ca5-9d8a-6c60134acd3d


src/routes/assistants/+page.server.ts
```diff
@@ -34,6 +35,7 @@ export const load = async ({ url, locals }) => {
 		...(modelId && { modelId }),
 		...(!createdByCurrentUser && { userCount: { $gt: 1 } }),
 		...(user ? { createdById: user._id } : { featured: true }),
+		...(query && { name: { $regex: query, $options: "i" } }),
```
mishig25, line 38: For this filter change, I've created this index https://github.com/huggingface/chat-ui/blob/assistants_search/src/lib/server/database.ts#L88-L89

Please let me know if it is enough (cc: @coyotte508 )

src/routes/assistants/+page.svelte
```diff
@@ -14,22 +14,36 @@
...
 			newKeys: { modelId: (e.target as HTMLSelectElement).value },
 			existingKeys: { behaviour: "delete_except", keys: ["user"] },
 		});
 		goto(newUrl);
 	};
+
+	const filterOnName = debounce(async (e: Event) => {
+		const value = (e.target as HTMLInputElement).value;
+		const newUrl = getHref($page.url, { newKeys: { q: value } });
+		await goto(newUrl);
```
mishig25, line 42: taking advantage of svelte partial hydration without creating explciit API routes for now

src/routes/assistants/+page.server.ts
coyotte508: If you want to efficiently search, besides the `$text` index (more suited to full text), if you want to do it in mongo you need prefixed case-sensitive regexes, and store search tokens.

Eg we have this in moon-landing:

```
generateSearchTokens(str: string): string[] {
		const lcase = str.toLowerCase().trim();

		return [
			...new Set(
				[
					lcase.replace(/[-._\s/]/g, ""),
					...lcase.split("/").map(x => x.replace(/[-._\s]/g, "")),
					...lcase.replace(/[-.\s_/]/g, " ").split(" "),
					...lcase
						.replace(/[-\s_/]/g, " ")
						.split(" ")
						.map(x => x.replace(/\./g, "")),
					/// ^ do not split on `.` but still remove them from search tokens (simpler for search)
					/// allows for searches like "2.7B" to return things like "EleutherAI/gpt-neo-2.7B"
				].filter(x => !!x)
			),
		];
	}
```

You can apply this to the name - probably tweaking it a bit since the strings we process are a bit different - and store the search tokens.

Then create an index on `searchTokens: 1, userCount: -1`

Then the search, you do `{ searchTokens: { $all: q.split(/[-._/\s]+/).map(makeRegex) } }` with

```
				const makeRegex = (s: string) => new RegExp("^" + Utils.escapeForRegExp(s));
```

(which prefixes the regex by using `^`)

At least that's how we do it for the hub. Alterative to this and `$text` is using meilisearch :sweat_smile:

mishig25: Right now, we plan to only filter on `assistant.name`. Would this filter `(query && { name: { $regex: query, $options: "i" } }),` too slow considering that we have 10k+ assistants atm ?

We can definitely create/use moon-landing-like `searchTokens` if we also want to include `assistant.description` besides `assistant.name`

coyotte508: `searchTokens` are suitable for assistant.name. If we included description we'd have to move to a `$text` index instead.

We use `searchTokens` on the hub to search org names, model names, user names etc

> Would this filter (query && { name: { $regex: query, $options: "i" } }), be too slow considering that we have 10k+ assistants atm ?

As long as the follow up PR is not forgotten - but may as well do it right the first time

mishig25: @gary149 wdyt ?

mishig25: @coyotte508 the question is:
1. Can we start with this simple `name: { $regex: query, $options: "i" }` only on `assistant.name` without having to create `searchTokens`? By can we, I mean: would the performance be good enough for the number of assistants we have? Or we don't know how the experience is gonna be without trying it empirically?
2. However, if we have to use `searchTokens`, then it would make sense to add `assistant.description` to `searchTokens` as well as `assistant.name`

coyotte508: IMO:

- Search name only: use `searchTokens`
- Search name & description: use a [text index](https://www.mongodb.com/docs/manual/core/indexes/index-types/index-text/create-text-index/#std-label-compound-text-index-example)

gary149: Ok let's use `searchTokens` then.

mishig25: ready for review

src/routes/assistants/+page.server.ts
```diff
@@ -16,6 +16,7 @@ export const load = async ({ url, locals }) => {
 	const modelId = url.searchParams.get("modelId");
 	const pageIndex = parseInt(url.searchParams.get("p") ?? "0");
 	const username = url.searchParams.get("user");
+	const query = url.searchParams.get("q");
```
coyotte508, line 19:
```suggestion
	const query = url.searchParams.get("q")?.trim();
```

coyotte508 (approved):

src/routes/assistants/+page.svelte
```diff
@@ -129,16 +148,32 @@
...
 					</a>
 				{/if}
 			{/if}
+			<div
+				class="ml-auto mr-2 flex h-9 w-9 items-center gap-x-2 rounded-full border border-gray-300 px-2 transition-all duration-300 ease-out hover:w-64 md:w-64 dark:border-gray-500"
+			>
+				<button class="flex-shrink-0 cursor-default" on:click={() => filterInputEl.focus()}
+					><CarbonSearch class="text-gray-400" /></button
+				>
+				<input
```
gary149, line 167: I get weird behaviours:

https://github.com/huggingface/chat-ui/assets/3841370/a1f51c27-96b3-4c08-b18d-d25fe71b2792

mishig25: when you load the page, you see `undefined` as the value in the box ?

mishig25: https://github.com/huggingface/chat-ui/pull/841/commits/c0387ce8eb8bcb8da9a002a5d5d3531a6e67154f trimming the value was turning `null` into `undefined`

shoud be fixed by here https://github.com/huggingface/chat-ui/pull/841/commits/d65f7c17ecb4e904e6dfff0e6e83d08ab079ff16

nsarrazin (approved):
To me this looks good for now after testing locally! I'd also like to be able to search on substrings and not necessarily whole words but it can come later

mishig25:
superceded by https://github.com/huggingface/chat-ui/pull/873

after talking with @gary149 , we decided to use `searchTokens` solution rather than [mongo's text index](https://www.mongodb.com/docs/manual/core/indexes/index-types/index-text/) because "mongo's text index" has a limitation of only matching on whole words, not letters or prefixes

julien-c:
yes good call imo

mishig25:
As suggested [here](https://github.com/huggingface/chat-ui/pull/873#pullrequestreview-1908968954), merged https://github.com/huggingface/chat-ui/pull/873 into this branch & will continue from this branch/PR

mishig25:
Order of steps?
1. Merge [#841](https://github.com/huggingface/chat-ui/pull/841) but hide it to users
2. Merge [#897](https://github.com/huggingface/chat-ui/pull/897) and run the migration (before we merge, 897 should use the correct/exact [searchTokens creation](https://github.com/huggingface/chat-ui/blob/5e38dffcf3298148b4e45a4ecf8830a728c4f0c7/src/lib/migrations/routines/01-update-search-assistants.ts#L23))
3. Create another PR that makes [#841](https://github.com/huggingface/chat-ui/pull/841) to users

PR #850, opened 2024-02-20
Count system prompt tokens

mishig25:
On updating model settings:

Screenshot 2024-03-18 at 3 16 28 PM

<br><br>

On updating assitants settings:

image

### DIFF
--- a/src/lib/components/AssistantSettings.svelte
+++ b/src/lib/components/AssistantSettings.svelte
@@ -12,6 +12,7 @@
 
 	import { useSettingsStore } from "$lib/stores/settings";
 	import { isHuggingChat } from "$lib/utils/isHuggingChat";
+	import TokensCounter from "./TokensCounter.svelte";
 
 	type ActionData = {
 		error: boolean;
@@ -28,6 +29,10 @@
 	export let models: Model[] = [];
 
 	let files: FileList | null = null;
+	let selectedModel: Model | undefined = models.find(
+		(model) => assistant?.modelId && assistant.modelId === model.id
+	);
+	let enteredSystemPrompt = "";
 
 	const settings = useSettingsStore();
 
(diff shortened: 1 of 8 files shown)


gary149:
There're some errors with some models I think 👀

mishig25:
> There're some errors with some models I think 👀

The reason being is that: https://huggingface.co/meta-llama/Llama-2-70b-chat-hf is a gated model. Therefore, the tokenizer code from frontend tries to grab the files but it is gated.

For now, the behaviour is that: if the model is not gated, the token count is shown. If it is gated, nothing appears.

Wwe have two poaaiblities:
1. Keep the current behaviour where no tokens are shown on gated models
2. Move the tokens counting code to backend/server so that tokens can be shown on all models.

wdyt ?

mishig25:
> I think the current behavior is good. But it shouldn't even try to tokenize if the tokenizer is not available (here it throws 100s of console errors).

handled it. Only the one error gets thrown in this case

gary149:
Just don't do it on llama-2? (as its gated it will never work)

nsarrazin (commented):
When it works it seems to work well, but I sometimes get console errors when switching between different models in the settings without reloading the page? But if I reload the page the tokenization works again. Not sure what is causing it

gary149:
> When it works it seems to work well, but I sometimes get console errors when switching between different models in the settings without reloading the page? But if I reload the page the tokenization works again. Not sure what is causing it

Yes trying to reproduce that too 👀

src/lib/components/TokensCounter.svelte
gary149, line 1: This won't work on local models with chat-ui. I would probably set an explicit key in the config for models that are compatible.

src/lib/components/AssistantSettings.svelte
```diff
@@ -390,12 +397,25 @@
...
+					bind:value={systemPrompt}
+				/>
+				{#if modelId}
+					{@const model = models.find((_model) => _model.id === modelId)}
+					{#if model?.tokenizer && systemPrompt}
+						<TokensCounter
+							classNames="absolute bottom-2 right-2"
+							prompt={systemPrompt}
+							modelTokenizer={model.tokenizer}
+							max_new_tokens={model?.parameters?.max_new_tokens}
```
gary149, line 414: use `truncate` value instead?

mishig25: handled in f47c6bba93b8e8c8416a586138d37921631cb222

nsarrazin (approved):
Works quite well! 🔥

gary149 (approved):

PR #858, opened 2024-02-21
Direct model page

nsarrazin:
The idea is to have a url that takes you directly to a model

`http://localhost:5173/chat/models/mistralai/Mixtral-8x7B-Instruct-v0.1`

We can generate custom social thumbnails for each model so we can have better comms built-in.

image
image
image

WIP, just putting out there for now


nsarrazin:
cc @gary149 if you want to have a look

gary149:
Or maybe just render the thumbnail dynamically using the existing route: http://localhost:5173/chat?model=meta-llama/Llama-2-70b-chat-hf

julien-c:
IMO it'd be better to have dedicated routes no? (we can still backward-support existing URLs by redirecting them?)

julien-c:
(nice thumbnails!)

gary149:
Google seems to be considering it as a dedicated route already? I won't change it

image

nsarrazin:
> Google seems to be considering it as a dedicated route already? I won't change it
> image

Checked the google link, it goes to `https://huggingface.co/chat/settings/mistralai/Mixtral-8x7B-Instruct-v0.1` so seems like a direct route would be worth it for SEO?

gary149:
> Checked the google link, it goes to https://huggingface.co/chat/settings/mistralai/Mixtral-8x7B-Instruct-v0.1 so seems like a direct route would be worth it for SEO?

So you mean removing `/settings`? dont see how this will have a big impact on SEO.

nsarrazin:
No I meant that google seems to have indexed the settings page, so it could also index a direct route `/model/<model_name>`. Not saying we remove the settings page 😁

gary149:
> No I meant that google seems to have indexed the settings page, so it could also index a direct route /model/<model_name>. Not saying we remove the settings page 😁

Yes because for me the settings page was the direct page, but yes maybe you are right the UX is a bit weird.

gary149 (approved):
Ok (let's not forget to check if double og:image works)

src/routes/models/[...model]/thumbnail.png/ModelThumbnail.svelte
```diff
@@ -19,7 +19,7 @@
 	</div>
 
 	<div
-		class="flex h-[200px] w-full flex-col items-center justify-center rounded-b-none bg-yellow-500/10 pb-10 pt-10 text-4xl text-gray-500"
+		class="flex h-[200px] w-full flex-col items-center justify-center rounded-b-none bg-{PUBLIC_APP_COLOR}-500/10 pb-10 pt-10 text-4xl text-gray-500"
```
Comment (gary149) [inline src/routes/models/[...model]/thumbnail.png/ModelThumbnail.svelte:22]: Oh yes I removed it for simplicity :) also because it means you need to explicitly set it somewhere or it will not be catched by tailwindcss JIT engine).

PR #863, opened 2024-02-22
[Mongo] count messages (aggregate) only when needed

mishig25:
Perform `count messages` only when necessary

### DIFF
--- a/src/routes/+layout.server.ts
+++ b/src/routes/+layout.server.ts
@@ -91,8 +91,9 @@ export const load: LayoutServerLoad = async ({ locals, depends }) => {
 			(
 				await collections.conversations
 					.aggregate([
-						{ $match: authCondition(locals) },
+						{ $match: authCondition(locals), "messages.from": "assistant" },
 						{ $project: { messages: 1 } },
+						{ $limit:  messagesBeforeLogin + 1 },
 						{ $unwind: "$messages" },
 						{ $match: { "messages.from": "assistant" } },
 						{ $count: "messages" },
--- a/src/routes/+layout.server.ts
+++ b/src/routes/+layout.server.ts
@@ -46,25 +46,27 @@ export const load: LayoutServerLoad = async ({ locals, depends }) => {
 		});
 	}
 
-	// get the number of messages where `from === "assistant"` across all conversations.
-	const totalMessages =
-		(
-			await collections.conversations
-				.aggregate([
-					{ $match: authCondition(locals) },
-					{ $project: { messages: 1 } },
-					{ $unwind: "$messages" },
-					{ $match: { "messages.from": "assistant" } },
-					{ $count: "messages" },
-				])
-				.toArray()
-		)[0]?.messages ?? 0;
-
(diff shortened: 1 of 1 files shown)


src/routes/+layout.server.ts
```diff
@@ -46,25 +46,27 @@ export const load: LayoutServerLoad = async ({ locals, depends }) => {
...
-
-	const loginRequired = requiresUser && !locals.user && userHasExceededMessages;
+	let loginRequired = false;
+
+	if (requiresUser && !locals.user && messagesBeforeLogin) {
+		// get the number of messages where `from === "assistant"` across all conversations.
+		const totalMessages =
+			(
+				await collections.conversations
+					.aggregate([
```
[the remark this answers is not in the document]
mishig25: overall, `aggregate` is an expensive operation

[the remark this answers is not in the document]
nsarrazin: Nice, hopefully this will help quite a bit! Can we also reduce the number of times where this is needed even more by first checking the conversations count ?

For example if the user has more than `messagesBeforeLogin` conversations, then there shouldn't be a need to aggregate at all, wdyt? Not entirely sure about it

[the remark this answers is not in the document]
mishig25: handled in https://github.com/huggingface/chat-ui/pull/863/commits/fe99cf2dc9d89c6ba91c8fc096a273cc317de23e

nsarrazin (approved):
:fire: thanks for tackling it!

src/routes/+layout.server.ts
```diff
@@ -101,6 +81,28 @@ export const load: LayoutServerLoad = async ({ locals, depends }) => {
...
+	let loginRequired = conversations.length > messagesBeforeLogin;
+
+	if (!loginRequired && messagesBeforeLogin && requiresUser && !locals.user) {
+		// get the number of messages where `from === "assistant"` across all conversations.
+		const totalMessages =
+			(
+				await collections.conversations
+					.aggregate([
+						{ $match: authCondition(locals) },
+						{ $project: { messages: 1 } },
```
coyotte508, line 95:
```suggestion
						{ $match: authCondition(locals), "messages.from": "assistant" },
						{ $project: { messages: 1 } },
						{ $limit:  messagesBeforeLogin + 1 },
```

coyotte508: This way, if there are 100k convos and `messagesBeforeLogin` is 10, we only fetch 11 convos (in which an assistant replied) from DB

Note that this shouldn't happen (because the limit would've already been triggered long before the 100k convos), but this would have improved the situation before this PR. Anyway, just in case :)

coyotte508 (approved):

src/routes/+layout.server.ts
mishig25: however, would `{ $match: authCondition(locals), "messages.from": "assistant" }` work because it is happening before `$unwind: "$messages"` ?

coyotte508: yes it works

The only thing, it coudl match one convo with 100 assistant messages, that's why we need the following `$unwind`

but it works, you can query array elements like this in Mongo

nsarrazin:
shall we merge it?

PR #864, opened 2024-02-22
/models tweaks

gary149:
mainly change the behaviour of /models, clicking goes to a new conversation instead of settings.


nsarrazin (approved):
Works well, I like the added logo!

PR #865, opened 2024-02-22
Revert "[Mongo] count messages (aggregate) only when needed"

mishig25:
Reverts huggingface/chat-ui#863


mishig25:
checking 👁️

mishig25:
handled in [#866](https://github.com/huggingface/chat-ui/pull/866)

PR #867, opened 2024-02-22
Bug Fix: Json Decoder aggessively pulls json

MDCurrent:
**Background**
Had this problem on 2 different setups, essentially the decoder for llamacpp would grab partially streamed json and decode it without the new line 

**Description**

Just wait until we see a new line character before trying to decode the json. if we fail to decode a token, just toss it. *may lead to tokens being thrown out but after a couple of days of running this thru the ringer it seemed to work, and dropping a word is a better UX than just crashing when it fails to decode the response


MDCurrent:
Screenshot from 2024-02-22 09-40-58
this is the type of output id get when trying to hit my locall llamacpp server

nsarrazin (approved):
Seems to work well, will merge!

PR #869, opened 2024-02-22
[Assistants] Fix `remove` behaviour

mishig25:
Follow up to https://github.com/huggingface/chat-ui/pull/859, specifically to [this comment](https://github.com/huggingface/chat-ui/pull/859#discussion_r1497978234)

Bug: when a user removes an assistant, the assistant was NOT being removed from the settings list.
[This PR fixes this problem](https://github.com/huggingface/chat-ui/pull/869#discussion_r1499470412)

### DIFF
--- a/src/routes/+layout.server.ts
+++ b/src/routes/+layout.server.ts
@@ -76,8 +76,11 @@ export const load: LayoutServerLoad = async ({ locals, depends }) => {
 		.limit(300)
 		.toArray();
 
+	const userAssistants = settings?.assistants?.map((assistantId) => assistantId.toString()) ?? [];
+	const userAssistantsSet = new Set(userAssistants);
+
 	const assistantIds = [
-		...(settings?.assistants?.map((assistantId) => assistantId) ?? []),
+		...userAssistants.map((el) => new ObjectId(el)),
 		...(conversations.map((conv) => conv.assistantId).filter((el) => !!el) as ObjectId[]),
 	];
 
@@ -147,7 +150,7 @@ export const load: LayoutServerLoad = async ({ locals, depends }) => {
 				settings?.shareConversationsWithModelAuthors ??
 				DEFAULT_SETTINGS.shareConversationsWithModelAuthors,
 			customPrompts: settings?.customPrompts ?? {},
-			assistants: settings?.assistants?.map((el) => el.toString()) ?? [],
+			assistants: userAssistants,
 		},
 		models: models.map((model) => ({
 			id: model.id,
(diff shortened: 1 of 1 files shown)


src/routes/+layout.server.ts
```diff
@@ -166,12 +169,15 @@ export const load: LayoutServerLoad = async ({ locals, depends }) => {
...
 		})),
 		oldModels,
-		assistants: assistants.map((el) => ({
-			...el,
-			_id: el._id.toString(),
-			createdById: undefined,
-			createdByMe: el.createdById.toString() === (locals.user?._id ?? locals.sessionId).toString(),
-		})),
+		assistants: assistants
+			.filter((el) => userAssistantsSet.has(el._id.toString()))
```
mishig25, line 173: this `filter` is the key line

coyotte508 (approved):

src/routes/+layout.server.ts
```diff
@@ -76,8 +76,11 @@ export const load: LayoutServerLoad = async ({ locals, depends }) => {
 		.limit(300)
 		.toArray();
 
+	const userAssistants = settings?.assistants?.map((assistantId) => assistantId.toString()) ?? [];
+	const userAssistantsSet = new Set(userAssistants);
```
coyotte508, line 80: We have `EqualitySet` that we could bring to this codebase:

```ts

/**
 * Subclass of Set but works with non-primitives
 * using a bijective object to id mapping.
 * (passed to the constructor)
 *
 * By default the mapping is calling the String function on the
 * object.
 */
export class EqualitySet<K = ObjectId> extends Set<unknown> {
	private oToId: (o: K) => string;

	constructor(oToId?: (o: K) => string);
	constructor(iterable: Iterable<K>, oToId?: (o: K) => string);
	constructor(iterableOrFn?: Iterable<K> | ((o: K) => string), oToId?: (o: K) => string) {
		super();

		const iterable = oToId
			? (iterableOrFn as Iterable<K>)
			: iterableOrFn
				? Symbol.iterator in iterableOrFn
					? (iterableOrFn as Iterable<K>)
					: undefined
				: undefined;

		this.oToId = (iterable ? oToId : (iterableOrFn as (o: K) => string)) || String;

		/// do not pass iterable to super()
		/// because this.oToId is not yet defined.
		if (iterable) {
			for (const k of iterable) {
				this.add(k);
			}
		}
	}

	override delete(key: K): boolean {
		return super.delete(this.oToId(key));
	}

	override has(key: K): boolean {
		return super.has(this.oToId(key));
	}

	override add(key: K): this {
		return super.add(this.oToId(key));
	}

	*toIterable<Ret = K>(idToObject: (str: string) => Ret): Iterable<Ret> {
		for (const val of this) {
			yield idToObject(val as string);
		}
	}
}

```

mishig25: the process of chat-ui becoming moon-landing continues :)

PR #871, opened 2024-02-22
[Assistants settings] Update data backend

mishig25:


src/routes/settings/assistants/[assistantId]/+page.server.ts
```diff
@@ -21,6 +22,30 @@ async function assistantOnlyIfAuthor(locals: App.Locals, assistantId?: string) {
 	return assistant;
 }
 
+export async function load({ parent, params, locals }) {
+	const data = await parent();
```
Comment (coyotte508) [inline src/routes/settings/assistants/[assistantId]/+page.server.ts:26]: Note that this triggers the root's `+layout.server.ts`'s `load`.

So each time we switch assistant, this will trigger a load of _all assistants_

IMO we can get the redirect in `page.ts` - that way the load isn't triggered - since the frontend will be able to execute it. 

---

TLDR:

- Keep `page.ts` as is
- remove the redirect / `parent()` call here
- The rest of the PR is great!

Comment (mishig25) [reply src/routes/settings/assistants/[assistantId]/+page.server.ts]: handled in 1d5994fdca8a9b9ab9aae15cca2e5e1da884c84b

nsarrazin:
How does it work when you have both a server load function and a universal load function for the same `+page.svelte` ? Seems like the type checker is not super happy about it haha but couldn't find more info about it

coyotte508:
> How does it work when you have both a server load function and a universal load function for the same `+page.svelte` ? Seems like the type checker is not super happy about it haha but couldn't find more info about it

```ts
export async function load({ parent, data }) {
	const res = await parent();

	// checks

	return data;
}
```

Something like that

PR #873, opened 2024-02-23
assistants search use searchTokens

mishig25:


src/lib/server/database.ts
```diff
@@ -54,7 +55,33 @@ export const collections = {
 	bucket,
 };
 
+/**
```
mishig25, line 58: Below is my proposed way to run database migrations. So far, there was no need for migrations. Since this PR introduces new field `assistant.searchTokens`, we need all existing assistants to have this new field (i.e. run a migration script)

mishig25:
adding @Pierrci as reviewer just in case

src/lib/server/database.ts
nsarrazin: But how will this work for people running chat-ui locally? Ideally they shouldn't be running migrations script manually to get the latest changes? 

I guess in that case it's fairly minor (you just loose search on assistants) but is there a better way to run these automatically if needed?

mishig25: > But how will this work for people running chat-ui locally?

they would need to run the script manually once by declaring env vars `migrations=refreshSearchTokens` before `npm run [xyz]`.

> is there a better way to run these automatically if needed?

I might be wrong but it might be bit complicated to set up an automaitc mechanism. You'd need some kind of trigger to run it automatically, in which case it is as bad as running the script manually. OR we can setup some mechanism based on git (check if `database-migrations.ts` file was changed & then run the changed migration function automoatically; however, it seems overkill to me)

nsarrazin: I took a look at how [jellyfin](https://github.com/jellyfin/jellyfin) handles migration (a big open source project that users self-host and that needs to trigger migrations on user setups from time to time). Looks like this is how they handle migrations.

1. Each migration is represented by some kind of object that contains a GUID and a function that performs the migration. (see [example](https://github.com/jellyfin/jellyfin/blob/master/Jellyfin.Server/Migrations/Routines/FixPlaylistOwner.cs))
2. A list of migrations is stored in the user DB (list of GUIDs), starts empty.
4. A routine starts when the server starts, checks if there are any migrations that have not run yet (by comparing the GUIDs stored in the DBs vs the list of migrations in the code) [(see code)](https://github.com/jellyfin/jellyfin/blob/0bc41c015f4ec907de75fe215589b7e30a819b54/Jellyfin.Server/Migrations/MigrationRunner.cs#L108-L143)
6. The routine runs migrations that have not yet run, and adds their GUID in the DB when it's done, so they can be skipped the next time.

I know this is more complex and definitely unrelated to the search feature, but I think having an automated migrations setup would be very useful and I don't think we can ask users to run migrations scripts manually, this is different from moon-landing in that sense 😅 

I can do the work on this if we decide to go for it, don't want to add more to your plate for something unrelated to the PR

nsarrazin: Maybe we can finish reviewing this PR as is, merging it with the custom migrations script so it's not blocking and I can implement the automated migrations later if we want to.

Is there an easy way to fall back to the old search if the migration hasn't run yet? (as a temporary fix for chat-ui users)

mishig25: yep, followup PR specifically for migrations sounds great

Pierrci (commented):
LGTM from a quick glance :)

src/lib/server/database.ts
coyotte508: That would probably involve DB locks as well (to prevent running the same migration twice simultaneously)

nsarrazin (approved):
Tested locally, works great! What's the plan for running the migration in prod when we deploy?

mishig25:
> What's the plan for running the migration in prod when we deploy?

we have to add the flag 
```
migrations=refreshSearchTokens npm run start # or however the prod version is starting
```

nsarrazin:
That would be [here](https://github.com/huggingface/chat-ui/blob/main/Dockerfile#L32) I think! Not sure if there are implications due the fact that there's multiple replicas of the space running at once

Pierrci:
Mmm yes, I'm not an infra specialist, but in the current state of the PR, I think it means the migration would run on each of the replicas. From what I can tell, the migration is idempotent, so it shouldn't really be a problem other than in terms of DB resource potentially (though highly unlikely). A quick optimization would be to filter out documents that already have `searchTokens` at https://github.com/huggingface/chat-ui/pull/873/files#diff-0ebf5af0c102c78cdfffed0780bccd4fe5707ac0e33708e10e4eeafdbf52e622R10, but it wouldn't prevent redundant writes completely if the 3 replicas are started at the same time (not sure about that).

Another option I see to prevent 3 DB migrations (other than creating a DB collection for migrations to be used both as a lock and history record) is to reduce the number of replicas to 1 temporarily; not sure if it's possible, I guess it depends on the usage.

On another topic, the fact that the migration flag has to be set in the dockerfile is not optimal either; maybe it's possible to use an env var so that it can be set externally and not completely hardcoded? (but I guess it's ok if it's really a one shot and that another commit comes quickly after to remove this flag and before any restart following the migration)

coyotte508:
given this migration is not problematic we can just let it be, and see about a better mechanism for the future?

nsarrazin:
Yeah agreed, don't want to be blocking if the migration won't cause any problems running multiple times. 
 
We can modify the dockerfile in this PR to add the env variable, deploy it to prod, and then merge another PR afterwards to remove the env variable from the dockerfile.

wdyt @mishig25 ?

mishig25:
@gary149 ready to go ?

gary149 (approved):
yes you can merge into assistants_search, I'll tweak the UI there then.

PR #886, opened 2024-02-29
Add limits on API endpoints

nsarrazin:
This PR adds a new env var `RATE_LIMITS` which can contain limits for conversations, messages depth, assistants and messages per minute.

For HuggingChat the defauls are set as follow:

```
RATE_LIMITS=`{
  "conversations" : 3000, // max number of conversations per user
  "messages" : 500, // max messages per conversation, both user and assistants are counted
  "assistants" : 100, // max number of assistants per user
  "messagesPerMinute" : 16, // max messages per minute per user
}`
```

though we could tweak them in the future.

### DIFF
--- a/src/lib/server/rateLimits.ts
+++ b/src/lib/server/rateLimits.ts
@@ -1,0 +1,22 @@
+import { z } from "zod";
+import { RATE_LIMITS, RATE_LIMIT } from "$env/static/private";
+import JSON5 from "json5";
+
+// RATE_LIMIT is the legacy way to define messages per minute limit
+export const rateLimitsSchema = z
+	.object({
+		conversations: z.coerce.number().optional(), // how many conversations
+		messages: z.coerce.number().optional(), // how many messages in a conversation
+		assistants: z.coerce.number().optional(), // how many assistants
+		messagesPerMinute: z
+			.preprocess((val) => {
+				if (val === undefined) {
+					return RATE_LIMIT;
+				}
+				return val;
+			}, z.coerce.number().optional())
+			.optional(), // how many messages per minute
+	})
+	.optional();
+
+export const RateLimits = rateLimitsSchema.parse(JSON5.parse(RATE_LIMITS));
--- a/src/routes/conversation/[id]/+server.ts
+++ b/src/routes/conversation/[id]/+server.ts
@@ -1,4 +1,4 @@
-import { MESSAGES_BEFORE_LOGIN, RATE_LIMIT } from "$env/static/private";
+import { MESSAGES_BEFORE_LOGIN } from "$env/static/private";
 import { authCondition, requiresUser } from "$lib/server/auth";
 import { collections } from "$lib/server/database";
 import { models } from "$lib/server/models";
@@ -19,6 +19,7 @@ import { buildSubtree } from "$lib/utils/tree/buildSubtree.js";
 import { addChildren } from "$lib/utils/tree/addChildren.js";
 import { addSibling } from "$lib/utils/tree/addSibling.js";
 import { preprocessMessages } from "$lib/server/preprocessMessages.js";
+import { RateLimits } from "$lib/server/rateLimits.js";
 
 export async function POST({ request, locals, params, getClientAddress }) {
 	const id = z.string().parse(params.id);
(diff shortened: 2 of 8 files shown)


julien-c:
(in theory the numbers we use shouldn't be published, imo)

nsarrazin:
@julien-c @gary149 moved the `RATE_LIMITS` var to our secrets, see [here](https://huggingface.slack.com/archives/C052EKHD1U0/p1709299176259599?thread_ts=1709221246.047669&cid=C052EKHD1U0)

src/lib/server/rateLimits.ts
```diff
@@ -0,0 +1,23 @@
...
+
+// RATE_LIMIT is the legacy way to define messages per minute limit
+export const rateLimitsSchema = z
+	.object({
+		conversations: z.coerce.number().optional(), // how many conversations
+		messages: z.coerce.number().optional(), // how many messages in a conversation
+		assistants: z.coerce.number().optional(), // how many assistants
+		messageLength: z.coerce.number().optional(), // how long can a message be before we cut it off
+		messagesPerMinute: z
+			.preprocess((val) => {
```
coyotte508, line 13: nice, didn't know about `preprocess`

src/lib/server/rateLimits.ts
```diff
@@ -0,0 +1,23 @@
+import { z } from "zod";
+import { RATE_LIMITS, RATE_LIMIT } from "$env/static/private";
+import JSON5 from "json5";
+
+// RATE_LIMIT is the legacy way to define messages per minute limit
+export const rateLimitsSchema = z
```
coyotte508, line 6: It's not `rateLimits` except for the `messagesPerMinute`

It's more like `USAGE_LIMITS`

src/routes/conversation/[id]/+server.ts
```diff
@@ -101,10 +102,17 @@ export async function POST({ request, locals, params, getClientAddress }) {
 		await collections.messageEvents.countDocuments({ ip: getClientAddress() })
 	);
 
-	if (RATE_LIMIT != "" && nEvents > parseInt(RATE_LIMIT)) {
+	if (RateLimits?.messagesPerMinute && nEvents > RateLimits.messagesPerMinute) {
```
Comment (coyotte508) [inline src/routes/conversation/[id]/+server.ts:105]: Should we only fetch `nEvents` if `RateLimits?.messagePerMinute` is defined?

src/lib/server/rateLimits.ts
```diff
@@ -0,0 +1,23 @@
...
+				if (val === undefined) {
+					return RATE_LIMIT;
+				}
+				return val;
+			}, z.coerce.number().optional())
+			.optional(), // how many messages per minute
+	})
+	.optional();
+
+export const RateLimits = rateLimitsSchema.parse(JSON5.parse(RATE_LIMITS));
```
coyotte508, line 23: Not sure about casing, should it be `kebabCase` instead?

`RateLimits` has the casing of a class / type / enum

coyotte508 (commented):
mostly style comments, code ok otherwise

src/routes/conversation/[id]/+server.ts
Comment (nsarrazin) [reply src/routes/conversation/[id]/+server.ts]: yes indeed, fixed

src/lib/server/rateLimits.ts
nsarrazin: fixed as well!

src/routes/conversation/[id]/+server.ts
```diff
@@ -133,6 +142,9 @@ export async function POST({ request, locals, params, getClientAddress }) {
 		})
 		.parse(json);
 
+	if (usageLimits?.messageLength && (newPrompt?.length ?? 0) > usageLimits.messageLength) {
+		throw error(400, "Message too long.");
+	}
```
Comment (coyotte508) [inline src/routes/conversation/[id]/+server.ts:147]: We should probably prevent in the front-end a too-long message

Also make sure to convert \r\n to \n (if it's a native form submission) cc @krampstudio

coyotte508 (approved):

src/routes/conversation/[id]/+server.ts
Comment (coyotte508) [reply src/routes/conversation/[id]/+server.ts:153]: rate limits make sense to keep secret, but message length... :thinking:

PR #887, opened 2024-02-29
Replace  `.env.template` by `.env.huggingchat`

nsarrazin:
Just to make the name of the `.env` more explicit


mishig25 (approved):
lgtm! Does it mean that after this PR gets merged, people need to `rn .env.template .env.huggingchat` ?

PR #897, opened 2024-03-05
Automated migration setup

nsarrazin:
This PR adds a check on startup that runs migrations if needed.

## How it works

Each migration is defined by it's. `_id`  field which is unique and hardcoded. When the migration is applied we add it to the `"migrationResults"` collections, so we can check if it has already been run. Each migration is wrapped in a transaction to ensure consistency.

The migration check runs at the top-level of `hooks.server.ts` to ensure it only runs once on startup.

### DB Locks

Because we can have multiple instances of chat-ui running against a single DB we need to ensure we have a lock before running migrations. 

The code for lock related operations is under [lock.ts](https://github.com/huggingface/chat-ui/blob/f7fde32646f89ea31bcf461147e9f4da08f73c02/src/lib/migrations/lock.ts#L19) with the [associated unit-tests](https://github.com/huggingface/chat-ui/blob/20441bc103398d79bb6d916fd7e714d0e0d166ff/src/lib/migrations/migrations.spec.ts).

How it works: 
1. We have a `semaphore` collection with a [unique index on keys](https://github.com/huggingface/chat-ui/blob/f7fde32646f89ea31bcf461147e9f4da08f73c02/src/lib/server/database.ts#L152-L153) 
2. We can set the lock status by using `insertOne`, if it fails to insert then a lock must already exist (see [acquireLock](https://github.com/huggingface/chat-ui/blob/f7fde32646f89ea31bcf461147e9f4da08f73c02/src/lib/migrations/lock.ts#L28-L42))
3. the lock has an `updatedAt` field with a TTL index for mongo to discard if it times out
5. On server startup, all instances try to acquire the lock, the one that gets it can run migrations and the other will wait until the lock is released to start the rest of chat-ui.

 ### Migrations

You can find an example migration [here](https://github.com/huggingface/chat-ui/blob/f7fde32646f89ea31bcf461147e9f4da08f73c02/src/lib/migrations/routines/01-update-search-assistants.ts) 

Migrations can have both `up` (required) and `down` (optional) methods (though we don't use the `down` method for now) and optionally they can also set:
```
	runForHuggingChat?: "only" | "never"; // leave unspecified to run for both
```

We should port the migration from https://github.com/huggingface/chat-ui/pull/841 when merging this PR.

### DIFF
--- a/src/lib/migrations/lock.ts
+++ b/src/lib/migrations/lock.ts
@@ -1,72 +1,37 @@
 import { collections } from "$lib/server/database";
 
-// 5min time out on lock
-const LOCK_TIMEOUT = 5 * 60 * 1000;
-
-export async function createAndAcquireLock(): Promise<boolean> {
+export async function acquireLock(key = "migrations") {
 	try {
-		const upsert = await collections.semaphores.updateOne(
-			{},
-			{
-				$setOnInsert: {
-					isDBLocked: true,
-					key: "semaphore",
-					createdAt: new Date(),
-					updatedAt: new Date(),
-				},
-			},
-			{
-				upsert: true,
-			}
-		);
-		return !!upsert.upsertedCount; // true if the document was inserted
+		const insert = await collections.semaphores.insertOne({
+			key,
+			createdAt: new Date(),
+			updatedAt: new Date(),
+		});
+
+		return !!insert.acknowledged; // true if the document was inserted
 	} catch (e) {
 		// unique index violation, so there must already be a lock
 		return false;
 	}
 }
-export async function acquireLock() {
-	const res = await collections.semaphores.findOneAndUpdate(
-		{
-			isDBLocked: false,
-		},
-		{
-			$set: {
-				isDBLocked: true,
-				updatedAt: new Date(),
-			},
-		}
-	);
 
-	return !!res.ok && res.value !== null;
+export async function releaseLock(key = "migrations") {
+	await collections.semaphores.deleteOne({
+		key,
+	});
 }
 
-export async function releaseLock() {
+export async function isDBLocked(key = "migrations"): Promise<boolean> {
+	const res = await collections.semaphores.countDocuments({
+		key,
+	});
+	return res > 0;
+}
+
+export async function refreshLock(key = "migrations") {
 	await collections.semaphores.updateOne(
 		{
-			isDBLocked: true,
-		},
-		{
-			$set: {
-				isDBLocked: false,
-				updatedAt: new Date(),
-			},
-		}
-	);
-}
-
-export async function isDBLocked(): Promise<boolean> {
-	const res = await collections.semaphores.findOne({});
-	if (!res) {
-		throw new Error("No lock found");
-	}
-	return res.isDBLocked;
-}
-
-export async function refreshLock() {
-	await collections.semaphores.updateOne(
-		{
-			isDBLocked: true,
+			key,
 		},
 		{
 			$set: {
@@ -74,21 +39,4 @@ export async function refreshLock() {
 			},
 		}
 	);
-}
-
-export async function discardExpiredLock() {
-	const res = await collections.semaphores.updateOne(
-		{
-			isDBLocked: true,
-			updatedAt: { $lt: new Date(Date.now() - LOCK_TIMEOUT) },
-		},
-		{
-			$set: {
-				isDBLocked: false,
-				updatedAt: new Date(),
-			},
-		}
-	);
-
-	return res.modifiedCount > 0;
 }
--- a/src/lib/migrations/lock.ts
+++ b/src/lib/migrations/lock.ts
@@ -1,0 +1,93 @@
+import { collections } from "$lib/server/database";
+
+// 5min time out on lock
+const LOCK_TIMEOUT = 5 * 60 * 1000;
+
+export async function createAndAcquireLock(): Promise<boolean> {
+	try {
+		const upsert = await collections.semaphores.updateOne(
+			{},
+			{
+				$setOnInsert: {
+					isDBLocked: true,
+					key: "semaphore",
+					createdAt: new Date(),
+					updatedAt: new Date(),
+				},
+			},
+			{
+				upsert: true,
+			}
+		);
+		return !!upsert.upsertedCount; // true if the document was inserted
+	} catch (e) {
+		return false;
+	}
+}
+export async function acquireLock() {
+	const res = await collections.semaphores.findOneAndUpdate(
+		{
+			isDBLocked: false,
+		},
+		{
+			$set: {
+				isDBLocked: true,
+				updatedAt: new Date(),
+			},
+		}
+	);
+
+	return !!res.ok && res.value !== null;
+}
+
+export async function releaseLock() {
+	await collections.semaphores.updateOne(
+		{
+			isDBLocked: true,
+		},
+		{
+			$set: {
+				isDBLocked: false,
+				updatedAt: new Date(),
+			},
+		}
+	);
+}
+
+export async function isDBLocked(): Promise<boolean> {
+	const res = await collections.semaphores.findOne({});
+	if (!res) {
+		throw new Error("No lock found");
+	}
+	return res.isDBLocked;
+}
+
+export async function refreshLock() {
+	await collections.semaphores.updateOne(
+		{
+			isDBLocked: true,
+		},
+		{
+			$set: {
+				updatedAt: new Date(),
+			},
+		}
+	);
+}
+
+export async function discardExpiredLock() {
+	const res = await collections.semaphores.updateOne(
+		{
+			isDBLocked: true,
+			updatedAt: { $lt: new Date(Date.now() - LOCK_TIMEOUT) },
+		},
+		{
+			$set: {
+				isDBLocked: false,
+				updatedAt: new Date(),
+			},
+		}
+	);
+
+	return res.modifiedCount > 0;
+}
--- a/src/lib/migrations/lock.ts
+++ b/src/lib/migrations/lock.ts
@@ -21,6 +21,7 @@ export async function createAndAcquireLock(): Promise<boolean> {
 		);
 		return !!upsert.upsertedCount; // true if the document was inserted
 	} catch (e) {
+		// unique index violation, so there must already be a lock
 		return false;
 	}
 }
--- a/src/lib/migrations/migrations.spec.ts
+++ b/src/lib/migrations/migrations.spec.ts
@@ -1,0 +1,86 @@
+import { afterEach, describe, expect, it } from "vitest";
+import { migrations } from "./routines";
+import {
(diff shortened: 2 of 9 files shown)


nsarrazin:
cc @coyotte508 might be worth taking a look if you have time, especially the following file:
* [`lock.ts`](https://github.com/huggingface/chat-ui/blob/7a384538c021f3a4f50638030efc6d996cb30c33/src/lib/migrations/lock.ts) has all the functions used to create/acquire/release locks
* [This part](https://github.com/huggingface/chat-ui/blob/7a384538c021f3a4f50638030efc6d996cb30c33/src/lib/migrations/migrations.ts#L69-L131) of `migrations.ts` where we run the migrations
* the [indexes](https://github.com/huggingface/chat-ui/blob/c4395b45d971a93209f67e4fe70a0d872250c062/src/lib/server/database.ts#L150-L152) and document definitions ([semaphore](https://github.com/huggingface/chat-ui/blob/c4395b45d971a93209f67e4fe70a0d872250c062/src/lib/types/Semaphore.ts#L7), [MigrationResult](https://github.com/huggingface/chat-ui/blob/c4395b45d971a93209f67e4fe70a0d872250c062/src/lib/types/MigrationResult.ts#L6))

src/lib/server/database.ts
```diff
@@ -119,4 +145,8 @@ client.on("open", () => {
 	assistants.createIndex({ modelId: 1, userCount: -1 }).catch(console.error);
 	reports.createIndex({ assistantId: 1 }).catch(console.error);
 	reports.createIndex({ createdBy: 1, assistantId: 1 }).catch(console.error);
+
+	// Unique index for semaphore and migration results
+	migrationResults.createIndex({ guid: 1 }, { unique: true }).catch(console.error);
+	semaphores.createIndex({ key: 1 }, { unique: true }).catch(console.error);
```
coyotte508, line 151: Should a TTL index be created on `updatedAt` to release the lock in case of crash?

eg `semaphores.createIndex({ updatedAt: 1 },  { expiresAfterSeconds: ... })`

src/lib/migrations/routines/01-update-search-assistants.ts
```diff
@@ -0,0 +1,49 @@
+import type { Migration } from ".";
+import { getCollections } from "$lib/server/database";
+import type { AnyBulkWriteOperation } from "mongodb";
+import type { Assistant } from "$lib/types/Assistant";
+
+const migration: Migration = {
+	guid: "78c019ad-6bac-4958-a8ad-57a880e79bbf",
```
coyotte508, line 7: Any reason for not using `ObjectId`? (which could be the `_id` of the `migrationResults` collection)

src/lib/migrations/migrations.ts
```diff
@@ -0,0 +1,137 @@
...
+			}
+
+			// otherwise all is good and we cna run the migration
+			console.log(`[MIGRATIONS] "${migration.name}" not applied yet. Applying...`);
+
+			await collections.migrationResults.updateOne(
+				{ guid: migration.guid },
+				{
+					$set: {
+						guid: migration.guid,
```
coyotte508, line 97:
```suggestion
```

No need to specify it again here

src/lib/migrations/lock.ts
```diff
@@ -0,0 +1,94 @@
...
+			isDBLocked: true,
+			updatedAt: { $lt: new Date(Date.now() - LOCK_TIMEOUT) },
+		},
+		{
+			$set: {
+				isDBLocked: false,
+				updatedAt: new Date(),
+			},
+		}
+	);
```
coyotte508, line 91: We can let mongodb handle that with TTL index.

src/lib/migrations/lock.ts
coyotte508, line 1: `acquireLock`, `releaseLock`, `isDBLocked` and so on do not rely on `key`, I think they should in case we use locks for something else? (and what's the use of `key` if we don't use it?)

instead of a `isDBLocked` param, why not check the presence or absence? Eg acquiring a lock can be `insertOne`, and an error (on unique constraint) means the lock is already acquired by somebody else. And MongoDB TTL indexes can automatically remove expired locks.

src/lib/server/database.ts
nsarrazin: Yeah, will use a ttl index instead. We will loose the ability to detect when a lock times out I think  (could be used for rollbacks) but we don't do it anyway at the moment

src/lib/migrations/lock.ts
nsarrazin: The idea of the `key` was just to have a unique constant to make sure only one element can exist. But it's easy to make it a parameter indeed, in case we want to reuse the collection in the future. Will do that!

I can replace the `isDBLocked` param and check for presence/absence, but then we will loose the ability to detect a fresh install. Though i can't think of any use cases for this at the moment so it might not be an issue!

src/lib/server/database.ts
nsarrazin: Simplified things, now using a ttl index!

src/lib/migrations/migrations.ts
nsarrazin: fixed

src/lib/migrations/routines/01-update-search-assistants.ts
nsarrazin: Good catch, fixed it!

src/lib/migrations/lock.ts
nsarrazin: Simplified things, i'm now using `insertOne` with a unique index based on the key, and got rid of `isDBLocked`. Thanks for the tip!

coyotte508 (approved):
A bit of an issue with spaces with continued availability - the old app can run for a bit after migrations are run.

So ideally, migrations happen in two steps: a first PR is deployed that fills in the new data (eg search tokens), and a second PR is deployed with the migration.

nsarrazin:
> So ideally, migrations happen in two steps: a first PR is deployed that fills in the new data (eg search tokens), and a second PR is deployed with the migration + the feature visible for users.

If we deploy the data generation before deploying the feature, isn't there a gap between the two deployments where new documents would be missed ?  Not sure if I got it correctly :sweat_smile:

coyotte508:
> > So ideally, migrations happen in two steps: a first PR is deployed that fills in the new data (eg search tokens), and a second PR is deployed with the migration + the feature visible for users.
> 
> If we deploy the data generation before deploying the feature, isn't there a gap between the two deployments where new documents would be missed ? Not sure if I got it correctly 😅

What I mean:

- First PR fills in `searchTokens` when assistants are created / updated, and does nothing else
- Second PR has the migration + the rest of the feature

So:

- Deploy first PR
- When users update assistants, the `searchTokens` fields is updated for them
- Deploy second PR & run migration
- Even if old code runs for a bit, no problems, as `searchTokens` are properly populated by both the migration and the first PR

nsarrazin:
Oh my bad :facepalm: That makes a lot of sense, will keep in mind for deploying #841 cc @mishig25

src/lib/migrations/migrations.ts
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	// check if all migrations have already been run
+	const migrationResults = await collections.migrationResults.find().toArray();
+
+	// if all the migrations._id are in the migrationResults, we can exit early
+	if (
+		migrations.every((m) => migrationResults.some((m2) => m2._id.toString() === m._id.toString()))
+	) {
+		console.log("[MIGRATIONS] All migrations already applied.");
```
mishig25, line 19: maybe there is no need for this statement

src/lib/migrations/migrations.ts
```diff
@@ -0,0 +1,116 @@
...
+	// iterate over all migrations
+	for (const migration of migrations) {
+		// check if the migration has already been applied
+		const existingMigrationResult = migrationResults.find(
+			(m) => m._id.toString() === migration._id.toString()
+		);
+
+		// check if the migration has already been applied
+		if (existingMigrationResult) {
+			console.log(`[MIGRATIONS] "${migration.name}" already applied. Skipping...`);
```
mishig25, line 58: again, I think we can just remove this statement as well

mishig25: shouldnt there also be `continue` under thus block?
`if (existingMigrationResult)`

src/lib/migrations/lock.ts
```diff
@@ -0,0 +1,42 @@
+import { collections } from "$lib/server/database";
+
+export async function acquireLock(key = "migrations") {
```
mishig25, line 3: shouldn't the `key` be the id of the migration ? or it doesn't matter as long as all the functions use same `key` ?

coyotte508: I guess the reasoning is to run the migrations in order, and for that one process only should have the lock on _all_ migrations

src/lib/migrations/migrations.ts
nsarrazin: I did it with if/else but maybe if/continue is cleaner

src/lib/migrations/lock.ts
nsarrazin: Yes that was the idea! if we need locks for other things in the future we can specify a different key. but migrations should be run in order so one lock for all of them made sense I think

src/lib/migrations/migrations.ts
mishig25: is the `if/else`  logic correct? If so, all good

nsarrazin: should be hopefully :laughing:

src/lib/migrations/migrations.ts
nsarrazin: I see your point, ideally we would have log levels because this is a bit verbose, but when people need support for chat-ui I think it'd be nice to know if the migrations are working correctly in the logs. :eyes: 

Maybe we can have log levels in the future like debug/info/warning ?

src/lib/migrations/migrations.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+
+		expect(locks.length).toBe(1);
+		expect(semaphores).toBeDefined();
+		expect(semaphores.length).toBe(1);
+		expect(semaphores?.[0].key).toBe("migrations");
+	});
+
+	it("should read the lock correctly", async () => {
+		expect(await acquireLock()).toBe(true);
+		expect(await isDBLocked()).toBe(true);
```
mishig25, line 27:
```suggestion
		expect(await isDBLocked()).toBe(true);
		expect(await acquireLock()).toBe(false);
```

just making sure

mishig25: also do these tests run as part of github workflow ?

src/lib/migrations/migrations.ts
mishig25: perfect!

mishig25 (approved):
lgtm !

PR #912, opened 2024-03-07
Include last chunk in websearch context

nsarrazin:
While working on RAG for assistants, I noticed that the last chunk was always missing from the websearch context. In my case it was quite important because my context only had one chunk 😆


mishig25 (approved):
lgtm !

PR #915, opened 2024-03-08
Add openai embeddings

Martok88:
This is working, but I haven't tested it extensively. I added a new optional parameter, maxBatchSize, which defaults to 100 if not set in the config. Apparently OpenAI's embedding endpoint can handle batches of 2048, but I also found reports of some users experiencing problems with large batches. 
It still needs a documentation update, and should probably be reworked to use the OpenAI library.

```
TEXT_EMBEDDING_MODELS = `[
  {
    "name": "text-embedding-3-small",
    "displayName": "text-embedding-3-small",
    "description": "OpenAI hosted embedding model",
    "chunkCharLength": 768,
    "endpoints": [
      {
        "type": "openai",
        "url": "https://api.openai.com/v1/embeddings",
      }
    ]
  }
]`
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ TEXT_EMBEDDING_MODELS = `[
 ```
 
 The required fields are `name`, `chunkCharLength` and `endpoints`.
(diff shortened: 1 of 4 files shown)


nsarrazin (approved):

nsarrazin:
I want to do a documentation pass in a later PR so I'll expand on this endpoint as well when doing it :smile:

Martok88:
Looks good!

nsarrazin:
Thanks for the contribution! 🚀

PR #920, opened 2024-03-09
Fix prompt caching on llama.cpp endpoints

reversebias:
In versions of llama.cpp since [3677](https://github.com/ggerganov/llama.cpp/pull/3677), the prompt cache is dropped by the [server](https://github.com/ggerganov/llama.cpp/blob/master/examples/server/README.md?plain=1#L260) unless `cache_prompt: true` is included in the request. 

This change reduces prompt processing times in long chat threads: local inference with large models can have 10s of seconds of processing time for chats with 1000s of context tokens, this massively improves the responsiveness.


nsarrazin (approved):

nsarrazin:
Thanks for the contribution! 🚀

PR #921, opened 2024-03-09
fix: Prompt bypasses login 

### LINKED ISSUE #879: Starting a conversation with a prompt example bypasses the login modal
If I reach the logged out limit for user messages, I should not be able to start more conversations, as it should show a modal telling me to login to continue chatting.

Currently I can bypass it by starting a conversation by clicking on one of the 4 prompt examples below an assistant/model. It still fails to generate and shows the correct error message, but it would be nice if it showed the login modal instead.

KartikGS:
fixes #879 
[Prompt_login.webm](https://github.com/huggingface/chat-ui/assets/94371461/8f881416-e426-4668-8acf-4b9a838f3cda)

### DIFF
--- a/src/lib/components/chat/ChatWindow.svelte
+++ b/src/lib/components/chat/ChatWindow.svelte
@@ -195,9 +195,30 @@
 					model={currentModel}
 				/>
 			{:else if !assistant}
-				<ChatIntroduction {models} {currentModel} on:message />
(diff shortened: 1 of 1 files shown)


nsarrazin (approved):
Works perfectly, thanks for the fix! 🚀

PR #923, opened 2024-03-11
Anthropic Endpoint Support

Saghen:
Adds support for Anthropic via their official SDK and the newer [Messages API](https://docs.anthropic.com/claude/reference/messages-streaming). Left a todo regarding the `preprompt` which I passed to Anthropic via their `system` prompt option, rather than passing it in during `buildPrompt`. Let me know which would be preferable
https://github.com/huggingface/chat-ui/pull/923/files#diff-ef33885c5572e45b8c3ba8c5d6b489b258dc0a3fdd7bef12556481f71f5cd45fR53-R54

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,6 +82,7 @@
 		"zod": "^3.22.3"
 	},
 	"optionalDependencies": {
+		"@anthropic-ai/sdk": "^0.17.1",
 		"aws4fetch": "^1.0.17",
 		"openai": "^4.14.2"
 	}
--- a/src/lib/server/endpoints/anthropic/endpointAnthropic.ts
+++ b/src/lib/server/endpoints/anthropic/endpointAnthropic.ts
@@ -1,0 +1,95 @@
+import { z } from "zod";
+import { ANTHROPIC_API_KEY } from "$env/static/private";
+import type { Endpoint } from "../endpoints";
+import type { TextGenerationStreamOutput } from "@huggingface/inference";
+
+export const endpointAnthropicParametersSchema = z.object({
+	weight: z.number().int().positive().default(1),
+	model: z.any(),
+	type: z.literal("anthropic"),
+	baseURL: z.string().url().default("https://api.anthropic.com"),
+	apiKey: z.string().default(ANTHROPIC_API_KEY ?? "sk-"),
+	defaultHeaders: z.record(z.string()).optional(),
+	defaultQuery: z.record(z.string()).optional(),
+});
+
+export async function endpointAnthropic(
+	input: z.input<typeof endpointAnthropicParametersSchema>
+): Promise<Endpoint> {
+	const { baseURL, apiKey, model, defaultHeaders, defaultQuery } =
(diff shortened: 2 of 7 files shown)


julien-c:
cool addition!

ptrandev:
ok go off king

src/lib/server/endpoints/anthropic/endpointAnthropic.ts
```diff
@@ -0,0 +1,89 @@
...
+				messages: [{ role: "user", content: prompt }],
+				max_tokens: model.parameters?.max_new_tokens,
+				temperature: model.parameters?.temperature,
+				top_p: model.parameters?.top_p,
+				top_k: model.parameters?.top_k,
+				stop_sequences: model.parameters?.stop,
+				// TODO: is it correct to provide the preprompt here since the client
+				// natively supports it? or in the buildPrompt function?
+				system: preprompt,
+			});
```
nsarrazin, line 56:
```suggestion
	return async ({ messages, preprompt }) => {
		let system = preprompt;
		if (messages?.[0]?.from === "system") {
			system = messages[0].content;
		}

		const messagesFormatted = messages
			.filter((message) => message.from !== "system")
			.map((message) => ({
				role: message.from,
				content: message.content,
			})) as unknown as {
			role: "user" | "assistant";
			content: string;
		}[];

		let tokenId = 0;
		return (async function* () {
			const stream = anthropic.messages.stream({
				model: model.id ?? model.name,
				messages: messagesFormatted,
				max_tokens: model.parameters?.max_new_tokens,
				temperature: model.parameters?.temperature,
				top_p: model.parameters?.top_p,
				top_k: model.parameters?.top_k,
				stop_sequences: model.parameters?.stop,
				system,
			});
```

			
			
This should do the trick. `buildPrompt` is not needed in this case, it's used for converting an array of message to a formatted string for APIs that don't have a messages interface, but in the case of anthropic looks like you can just pass an array of message directly. So we format it correctly (they don't support `role: "system"`) and pass the system prompt separately in `system`

src/lib/server/endpoints/anthropic/endpointAnthropic.ts
```diff
@@ -0,0 +1,89 @@
+import { z } from "zod";
+import { buildPrompt } from "$lib/buildPrompt";
```
nsarrazin, line 2:
```suggestion
```

Don't need this anymore 😁

package.json
```diff
@@ -50,6 +50,7 @@
 	},
 	"type": "module",
 	"dependencies": {
+		"@anthropic-ai/sdk": "^0.17.1",
```
nsarrazin, line 53: `@anthropic-ai/sdk` should be installed optionally, like the other inference specific packages

`npm i --save-optional @anthropic-ai/sdk`

Saghen:
Thanks! Made the changes and rebased onto main

nsarrazin (approved):
Works great! Thanks for the contribution :rocket:

PR #938, opened 2024-03-18
[Assistants] trending feature

mishig25:
### Description

There is a new field `assistant.last24HoursCount` that will be used to sort assistants in `trending` manner.

```js
last24HoursCount:{
	count: 3356, // total number of requests an assistant received over the last 24 hours
	byHour: {
		0: 567, // number of requests an assistant received in the last 24 hours from 01:00-02:00
		1: 928, // number of requests an assistant received in the last 24 hours from 02:00-03:00
		2: 734,
		...
		23: 374, // number of requests an assistant received in the last 24 hours from 23:00-24:00
	}
}
```

The trick is: we created a 24-hour window rolling counter for assistants by running the function below every hour:
```js
const hourExpired = new Date().getUTCHours() + 1;
// then in mongo database
assistant.last24HoursCount.count -= assistant.last24HoursCount[hourExpired]; // 24-hour window rolling counter
assistant.last24HoursCount[hourExpired] = 0; // reset
```

using semaphore/lock/migration helpers from https://github.com/huggingface/chat-ui/pull/897 to make sure that this periodic function runs correctly & runs only on one instance (no double running)

### Screenshots

| Desktop    | Mobile |
| -------- | ------- |
| Screenshot 2024-03-18 at 11 46 02 AM  |Screenshot 2024-03-18 at 11 46 12 AM    |

### Deployment plan

1. Get this PR approved
2. Hide the sort assistants UX/frontend while the data is being collected
3. Deploy
4. Enable/show sort assistants UX/frontend


coyotte508 (commented):
A few comments:

- You can use a dedicated collection for assistant analytics, with a unique key on `date` (truncated to the hour) + assistant's `_id`

```ts
await collections.assistantStats.updateOne({
  // or `"date.at": startOfHour(new Date()) - with `date-fns` 
  "date.at": {$dateTrunc: {date: new Date(), unit: "hour"}},
  assistantId: ...
}, {
  $inc: {count: 1}
}, {upsert: true});
```

This will be more robust 

And periodically you can run a `$merge` aggregation query:

```ts
await collections.assistantStats.aggregate([..., {$merge: {into: "assistants"}})
await collections.assistants.updateMany({
  "last24HoursCount.count": {$gt: 0},
  "last24HoursCount.updatedAt": ...
}, {$set: {"last24HoursCount.count": 0}});
```

- The "cron" shouldn't be in "migrations.ts". There should maybe be an env var to enable it. And you can reuse the `collections` exported from `src/lib/server/database.ts`, no need to create a separate DB connection.

eg

```ts
setInterval(refreshAssistantsStats, 3600_000)

async function refreshAssistantsStats() {
  if (!cantLockDb()) {
    return;
  }

  // run DB queries
}
```

gary149 (approved):
Ok for UI

src/lib/assistantStats/refresh-assistants-counts.ts
```diff
@@ -0,0 +1,119 @@
...
+				},
+			},
+			{
+				$merge: {
+					into: "assistants",
+					on: "_id",
+					whenMatched: [
+						{
+							$set: {
+								"last24HoursCount.count": "$$ROOT.totalCount",
```
coyotte508, line 32: I don't think you need to put it in a `Migration` anymore?

mishig25:
based on https://github.com/huggingface/chat-ui/pull/938#pullrequestreview-1942863254, I've committed https://github.com/huggingface/chat-ui/pull/938/commits/93c9ebc3ddab1c436f2210b9a1b5c839160ff14d

Description:
1. Created a new collection called `AssistantStats`. Fields are: `_id: Assistant["_id"], dateWithHour: Date, count: number`
2. `dateWithHour` is a date field. However, the value should have minutes, seconds, ms truncated. [`let date = new Date(); date.setUTCMinutes(0, 0, 0)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setUTCMinutes) so that we can use this field `dateWithHour` to increment `count` field everytime an assistant gets a visit within same hour
3. Every hour, there is a "cron-like" aggregation query that does: `for the last 24 hours, group on _id: Assistant["_id"] & sum count: number` and afterwards using [$merge.whenMatched pipeline](https://www.mongodb.com/docs/manual/reference/operator/aggregation/merge/#std-label-merge-whenMatched-pipeline) update `assistant.last24HoursCount` field

src/lib/assistantStats/refresh-assistants-counts.ts
mishig25: I was in assumption that multiple machines are running chat-ui (and the call below might be called my multiple machines within a same period). therefore, we would need to use Migration & locks. No ?

https://github.com/huggingface/chat-ui/blob/58f3eace2bf698132be68d5a69b2f61b88a0111f/src/hooks.server.ts#L26

coyotte508: locks yes, but no need for `Migration`

We can create a lock on `refresh-assistant-counts`

mishig25: handled in 08037405fa7c562f68501ebbfb4b1150f8e5a4ae

1. Removed usage of `Migration`
2. Also set those lines below to make sure that different machines will try to run the refresh "cron" job at the same time (i.e. synced even if the machines start at different times)

https://github.com/huggingface/chat-ui/blob/08037405fa7c562f68501ebbfb4b1150f8e5a4ae/src/lib/assistantStats/refresh-assistants-counts.ts#L70-L79

coyotte508:
Did several backend changes

- For stats collection, the `_id` shouldn't be `assistantId` (otherwise we can create only one document per assistant, while we should have 1 / hour / assistant)
- added `date.at` and `date.span` fields to be consistent with other stat collections - also in case we want to be more granular in the future (instead of 1 doc/hour, 1 doc/10 minutes for example)
- Reworked the locking mechanism
	- Made `lockKey` mandatory
	- Return the `lockId`, to make sure we return / release a lock we own
- Maintain the lock for the cron job during the whole process's lifespan -  no need to sync job start
- Changed the aggregation to use `$unionWith` - a nice trick originating from @Kakulukian I think. No need for a `lastUpdatedAt` field with this!
- Fixed a bug on usage limits for assistant creation 

Maybe the lock mechanism (`maintainLock`) can be more standardized if we have more cron jobs

coyotte508 (approved):

mishig25:
@nsarrazin wdyt ?

nsarrazin (approved):
Looks good to me! If I can help with the deployment let me know 😁

mishig25:
checking one more time from my end & will ping you on deployment

mishig25:
as written in the [dscription](https://github.com/huggingface/chat-ui/pull/938#issue-2191919264)

> Deployment plan
> 1. Get this PR approved
> 2. Hide the sort assistants UX/frontend while the data is being collected
> 3. Deploy
> 4. Enable/show sort assistants UX/frontend

I've hidden the UI https://github.com/huggingface/chat-ui/pull/938/commits/4dd7e65e4ac9c81d5fb5e0277360fda43d818bde

Ready to merge & deploy @nsarrazin

PR #940, opened 2024-03-18
Store user agent in conversation

nsarrazin:
cc @alak, will this also work on your side ?


alak:
@nsarrazin yes I already add a dedicated user agent in all my network calls.

Issue #611, opened 2023-12-05
Feature Request: Custom slash-commands


secondtruth:
Inspired by #464, I want to suggest custom slash-commands defined in the env (on model level and/or global level), that will be expanded to text, maybe using handlebars templates. This could also be useful for inserting user-defined prompts by name.

**Example:**

    [
      {
        "command": "news",
        "content": "Summarize today's news. Today's date: {{date}}"
      }, {
        "command": "words-starting-with",
        "num_args": 2,
        "content": "Find {{args[1]}} words starting with {{args[0]}}"
      }
    ]

- Command 1: `/news`
- Command 2: `/words-starting-with a, 3`

PR #942, opened 2024-03-19
Set assistant prompt from URL

nsarrazin:
### Assistant view
image
### Settings edit/new
image

This uses the same `parseWeb` method we use in the web search, but instead of chunking and generating embeddings we just add the resulting text to the system prompt.

### DIFF
--- a/src/lib/components/AssistantSettings.svelte
+++ b/src/lib/components/AssistantSettings.svelte
@@ -74,12 +74,14 @@
 
 	let loading = false;
 
-	let ragMode: false | "links" | "domains" | "all" = assistant?.rag?.allowAllDomains
+	let ragMode: false | "links" | "domains" | "all" | "preprompt" = assistant?.rag?.allowAllDomains
 		? "all"
 		: assistant?.rag?.allowedLinks?.length ?? 0 > 0
 		? "links"
 		: (assistant?.rag?.allowedDomains?.length ?? 0) > 0
 		? "domains"
+		: (assistant?.rag?.prepromptUrl ?? "").length > 0
+		? "preprompt"
 		: false;
 </script>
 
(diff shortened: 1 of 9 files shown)


src/lib/components/AssistantSettings.svelte
```diff
@@ -119,18 +121,29 @@
...
 			formData.set("ragAllowAll", "false");
 			formData.set("ragLinkList", "");
+			formData.set("ragPrepromptUrl", "");
+		} else if (ragMode === "preprompt") {
+			formData.set("ragAllowAll", "false");
+			formData.set("ragLinkList", "");
+			formData.set("ragDomainList", "");
+			formData.set(
+				"preprompt",
+				"The system prompt input is disabled when using a system prompt URL."
```
gary149, line 144: We could just refactor later no? I don't get why we need this list this should be handled natively by the <form>.

src/routes/conversation/+server.ts
```diff
@@ -51,11 +53,23 @@ export const POST: RequestHandler = async ({ locals, request }) => {
 	});
 
 	if (assistant) {
-		values.preprompt = assistant.preprompt;
-	} else {
-		values.preprompt ??= model?.preprompt ?? "";
+		console.log(assistant?.rag?.prepromptUrl);
```
gary149, line 56:
```suggestion

```

src/routes/conversation/+server.ts
```diff
@@ -51,11 +53,22 @@ export const POST: RequestHandler = async ({ locals, request }) => {
 	});
 
 	if (assistant) {
-		values.preprompt = assistant.preprompt;
-	} else {
-		values.preprompt ??= model?.preprompt ?? "";
+		// if the assistant has a preprompt Url and it's not local, try to fetch it
+		if (assistant.rag?.prepromptUrl && !(await isURLLocal(new URL(assistant.rag.prepromptUrl)))) {
+			try {
+				const content = await parseWeb(assistant.rag.prepromptUrl);
```
gary149, line 59: I would not parse like for web content.

secondtruth:
I think, this will be a very useful feature. 👍

But I would suggest to make the internet access features composable, so that web search, prompts from URLs, etc. can be used together if desired. In that case, the settings UI could be structured this way:

> **Internet access**
> - [Checkbox] **Web search** – _expands the options:_
>   - (Radio) **Everywhere**
>   - (Radio) **Specific Domains**
>   - (Radio) **Specific Links**
> - [Checkbox] **System Prompt URL**

---

**Related to this feature:**
- #611 (Part "**Prompts from external APIs**")

nsarrazin:
Closing in favour of #949

PR #944, opened 2024-03-19
[Websearch] change context schema

mishig25:
### With this PR

Put the `context` text inside respective `contextSources` (suggested by @gary149 )

image

note: nothing is needed in terms of backward compatiblity. This change doesn't break anything from the old.

### DIFF
--- a/src/lib/server/preprocessMessages.ts
+++ b/src/lib/server/preprocessMessages.ts
@@ -11,8 +11,14 @@ export async function preprocessMessages(
 ): Promise<Message[]> {
 	return await Promise.all(
 		structuredClone(messages).map(async (message, idx) => {
+			const webSearchContext = webSearch?.contextSources
+				.map(({ context }) => context)
+				.flat()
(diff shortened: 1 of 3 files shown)


src/lib/server/preprocessMessages.ts
```diff
@@ -11,8 +11,14 @@ export async function preprocessMessages(
 ): Promise<Message[]> {
 	return await Promise.all(
 		structuredClone(messages).map(async (message, idx) => {
+			const webSearchContext = webSearch?.contextSources
+				.map(({ context }) => context)
+				.flat()
+				.sort((a, b) => a.idx - b.idx)
+				.map(({ text }) => text)
+				.join(" ");
```
gary149, line 19: Is this the best place to do this, does it ensure that the right sources are mapped to the right context? 👀

nsarrazin (approved):

gary149 (approved):

PR #948, opened 2024-03-20
Add Archyve search

### LINKED ISSUE #909: Add support for prompt augmentation via external API
## Problem

I love HF chat-ui, and I'd like to deploy it for more teams. However, missing RAG features prevents me from deploying it as widely as I'd like.

Some teams need RAG with PDFs as a data source. Many of these PDFs are oddly formatted, and different types of PDFs require different kinds of parsing/chunking/embedding/whatever. This is not chat-ui's concern - it's mine - and no feature of any LLM chat UI will ever solve this problem in the way I need it solved.

## Possible solution

What I'd really like is to have a feature where I can tell chat-ui to get a prompt from the user, call a ReST API to have the prompt translated/augmented/whatever, and then send the resulting prompt to the LLM. (Some way to hover and see what the actual, augmented prompt looked like would also be nice, in case something weird happens and the user wants to know why.)

I'll build the indexing system and present the ReST API to augment prompts, I just need a UI that will use it. I would even be happy to use an existing API as reference or adopt a standard if one exists, but I haven't seen one. Maybe we'll set the standard here.

## My plan

I was thinking of forking chat-ui to add these hooks to do "prompt translation" or "prompt augmentation" or whatever the best name is.

## Questions

Any advice on where I start? Where in the code would this logic go if I added it?

Are maintainers open to the idea of merging a feature like this if it works?

(Because you never know) Is this already present in chat-ui and I just haven't noticed?

oxaronick:
Add support for using Archyve to search a document library for additional context to add to the prompt. Closes #909.

- [x] change `WebSearchProvider` enum to `WebSearchProviders` object
- [x] add info to web search provider about how to template queries and whether to generate a better search query using an LLM
- [x] call Archyve API to get document chunks
- [x] investigate skipping internal chunking and similarity search, since Archyve has already provided relevant chunks based on similarity search
- [x] ~~write tests?~~ I don't see many tests in the code, but lmk if this PR would benefit from testing a particular component
- [x] limit number/size of chunks used in context

@nsarrazin Thoughts on the general direction I'm going?

### Testing

To test, I used an actual RFP from the Government of Canada, which is about the worst PDF I can think of. I uploaded the document into Archyve and integrated ChatUI with the Archyve API. Here is the result:
image

the Archyve API is documented at [archyve.io](https://archyve.io).

### DIFF
--- a/src/lib/server/websearch/runWebSearch.ts
+++ b/src/lib/server/websearch/runWebSearch.ts
@@ -32,6 +32,7 @@ export async function runWebSearch(
 	ragSettings?: Assistant["rag"]
 ) {
 	const prompt = messages[messages.length - 1].content;
+	const searchProvider = getWebSearchProvider();
 	const webSearch: WebSearch = {
 		prompt,
 		searchQuery: "",
@@ -39,6 +40,7 @@ export async function runWebSearch(
 		contextSources: [],
 		createdAt: new Date(),
 		updatedAt: new Date(),
+		provider: searchProvider,
 	};
 
 	function appendUpdate(message: string, args?: string[], type?: "error" | "update") {
@@ -71,9 +73,12 @@ export async function runWebSearch(
 				return { link, hostname: new URL(link).hostname, title: "", text: "" };
 			});
 		} else {
-			webSearch.searchQuery = await generateQuery(messages);
-			const searchProvider = getWebSearchProvider();
-			appendUpdate(`Searching ${searchProvider}`, [webSearch.searchQuery]);
+			if (webSearch.provider.generateQuery) {
+				webSearch.searchQuery = await generateQuery(messages);
+			} else {
+				webSearch.searchQuery = prompt;
+			}
+			appendUpdate(`Searching ${searchProvider.name}`, [webSearch.searchQuery]);
 
 			let filters = "";
 			if (ragSettings && ragSettings?.allowedDomains.length > 0) {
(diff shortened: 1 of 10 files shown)


gary149:
What is Archyve?

secondtruth:
Good work, as far as I can say. 👍

May I ask what Archyve is? I couldn't find it using Google.

oxaronick:
> What is Archyve?

Archyve is a Document Library we're building, which will include a vector database. Its API can be used by RAG apps to perform similarity searches on pre-indexed documents.

The idea is that different kinds of documents have different, and sometimes strange, indexing requirements. Archyve will handle that so apps like Chat UI only need to fetch context to include in a prompt. I'm also hoping that by defining an Archyve OpenAPI spec, it can be used by others implementing their own Document Libraries.

oxaronick:
> Good work, as far as I can say. 👍

Thanks, good to hear.

> May I ask what Archyve is? I couldn't find it using Google.

See other reply; code and website coming soon.

oxaronick:
I'm going to leave this as draft until you can see Archyve, but still would like to know if the direction I've taken is okay or not.

src/lib/server/websearch/runWebSearch.ts
```diff
@@ -111,58 +116,69 @@ export async function runWebSearch(
...
 
-		// fetch the model
-		const embeddingModel =
-			embeddingModels.find((m) => m.id === conv.embeddingModel) ?? defaultEmbeddingModel;
+		const usedSources = new Set<string>();
+		const topKClosestParagraphs = 8;
+		if (webSearch.provider.doSimilaritySearch) {
+			// fetch the model
+			const embeddingModel =
+				embeddingModels.find((m) => m.id === conv.embeddingModel) ?? defaultEmbeddingModel;
```
oxaronick, line 124: I moved these two `const`s up so they'd be available to the `if` branch and the `else` branch. Otherwise, the existing code is unchanged, just moved into the `if` branch here.

oxaronick:
Archyve site is up with preliminary API docs. Very simple for v1:

https://archyve.io/

(`nickthecook` is my personal GitHub account.)

oxaronick:
@nsarrazin I can get this to work, but I don't actually like the approach of implementing Archyve as a WebSearch provider:

1. Archyve's API requires authentication to fetch the actual chunks of text so source links in updatePad don't work
2. It's not "the web", so the toggle called "Web search" on the conversation page is misleading
3. ChatUI does similarity search on the results, which makes sense for web results, but Archyve results are already the output of a similarity search

It's starting to feel more like "Web" is one kind of search source, and I should add a second kind, like "Document Library". In the latter search source, ChatUI won't have to do nearly as much work, or have conditionals like `if (webSearch.provider.doSimilaritySearch)` throughout its code.

Thoughts?

oxaronick:
I'm going to take this down. No response in a while, and it doesn't feel like a great fit for Chat-UI.

PR #950, opened 2024-03-21
Google Vertex API support

madppiper:
This commit adds support for the google cloud vertex API.   It relies on https://github.com/googleapis/nodejs-vertexai to connect.

To enable: 
1.  [Select](https://console.cloud.google.com/project) or [create](https://cloud.google.com/resource-manager/docs/creating-managing-projects#creating_a_project) a Google Cloud project.
1.  [Enable billing for your project](https://cloud.google.com/billing/docs/how-to/modify-project).
1.  [Enable the Vertex AI API](https://console.cloud.google.com/flows/enableapi?apiid=aiplatform.googleapis.com).
1.  [Set up authentication with a service account](https://cloud.google.com/docs/authentication/getting-started)
    so you can access the API from your local workstation.

The service account credentials file can be imported as an environmental variable:

```env
    GOOGLE_APPLICATION_CREDENTIALS = clientid.json
```

Make sure docker has access to the file. Afterwards Google Vertex endpoints can be configured as following:

```
MODELS=`[
//...
    {
       "name": "gemini-1.0-pro", //model-name
       "displayName": "Vertex code-bison",
       "location": "europe-west3",
       "apiEndpoint": "", //alternative api endpoint url
       "endpoints" : [{
         "type": "vertex"
       }]
     },
]`

```

Tested locally, but I had to merge it over to a new branch, to make the reviewing process easier - so please evaluate. Setting up the google permissions can be a bit tricky and model availability can differ for location & your own access level. So beware of this if a connection fails.

### DIFF
--- a/README.md
+++ b/README.md
@@ -528,6 +528,43 @@ You can also set `"service" : "lambda"` to use a lambda instance.
 
 You can get the `accessKey` and `secretKey` from your AWS user, under programmatic access.
 
+##### Google Vertex models
+
+Chat UI can connect to the google Vertex API endpoints ([List of supported models](https://cloud.google.com/vertex-ai/generative-ai/docs/learn/models)).
+
+To enable: 
+1.  [Select](https://console.cloud.google.com/project) or [create](https://cloud.google.com/resource-manager/docs/creating-managing-projects#creating_a_project) a Google Cloud project.
+1.  [Enable billing for your project](https://cloud.google.com/billing/docs/how-to/modify-project).
+1.  [Enable the Vertex AI API](https://console.cloud.google.com/flows/enableapi?apiid=aiplatform.googleapis.com).
+1.  [Set up authentication with a service account](https://cloud.google.com/docs/authentication/getting-started)
+    so you can access the API from your local workstation.
+
+The service account credentials file can be imported as an environmental variable:
+
+```env
+    GOOGLE_APPLICATION_CREDENTIALS = clientid.json
+```
+
+Make sure to grant access to the directory in your docker file. Afterwards Google Vertex endpoints can be configured as following:
+
+```
+MODELS=`[
+//...
+    {
+       "name": "gemini-1.0-pro", //model-name
+       "displayName": "Vertex code-bison",
+       "location": "europe-west3",
+       "apiEndpoint": "", //alternative api endpoint url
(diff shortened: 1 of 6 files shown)


nsarrazin:
Hi! Thanks for the contribution, I'll take a look soon 😄

madppiper:
> Hi! Thanks for the contribution, I'll take a look soon 😄

No problem - you guys are doing a fantastic job. Let me know if this requires any improvements.

madppiper:
> Two small things I noticed:
> 
> * Could you run `npm run format` so the linter checks pass? 😁
> * Can you install `@google-cloud/vertexai` as an optional dependency like for the other endpoints?

Yea, of course. I just pushed both.


> And I was wondering, would it be bad practice to put the content of the JSON credentials into an env variable secret rather than specifying the file path and reading the file? Trying to read the file could cause quite some headaches with different infra & docker setups I imagine, but not sure if it's safe to just pass the JSON content directly.

I agree and I am not a fan either. But I checked against the google docs and it was my impression that the json file was the standard way of doing it with google. Personally I would have preferred an apiToken (which funnily enough you can generate, but isn't supported by the node api of theirs):

https://github.com/googleapis/nodejs-vertexai/issues/197

This is the doc in question: 
https://cloud.google.com/docs/authentication/application-default-credentials#GAC

>GOOGLE_APPLICATION_CREDENTIALS environment variable
>You can use the GOOGLE_APPLICATION_CREDENTIALS environment variable to provide the location of a credential JSON file. This JSON file can be one of the following types of files: ...


I should mention: there are still a few things untested from my part - though connecting to Gemini-Pro works, I couldn't test it with any other model, nor could I set up a custom endpoint, as I was lacking an account with the proper privileges and or available resources. So i did the apiUrl implementation blindly. 

It could also use a more thorough test of the parameters. In principle it should work, but I only tested it as far as it was required from our end.

analyticsrepo01:
Awesome, +1 to check if merge is approved

nsarrazin (approved):
Thanks for the explanation! 😄 I think we can merge this now, looks good to me

nsarrazin (approved):

nsarrazin:
lol you got double approved apparently 😆 thanks for the contribution!

PR #951, opened 2024-03-21
add api assistants endpoints

alak:
Adding 2 endpoints to the API:

/api/assistants: A paginated endpoint to list/search across all assistants

/api/user/assistants: A endpoint to list current user's assistants and the ones he already used in the past

### DIFF
--- a/src/routes/api/assistants/+server.ts
+++ b/src/routes/api/assistants/+server.ts
@@ -1,0 +1,51 @@
+import { base } from "$app/paths";
+import { collections } from "$lib/server/database.js";
+import type { Assistant } from "$lib/types/Assistant";
+import type { User } from "$lib/types/User";
+import { generateQueryTokens } from "$lib/utils/searchTokens.js";
+import type { Filter } from "mongodb";
+
+const NUM_PER_PAGE = 24;
+
+export async function GET({ url, locals }) {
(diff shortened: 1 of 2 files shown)


nsarrazin (approved):
Seems to work well in local testing! Do you also need to expose the endpoints for subscribing/creating/editing assistants or can you use the form actions directly?

coyotte508:
At one point error response between the API endpoint and the `hooks.server.ts` should be standardized cc @nsarrazin , for now in the hooks the error response looks like this:

```ts
	function errorResponse(status: number, message: string) {
		const sendJson =
			event.request.headers.get("accept")?.includes("application/json") ||
			event.request.headers.get("content-type")?.includes("application/json");
		return new Response(sendJson ? JSON.stringify({ error: message }) : message, {
			status,
			headers: {
				"content-type": sendJson ? "application/json" : "text/plain",
			},
		});
	}
```

It sends `{error: ...}` instead of `{message: ...}`

Or we could create a `handleError` in `hooks.server.ts` and throw errors in the api endpoints maybe

alak:
> Seems to work well in local testing! Do you also need to expose the endpoints for subscribing/creating/editing assistants or can you use the form actions directly?

Yes that's the next step

PR #952, opened 2024-03-22
🚸 When a user signs up they're directly logged in

coyotte508:
Currently it seems a user, when logging in for the first time, needs to

- sign up (with "sign in with HF")
- sign in

Because the sign up doesn't send the new cookie value to the user


nsarrazin (approved):
Thank you!! I've been trying on and off to figure out what was going wrong, couldn't track it down, merging right away :tada:

alak (approved):

PR #954, opened 2024-03-25
Limit number of fetched conversations in the API

nsarrazin:
We were missing the same limit that we use in the load function of the web app.

### DIFF
--- a/src/routes/api/conversations/+server.ts
+++ b/src/routes/api/conversations/+server.ts
@@ -2,7 +2,9 @@ import { collections } from "$lib/server/database";
 import { authCondition } from "$lib/server/auth";
 import type { Conversation } from "$lib/types/Conversation";
 
-export async function GET({ locals }) {
+export async function GET({ locals, url }) {
+	const p = parseInt(url.searchParams.get("p") ?? "0");
+
 	if (locals.user?._id || locals.sessionId) {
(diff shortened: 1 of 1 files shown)


nsarrazin:
cc @alak, we had a limit in the load function for the web app to avoid loading too many conversations at once, ported it over to the API endpoint with added pagination 😄

alak (approved):
LGTM.

I'll update my side today.

src/routes/api/conversations/+server.ts
```diff
@@ -14,6 +16,8 @@ export async function GET({ locals }) {
 				model: 1,
 			})
 			.sort({ updatedAt: -1 })
+			.skip(300 * p)
+			.limit(300)
```
mishig25, line 20:
```suggestion
			.skip(NUM_PER_PAGE * p)
			.limit(NUM_PER_PAGE)
```

and declating `const NUM_PER_PAGE = 300;` above would be more robust since `skip` & `limit` should be tied with one another through a variable

nsarrazin: Should we make it an env var? So it's also tied to the one in the load function ?

mishig25: up to you. For me, it is fine as long as it is a variable here

PR #957, opened 2024-03-26
Remove commented out code in tailwind config

nsarrazin:
We were not using Inter as a font for the app, even though we use it for thumbnails & figma.

This should fix it?


gary149:
Not sure the system font is nice for this no? (zero loading time)

nsarrazin:
Ok, there was leftover commented out code in the tailwind config so I assumed it was forgotten. Will update the PR to remove the commented out code

PR #958, opened 2024-03-26
feat: add support for anthropic on vertex

cheahjs:
Duplicates the existing Anthropic support, but targeting [Claude 3 models hosted on Vertex AI](https://cloud.google.com/vertex-ai/generative-ai/docs/partner-models/use-claude) by using the `AnthropicVertex` client instead. Authentication is done using Google Application Default Credentials.


nsarrazin:
Hey 👋 Thanks for the contrib! Just to clarify this is unrelated to #950 ? There's no way to reuse the endpoint type from either anthropic or vertex to achieve this?

nsarrazin (approved):
Sorry for the delay with this! PR looks good, I changed things to make `projectId` mandatory in the endpoint parameters and removed the top level env variable but other than that LGTM, thanks for contributing!

Issue #955, opened 2024-03-25
Assistants: expose some models params


gary149:
<img width="1271" alt="image" src="https://github.com/huggingface/chat-ui/assets/3841370/f88a26f4-cfbf-4bfe-9e0c-6e26d84c8b06">
<img width="1757" alt="image" src="https://github.com/huggingface/chat-ui/assets/3841370/7aac7c8c-7b6c-4622-ae31-5cbc28fe775b">

PR #959, opened 2024-03-27
 Expose sampling controls in assistants (#955) 

nsarrazin:
| Screenshot 2024-03-27 at 13 31 03 | Screenshot 2024-03-27 at 13 29 59 |
| ------- | ------- | 

It opens by default if values are prefilled. I updated the endpoint code so it should work with every endpoint type. There's a tooltip for each field that works on mobile & desktop.

Had to deviate a bit from the design in #955, it was hard to fit the label and the field horizontally (only looked good on big screen at full width) so I ended up putting the label above the field. 

Also used the native html details/summary elements instead of a button, and added a border around the content, otherwise it was easy to confuse with the user start messages below.

### DIFF
--- a/src/lib/components/AssistantSettings.svelte
+++ b/src/lib/components/AssistantSettings.svelte
@@ -272,13 +272,13 @@
 					class="group/details mt-2"
 					open={Object.values(assistant?.generateSettings ?? {}).some((v) => !!v)}
 				>
-					<summary class="ml-1 cursor-pointer text-xs group-open/details:font-semibold">
+					<summary class="cursor-pointer text-xs group-open/details:font-semibold">
 						Model settings
 					</summary>
 					<p class="text-xs text-red-500">{getError("inputMessage1", form)}</p>
-					<div class="mb-2 grid grid-cols-2 grid-rows-2 gap-2">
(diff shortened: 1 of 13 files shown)


gary149 (approved):

src/lib/components/AssistantSettings.svelte
```diff
@@ -31,16 +34,26 @@
...
 		compress = module.readAndCompressImage;
+
+		if (assistant) {
+			modelId = assistant.modelId;
+		} else {
+			if (models.map((model) => model.id).includes($settings.activeModel)) {
+				modelId = $settings.activeModel;
+			} else {
+				modelId = models[0].id;
+			}
```
mishig25, line 55:
```suggestion
			modelId = models.find((model) => model.id === $settings.activeModel)?.id ?? models[0].id;
```

didn't run it, maybe we can just use this nullish coalescing

mishig25 (approved):
lgtm !

src/lib/components/AssistantSettings.svelte
nsarrazin: tested it, it works, thanks!

Issue #961, opened 2024-03-27
Assistant: try from the settings


gary149:
today the flow is bloated, could work like in my screenshots, but maybe hard to implement? 

![image](https://github.com/huggingface/chat-ui/assets/3841370/b6dcfcfc-c407-4702-8769-aae4e4ada8c5)
![image](https://github.com/huggingface/chat-ui/assets/3841370/9a150c1e-269f-4af3-9ecb-0d56c3935751)

nsarrazin:
I think for implementation we would need to store "draft" assistants in the DB and clean them up periodically, as they need to be stored server-side to be used in conversations. Doable but not sure of the UX of trying out, the side panel on top of a modal might be a lot?

Could we just make a simpler flow by having the "save" button in the edit form taking you to a new conversation with the assistant? 

Or maybe the "Activate"/"Active" button could turn into a "Start chatting" button on this page when it's selected?
![image](https://github.com/huggingface/chat-ui/assets/25119303/d91067ce-72b4-4199-9bd3-62afa3224611)

gary149:
> Or maybe the "Activate"/"Active" button could turn into a "Start chatting" button on this page when it's selected?

ok to replace "Activate" to "Start chatting" everywhere. (and it close the settings and it opens a new conv with the right model/assistant)

PR #963, opened 2024-03-29
Add start chat button on models & assistants.

nsarrazin:
image

To help with #961, the idea is that when the model is active, instead of just showing "Active" we show a button you can click to start a chat

### DIFF
--- a/src/routes/settings/(nav)/[...model]/+page.svelte
+++ b/src/routes/settings/(nav)/[...model]/+page.svelte
@@ -92,7 +92,7 @@
 
 	<button
 		class="{isActive
-			? 'bg-gray-100 text-gray-800'
+			? 'bg-gray-200 text-gray-800'
 			: 'bg-black !text-white'} my-8 flex w-fit items-center rounded-full px-3 py-1 text-base"
 		name="Activate model"
 		on:click|stopPropagation={() => {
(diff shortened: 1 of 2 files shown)


gary149:
It's a bit weird in its current state. As you need to click "activate" then "Start chatting" vs from the issue:

> ok to replace "Activate" to "Start chatting" everywhere. (and it close the settings and it opens a new conv with the right model/assistant)

src/routes/settings/(nav)/[...model]/+page.svelte
```diff
@@ -89,15 +92,23 @@
...
+				goto(`${base}/`);
+			} else {
+				$settings.activeModel = $page.params.model;
+			}
 		}}
 	>
-		{isActive ? "Active model" : "Activate"}
+		{#if isActive}
+			<CarbonChat class="mr-1.5 text-sm" />
+			Start Chat
```
Comment (gary149) [inline src/routes/settings/(nav)/[...model]/+page.svelte:108]:
```suggestion
			New Chat
```

nsarrazin:
Handled in [4554a0f](https://github.com/huggingface/chat-ui/pull/963/commits/4554a0f971ccfb3cb1b0a48698cae294ef6101ef) !

Now it looks like this regardless of active status
image

PR #964, opened 2024-03-29
Group tokens randomly

nsarrazin:
To avoid sending packets which have the exact size of the tokens, we group them together at random

### DIFF
--- a/src/routes/conversation/[id]/+server.ts
+++ b/src/routes/conversation/[id]/+server.ts
@@ -409,6 +409,8 @@ export async function POST({ request, locals, params, getClientAddress }) {
 
 			let hasError = false;
 
+			let buffer = "";
+
 			try {
 				const endpoint = await model.getEndpoint();
 				for await (const output of await endpoint({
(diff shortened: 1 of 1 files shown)


nsarrazin:
cc @zacps 😄

src/routes/conversation/[id]/+server.ts
```diff
@@ -409,6 +409,8 @@ export async function POST({ request, locals, params, getClientAddress }) {
 
 			let hasError = false;
 
+			let buffer = "";
```
Comment (mishig25) [inline src/routes/conversation/[id]/+server.ts:412]: Should there be the snippet provided below at the end of the entire process just to be sure that all the tokens were seant to the frontend ?

```ts
if(buffer){
	update({
		type: "stream",
		token: buffer,
	});
}
```

Comment (nsarrazin) [reply src/routes/conversation/[id]/+server.ts:412]: We send the final message anyway to the front-end so it shouldn't change much but will add it for sure just in case

Comment (mishig25) [reply src/routes/conversation/[id]/+server.ts:412]: basically, is there a chance `buffer` was non-empty and you receive the end token. so this non-empty `buffer` doesn't end up getting sent to the frontend ?

mishig25 (approved):
lgtm!

src/routes/conversation/[id]/+server.ts
Comment (nsarrazin) [reply src/routes/conversation/[id]/+server.ts:412]: Solved in [6eb3486](https://github.com/huggingface/chat-ui/pull/964/commits/6eb34862d45b43134a2f8a4e906be1e2091e17b0)!

PR #973, opened 2024-04-03
Switch task model to mistral 7b v0.2

nsarrazin:


gary149:
is it working well?

nsarrazin:
> is it working well?

Now that the v0.1 endpoint is back up, we don't _need_ this.

Would be nice to find a more recent model for the task model though, as mistral v0.1 is deprecated in the UI and only used for tasks now

mishig25:
> Would be nice to find a more recent model for the task model though, as mistral v0.1 is deprecated in the UI and only used for tasks now

I will test gemma for the task

nsarrazin:
We can reopen something when we switch,closing for now :smile:

PR #983, opened 2024-04-07
Update package.json version

zacps:
To correspond with v0.8 release.


nsarrazin (approved):
Thanks for the catch! Will merge and see if #988 also works 😁

PR #985, opened 2024-04-08
Set ChatML as default chat prompt template

nsarrazin:
Wanted to make either chatprompttemplate or tokenizer mandatory, but since some endpoints type use chat completion/messages API, they don't need a chat template as it's handled by their backend.

This PR sets a default (ChatML), so that we won't throw an error when not setting the chat template. This reverts a breaking change.

https://github.com/huggingface/chat-ui/pull/744#issuecomment-2041644232


flexchar:
Thanks!!

PR #998, opened 2024-04-09
Update models and add check for assistants model on startup

nsarrazin:


src/lib/components/AssistantSettings.svelte
```diff
@@ -46,9 +47,9 @@
 		compress = module.readAndCompressImage;
 
 		if (assistant) {
-			modelId = assistant.modelId;
+			modelId = findCurrentModel(models, assistant.modelId).id;
 		} else {
-			modelId = models.find((model) => model.id === $settings.activeModel)?.id ?? models[0].id;
+			modelId = findCurrentModel(models, $settings.activeModel).id;
 		}
```
mishig25, line 53:
```suggestion
		const modelId = findCurrentModel(models, assistant ? assistant.modelId : $settings.activeModel).id;
```

didn't test. but I think we can do it like this

src/lib/components/chat/ChatWindow.svelte
```diff
@@ -353,16 +355,33 @@
...
-					be inaccurate or false.
+						{#if models.find((m) => m.id === currentModel.id)}
+							<a
+								href="{base}/settings/{currentModel.id}"
+								class="inline-flex items-center hover:underline"
+								>{currentModel.displayName}<CarbonCaretDown class="text-xxs" /></a
+							>
+						{:else}
+							<span class="inline-flex items-center line-through dark:border-gray-700">
+								{currentModel.id}
```
mishig25, line 366:
```suggestion
								Model: {currentModel.id}
```

nit: maybe add this label "Model: id" so that it is clear to the user we are talking about model. But up to you

src/lib/migrations/migrations.ts
```diff
@@ -71,8 +63,12 @@ export async function checkAndRunMigrations() {
...
 			}
 
-			// otherwise all is good and we cna run the migration
-			console.log(`[MIGRATIONS] "${migration.name}" not applied yet. Applying...`);
+			if (migration.runEveryTime) {
+				console.log(`[MIGRATIONS] "${migration.name}" should run every time. Applying...`);
+			} else {
+				// otherwise all is good and we cna run the migration
+				console.log(`[MIGRATIONS] "${migration.name}" not applied yet. Applying...`);
+			}
```
mishig25, line 71:
```suggestion
            // otherwise all is good and we can run the migration
            console.log(`[MIGRATIONS] "${migration.name}" ${migration.runEveryTime ? "should run every time" : "not applied yet"}. Applying...`);
```

src/lib/migrations/routines/index.ts
```diff
@@ -9,6 +10,7 @@ export interface Migration {
 	down?: (client: MongoClient) => Promise<boolean>;
 	runForFreshInstall?: "only" | "never"; // leave unspecified to run for both
 	runForHuggingChat?: "only" | "never"; // leave unspecified to run for both
+	runEveryTime: boolean;
```
mishig25, line 13:
```suggestion
	runEveryTime?: boolean;
```

let's make it optional since it was not present initially

src/lib/migrations/routines/02-update-assistants-models.ts
```diff
@@ -0,0 +1,46 @@
+import type { Migration } from ".";
+import { getCollections } from "$lib/server/database";
+import { ObjectId, type AnyBulkWriteOperation } from "mongodb";
+import type { Assistant } from "$lib/types/Assistant";
+
+const updateAssistantsModels: Migration = {
+	_id: new ObjectId("5f9f3f3f3f3f3f3f3f3f3f3f"),
+	name: "Update deprecated assistants models",
```
mishig25, line 8:
```suggestion
	name: "Update deprecated models in assistants with the default model",
```

src/lib/migrations/routines/02-update-assistants-models.ts
```diff
@@ -0,0 +1,46 @@
...
+		const modelIds = models.map((el) => el.id); // string[]
+		const defaultModelId = models[0].id;
+
+		let ops: AnyBulkWriteOperation<Assistant>[] = [];
+
+		// Find all assistants whose modelId is not in modelIds, and update it to use defaultModelId
+		for await (const assistant of assistants.find({ modelId: { $nin: modelIds } })) {
+			console.log({ assistant });
+			ops.push({
+				updateOne: {
```
mishig25, line 23: instead of updateOne and loop, could we just run a single operation:
```ts
collection.update({
	filter: { modelId: { $nin: modelIds }},
	update: { $set: { modelId: defaultModelId } },
})
```

and we wouldn't need the bulkwrite. Wdyt ?

mishig25: since we have index https://github.com/huggingface/chat-ui/blob/7bec0fbb9a9edd4b97e7700fe5fb5087b2b07cac/src/lib/server/database.ts#L148, this single `update` operation should be fine performance-wise. No?

nsarrazin: yeah that seems better, could it hang in a weird way if we have a lot of assistants? I think it should be safe but not 100% sure

mishig25: cc: @coyotte508

mishig25:
for https://github.com/huggingface/chat-ui/pull/998/commits/92a5fef8cda6b1f205c657bbea9ef8f100705a01 same in https://github.com/huggingface/chat-ui/blob/7bec0fbb9a9edd4b97e7700fe5fb5087b2b07cac/src/lib/components/chat/ChatWindow.svelte#L379

src/lib/migrations/routines/02-update-assistants-models.ts
```diff
@@ -0,0 +1,46 @@
...
+		const { assistants } = getCollections(client);
+
+		const modelIds = models.map((el) => el.id); // string[]
+		const defaultModelId = models[0].id;
+
+		let ops: AnyBulkWriteOperation<Assistant>[] = [];
+
+		// Find all assistants whose modelId is not in modelIds, and update it to use defaultModelId
+		for await (const assistant of assistants.find({ modelId: { $nin: modelIds } })) {
+			console.log({ assistant });
```
mishig25, line 21:
```suggestion
```

before merging

src/lib/migrations/routines/02-update-assistants-models.ts
nsarrazin: i've implemented it, seems a lot cleaner

src/lib/migrations/routines/02-update-assistants-models.ts
```diff
@@ -0,0 +1,46 @@
...
+
+const updateAssistantsModels: Migration = {
+	_id: new ObjectId("5f9f3f3f3f3f3f3f3f3f3f3f"),
+	name: "Update deprecated models in assistants with the default model",
+	up: async (client) => {
+		const models = (await import("$lib/server/models")).models;
+
+		const { assistants } = getCollections(client);
+
+		const modelIds = models.map((el) => el.id); // string[]
```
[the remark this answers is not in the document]
nsarrazin: yes, `model.id` come from the `.env` so it's parsed as a string in zod

mishig25 (approved):
lgtm !

src/lib/migrations/routines/02-update-assistants-models.ts
mishig25: handled in https://github.com/huggingface/chat-ui/pull/998/commits/a63c3d74e029aaada43bd52293f6baf31ea6a913

PR #1002, opened 2024-04-10
Add prompt examples for command-r-plus

nsarrazin:


Arche151:
@nsarrazin Hey Nathan, could you tell me which prompt template I need to set when using Command-R-Plus locally?

nsarrazin:
Hi! For command R+ we get the prompt template from [the tokenizer](https://github.com/huggingface/chat-ui/blob/6b67770a07bb4426deab59e87fe7b67b9b0cc81b/chart/env/prod.yaml#L37). Let me know if that helps!

Arche151:
Ok, I see, thanks for explaining!

I was able to get the model running in ChatUI by following your instructions, but unfortunately the output is not coherent. I am using Command-R+ in German and its German is really bad and riddled with typos in the ChatUI, while being flawless in Jan or LM Studio.

Here's my `.env.local`:

```
MONGODB_URL=mongodb://localhost:27017
HF_TOKEN=mytoken
MODELS=`[
  {
    "name": "Command-R+",
    "displayName": "Command-R+",
    "description": "Command-R+ very nice",
        "tokenizer": "Xenova/c4ai-command-r-v01-tokenizer",
    "parameters": {
      "temperature": 0.3,
      "top_p": 0.95,
      "repetition_penalty": 1.2,
      "top_k": 50,
      "truncate": 32768,
      "max_new_tokens": 1024,
      "stop": ["<endofstring>"]
    },
    "endpoints": [
      {
        "url": "http://localhost:8080",
        "type": "llamacpp"
      }
    ]
  }
]`

ENABLE_ASSISTANTS=true 

```
Do you see any mistakes in my configuration? Also, if possible, I would prefer a completely offline setup, that doesn't require my hf_token. Is there any chance, that the Chat UI prompt template for Command-R+ could be added to PROMPTS.md?

And could you maybe also help me with my wish to use WizardLM-2-8x22B with the ChatUI? I would really appreciate that :)

PS: Sorry for spamming you with my questions!

PR #1004, opened 2024-04-11
Fix unresponsive page after deleting all conversations

### LINKED ISSUE #937: After deleting all your conversations, all buttons become unusable and the page needs to be refreshed to return to normal.
image

after check this button, go to the new page, no operations can be performed.

francoislanc:
The "on:close" event function is not triggered by the dispatch("close") before the page redirection (the modal is not properly removed and the page becomes unresponsive).

I removed the dispatch("close") and directly apply what is done by the "on:close" event function.

I tested it and now the page is responsive after deletion of all the conversations. It should fix #937

### DIFF
--- a/src/routes/settings/(nav)/+page.svelte
+++ b/src/routes/settings/(nav)/+page.svelte
@@ -1,6 +1,4 @@
 <script lang="ts">
-	import { createEventDispatcher } from "svelte";
-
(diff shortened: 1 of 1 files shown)


nsarrazin (approved):
Works well in local testing! Thanks for contributing a fix 🚀

Issue #633, opened 2023-12-13
Feature request: add metrics exporter for backend


itaybar:
I think it would be great if there was any stats/metrics exporter (can be prometheus exporter/ open telemetry) to debug and monitor chatui errors and latencies

PR #1013, opened 2024-04-14
Add support for backend telemetry (tracing & metrics)

### LINKED ISSUE #692: feature request: add Open Telemetry (OTEL)/Prometheus exporter in chatui
Adding /metrics route which will show metrics like how many users has logged (since service started), how many requests, model distribution, failures (with status code, or status describe), etc.

I think This is important since Svelte Is the backend and the frontend this code be a very important feature to allow debugging and monitoring

zacps:
This uses the opentelemetry SDK to provide support for traces and two simple metrics (a conversations count and a messages count).

The telemetry entrypoint is built separately (unfortunately it does not seem like there is a method for doing this in one build as sveltekit overrides rollupOptions).

As written the metrics may be too high cardinality for some use cases (as they are keyed per user), but these attributes can always be dropped in a collector.

Fixes #692, #633


vite.telemetry.config.ts
```diff
@@ -0,0 +1,19 @@
+import { defineConfig } from "vite";
+
+export default defineConfig({
```
zacps, line 3: We unfortunately need a second build step because sveltekit overrides vite's `rollupOptions` which we need to specify a second entrypoint

scripts/clean-opentelemetry.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
zacps, line 1: Workaround for https://github.com/open-telemetry/opentelemetry-js/issues/3989

antonioramos1:
Very happy to see this contribution! I was also missing having better observability since we currently have little visibility on how many users we have and how they are using chat-ui

nsarrazin:
Hey so I was working on adding support for prometheus metrics because we need them for HuggingChat but I think we can probably do it with this PR! Thanks for the contribution. 🚀 

I'd like to have a `/metrics` route on a different port which exports metrics in a format compatible with prometheus. I think that's doable with opentelemetry, will dive into this

nsarrazin:
The type checking error seems to be reported here https://github.com/open-telemetry/opentelemetry-js/issues/3944

zacps:
> Hey so I was working on adding support for prometheus metrics because we need them for HuggingChat but I think we can probably do it with this PR!

It looks like this is possible via the [prometheus exporter](https://www.npmjs.com/package/@opentelemetry/exporter-prometheus). If we add this I would like to keep the opentelemetry exporter as an option.

```mermaid
flowchart LR
    c["chat-ui"]-->|Prometheus|m["/metrics"]
```

Alternatively we could only support otlp exports and anyone who wanted a prometheus endpoint could get it through the opentelemetry-collector:

```mermaid
flowchart LR
    chat-ui --> |OTLP|opentelemetry-collector-->|Prometheus|/metrics
    opentelemetry-collector-->|prometheus-remote-write|endpoint
    
```

In my experience with opentelemetry so far using a collector has been pretty useful to add/transform labels, drop healthcheck data, etc

natikgadzhi:
I like this! I know @nsarrazin alteady has metrics merged, but I was thinking about doing otel telemetry, and I wonder if this is salvageable to keep Prometheus metrics endpoint as it is via exporter and get traces too. I'll take a look next week.

PR #1015, opened 2024-04-15
Update marked from v4 to v12

Saghen:
The current version of marked (v4) momentarily renders the following as a header instead of as a paragraph + list
```
Foo bar:
- Hello world
-
```
Any text added after the `-` returns the rendered markdown back to a list. This leads to flashes of incorrect markdown, which becomes especially noticeable with longer lists (tested with Claude Haiku). The latest version of marked (v12) fixes the issue. The types have been moved into the marked package, so @types/marked was removed. The PR also updates marked-katex-extension to v5 which supports marked v12.


nsarrazin (approved):
:wave: Thanks for the contrib! Tested it locally, seems to work pretty well! It still shows a header with the following text:

```
This is a header:
 - 
 ```
Screenshot from 2024-04-16 08-45-30

but I guess that's just normal header definition in markdown and we can't do much about it? Regardless, thanks for updating the (quite outdated) deps :rocket:

PR #1018, opened 2024-04-16
Privacy update

julien-c:
I've been reluctant to do this in the past because it's a one-way change (we can probably never start sharing conversation data with model authors again in the future from this app),

but internal discussion brought this subject again and i think it's as good a change as any

### Screenshot of privacy page 👀

Screenshot 2024-04-16 at 15 16 26

### DIFF
--- a/PRIVACY.md
+++ b/PRIVACY.md
@@ -4,7 +4,7 @@
 
 Users of HuggingChat are authenticated through their HF user account.
 
-We endorse Privacy by Design. As such, your conversations are private to you and will not shared with anyone, including model authors, and including for research or model training purposes.
+We endorse Privacy by Design. As such, your conversations are private to you and will not be shared with anyone, including model authors, for any purpose.
 
 You conversation data will only be stored to let you access past conversations. You can click on the Delete icon to delete any past conversation at any moment.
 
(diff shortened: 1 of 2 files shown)


nsarrazin (approved):

PRIVACY.md
```diff
@@ -1,25 +1,31 @@
 ## Privacy
 
-> Last updated: October 4, 2023
+> Last updated: April 15, 2024
 
 Users of HuggingChat are authenticated through their HF user account.
 
-By default, your conversations may be shared with the respective models' authors to improve their training data and model over time. Model authors are the custodians of the data collected by their model, even if it's hosted on our platform.
+We endorse Privacy by Design. As such, your conversations are private to you and will not shared with anyone, including model authors, and including for research or model training purposes.
```
Pierrci, line 7:
```suggestion
We endorse Privacy by Design. As such, your conversations are private to you and will not be shared with anyone, including model authors, for any purpose.
```
a bit clearer?

julien-c: i'll tweak but merging your suggestion

julien-c: 6bb7e17b461fda5dbfa80d49ab1308f58e483142

Pierrci (approved):

PR #1020, opened 2024-04-16
add an endpoint to api to get an assistant by id

alak:
add an endpoint /chat/api/assistant/[id] to get an assistant by id.

### DIFF
--- a/src/routes/api/assistant/[id]/+server.ts
+++ b/src/routes/api/assistant/[id]/+server.ts
@@ -1,0 +1,17 @@
+import { collections } from "$lib/server/database";
+import { ObjectId } from "mongodb";
+
+export async function GET({ params }) {
+	const id = params.id;
+	const assistantId = new ObjectId(id);
+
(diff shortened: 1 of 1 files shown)


src/routes/api/assistant/[id]/+server.ts
```diff
@@ -0,0 +1,20 @@
+import { collections } from "$lib/server/database";
+import { authCondition } from "$lib/server/auth";
+import { z } from "zod";
+import { ObjectId } from "mongodb";
+
+export async function GET({ locals, params }) {
+	const id = z.string().parse(params.id);
```
Comment (coyotte508) [inline src/routes/api/assistant/[id]/+server.ts:7]: I don't think this is needed, unless you use a regex like `z.string().regex(/.../)`. Otherwise you can directly use `params.id`

coyotte508 (approved):

nsarrazin (approved):

nsarrazin:
feel free to merge when ready and I'll deploy :rocket:

PR #1041, opened 2024-04-21
Support Gemini 1.5 Pro from Vertex AI

flexchar:
Hey crew!
I went down the rabbit hole wanting to have Gemini work. I actually like it quite a lot. Decent work by Google. 😩

The previous implementation would bug down after one or two words. The prompt building was completely unnecessary. I used inspiration from my side experience and implementation for OpenAI endpoint to use native methods offered by Google's SDK. 

We can now use system template, disable safety filtering (*requires approval by google), and no more "continue" button.  Basically have a good experience.

I also updated confusing README. I had to look at the implementation to get it working, not sure how that slipped through.

Not to be sneaky, I also added git ignore for service accounts needed for Vertex AI and committed a small fix for model descriptions.

Cheers :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ SECRET_CONFIG
 !.env.ci
 !.env
 !.env.template
+gcp-*.json
--- a/README.md
+++ b/README.md
@@ -607,12 +607,17 @@ Make sure docker has access to the file. Afterwards Google Vertex endpoints can 
 MODELS=`[
 //...
     {
-       "name": "gemini-1.0-pro", //model-name
-       "displayName": "Vertex Gemini Pro 1.0",
-       "location": "europe-west3",
-       "apiEndpoint": "", //alternative api endpoint url
+       "name": "gemini-1.5-pro",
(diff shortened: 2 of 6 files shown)


ruru-m07 (approved):

nsarrazin (approved):
Hey! 👋 Thanks for the great PR. I fixed the last type issues, if you can confirm that the PR still works well for you then I'll merge it 😁

flexchar:
It still works but the tokens are not displayed until the stream ends. I implemented this on the older _HEAD_ and then made PR based on the latest `main`. I cannot figure out right now why it doesn't work so we will have to postpone a bit...

flexchar:
Yes! That was it. Also seen in this file on line 27 at the ternary check: `src/lib/server/endpoints/openai/openAIChatToTextGenerationStream.ts`.

julien-c (commented):
wow, very solid PR @flexchar!

nsarrazin:
Thanks for the PR! I'll be merging this now 🚀

PR #1043, opened 2024-04-21
Move featured assistant check behind `REQUIRE_FEATURED_ASSISTANTS`

### LINKED ISSUE #997: Community Assistants
Hi, I've looked through all the possible issues but I didn't find what I was looking for. 

On self-hosted is the option to have the community assistants such as the ones on https://huggingface.co/chat/ not available? I've also noticed that when I create Assistants on my side they do not show up on community tabs either they are purely user restricted, I am missing something? I've configured the hf token and the API base, any hints are appreciated.

image

nsarrazin:
Closes #997

With this PR, to get the current prod behaviour you need `REQUIRE_FEATURED_ASSISTANTS=true` which we set for HuggingChat.

Otherwise for all other deployments, assistants are shown in the community page by default


julien-c:
makes sense, IMO!

PR #1050, opened 2024-04-23
Removed mistralai/Mistral-7B-Instruct-v0.1 from HuggingChat

KingNishHF:
The mistralai/Mistral-7B-Instruct-v0.1 was removed due to it being unnecessary. Since it was eliminated from the task model, there was no further use for that Model, leading to its removal.


nsarrazin:
Hey will close this since we merged #1052 but thanks for the suggestion 🤗

PR #1053, opened 2024-04-23
fix: See all your assistants regardless of usercount

nsarrazin:


KingNishHF:
Still, not working for all assistant only assistant have 50+ views are visible.

PR #1055, opened 2024-04-23
Added MTSAIR/MultiVerse_70B

KingNishHF:
This model outperforms Command R+ , Llama 3 70B and many more, on open llm leaderboard. This model also a great alternative as command R+.

License - https://huggingface.co/MTSAIR/MultiVerse_70B/discussions/7#66278c8e430a12425331b183

Model Link - https://huggingface.co/MTSAIR/MultiVerse_70B

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -28,6 +28,33 @@ MODELS=`[
     ]
   },
   {
+    "name" : "MTSAIR/MultiVerse_70B",
+    "tokenizer": "MTSAIR/MultiVerse_70B",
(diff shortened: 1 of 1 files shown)


nsarrazin:
Hi! If you want to suggest a model to be used on HuggingChat, it's best if you do so [here](https://huggingface.co/spaces/huggingchat/chat-ui/discussions/372). 

We take community feedback into account when hosting new models but it's not the only factor, deploying a model requires some backend work on our side that's not visible in this repo, so it's best to just have a discussion on the hub first. 🤗

Issue #1072, opened 2024-04-25
[v0.8.3] serper, serpstack API, local web search not working


adhishthite:
## Context

I have serper.dev API key, serpstack API key and I have put it correctly in my `.env.local` file.

<img width="478" alt="image" src="https://github.com/huggingface/chat-ui/assets/31769894/5082893a-7ecd-4ab5-9cb9-059875118dcd">

## Issue

However, even if I enable Web Search, it still does not reach out to those APIs, and shows me "an error occured" no the Web Search part.

<img width="931" alt="image" src="https://github.com/huggingface/chat-ui/assets/31769894/da96c121-89e0-402b-8e93-33c9e6709c71">

I don't see calls reaching Serper and SerpStack as well.

<img width="1365" alt="image" src="https://github.com/huggingface/chat-ui/assets/31769894/7230b1a0-2567-424f-8884-8fc53417fa41">

<img width="1302" alt="image" src="https://github.com/huggingface/chat-ui/assets/31769894/b35c1a7f-1c2c-4c8a-9c46-5c2171f73f9b">

It was working for a bit on `v0.8.2`, but then it stopped working there as well. Now, for `v.0.8.3`, it's not working at all. Am I missing something? I have tried using either of those APIs too, but it still does not work.

Please help.

nsarrazin:
Oh `USE_LOCAL_WEBSEARCH` will attempt to search the web by fetching google search directly and parsing the resulting page.  It will override other websearch APIs. It's a bit more unstable than using an API, try removing this flag and let me know if that works. 

Also we don't support using multiple search queries at the same time so if you set both serpapi and serper, the priority will be given to serper with our current setup.

adhishthite:
@nsarrazin 

Thanks! I disabled the local web search and kept only Serper.

I see, but here's what's happening now. Even though Serper received the query, the UI shows an error.

<img width="936" alt="image" src="https://github.com/huggingface/chat-ui/assets/31769894/e8592cff-1b9a-42b9-b479-1fdfa5938bba">

<img width="1273" alt="image" src="https://github.com/huggingface/chat-ui/assets/31769894/4e8f57c1-bbf3-437b-bd80-c594b1ffa71e">

Is there something I am missing?

adhishthite:
@nsarrazin  here you go

<img width="761" alt="image" src="https://github.com/huggingface/chat-ui/assets/31769894/d1b59da9-fcd1-4213-a85a-743cad658256">

[...]

PR #1077, opened 2024-04-25
Fix OpenAI Embeddings Auth Issue

adhishthite:
Fixing this issue: https://github.com/huggingface/chat-ui/issues/1072

OpenAI depends on `Api-Key` instead of `Authorization`


nsarrazin:
Hey thanks for contributing a PR! I'm a bit surprised, looking at the [OpenAI doc](https://platform.openai.com/docs/api-reference/authentication) it looks like they use `Authorization`... Could it be an azure specific thing ?

In which case, I think the easiest thing would be to follow the [OpenAI endpoint for text generation](https://github.com/huggingface/chat-ui/blob/745e51e28bab8c380564efb32056a3508ee94f11/src/lib/server/endpoints/openai/endpointOai.ts#L17-L19) and add options for `defaultHeaders` and `defaultQuery` in the embedding endpoint.

Let me know if that sounds like something you can/want to do, otherwise I'll open a PR :smile:

adhishthite:
@nsarrazin Thanks! I will do this. Give me a day. It's almost 3:00 AM here, I'll do it once I resume work tomorrow.

adhishthite:
This seems like an Azure only thing.

adhishthite:
@nsarrazin 

How about I create a new `type` altogether for Azure OpenAI? Since the Auth and the URL type is different, (and might change in the future), having a new `azure-openai` type for embeddings will be much easier to modify and manage.

Please lmk what you think.

nsarrazin:
Hey @adhishthite I'd be more in favour of just adding a `defaultHeaders` option like [here](https://github.com/huggingface/chat-ui/blob/745e51e28bab8c380564efb32056a3508ee94f11/src/lib/server/endpoints/openai/endpointOai.ts#L17)

Then you can do

```
					headers: {
						Accept: "application/json",
						"Content-Type": "application/json",
						...(apiKey ? { Authorization: `Bearer ${apiKey}` } : {}),
                                                ...defaultHeaders
                                                }
```

something like this, which should be enough for your auth needs I think :grin:

nsarrazin:
I pushed a commit to your branch, where I added the `defaultHeaders` option in the endpoint.

You should be able to make it work with azure by adding 
```
"defaultHeaders" : {
    "Api-Key" : "your api key"
}
```

to your embedding endpoint config. Let me know if that works!

nsarrazin:
going to merge this, let me know if it needs something else!

nsarrazin (approved):

PR #1082, opened 2024-04-27
Sort by Trending to default

KingNishHF:
Switched the default sorting from 'popular' to 'trending'.

WHY?? : Because everything on hugging face as default sort by trending


nsarrazin:
cc @mishig25 @gary149 this makes sense to me, I also think we'll see more variety in assistants if trending is shown by default

gary149 (approved):
ok for me

nsarrazin (approved):
Committed a change to apply the change in the load function too, seems to work now!

nsarrazin:
Thanks for the contribution! :rocket:

KingNishHF:
> Thanks for the contribution! 🚀

Currently it's not working.
I think something more needed to be change on because default not changed.

PR #1083, opened 2024-04-29
Sort By trending

KingNishHF:
Previous not working - https://github.com/huggingface/chat-ui/pull/1082

I read about this and find that HTML select which was set first so
@nsarrazin


nsarrazin:
Hey, the change wasn't deployed properly it seems but it should be working in prod now.  Let me know if it also works for you and then I think we can close this 😁

KingNishHF:
> Hey, the change wasn't deployed properly it seems but it should be working in prod now. Let me know if it also works for you and then I think we can close this 😁

ookk 😅

PR #1085, opened 2024-04-29
Move vars to dynamic, add metrics

nsarrazin:
This PR moves all env vars to $env/dynamic

We add prometheus metrics

Get rid of `Dockerfile.local` file

### Docker builds
* chat-ui `docker build -t chat-ui .`
* chat-ui-db `docker build -t chat-ui-db --build-arg="INCLUDE_DB=true" . `
* hchat  `docker build -t huggingchat --build-arg="APP_BASE=/chat" --build-arg="PUBLIC_APP_COLOR=yellow" .`


src/routes/healthcheck/+server.ts
```diff
@@ -0,0 +1,3 @@
+export async function GET() {
+	return new Response("OK", { status: 200 });
```
rtrompier, line 2: It could be interesting to check the DB connectivity here and other dependencies (if there are)

rtrompier (approved):

src/routes/metrics/+server.ts
```diff
@@ -0,0 +1,9 @@
+import { register } from "$lib/server/metrics";
+
+export async function GET() {
```
rtrompier, line 3: Can we expose this route on another port? 
It would be better if we can avoid exposing metrics publicly

rtrompier (approved):

XciD:
Good start for metrics, we should check if we can create some specific ones like:
number of api-inference failures per model for example

.dockerignore
```diff
@@ -8,4 +8,4 @@ node_modules/
 .svelte-kit/
 .env*
 !.env
-!.env.local
\ No newline at end of file
+.env.local
```
rodrigobdz, line 11: @nsarrazin What is the recommended way to use `.env.local` with Docker? 

## Context
The [readme](https://github.com/huggingface/chat-ui/blob/93b39a0beb72378c76d5d146bfd3a8355c1d110d/README.md) is currently optimised for local envs and requires changes for Docker envs. 

## Issue
I stumbled upon the issue that the `.env.local` was ignored when building the Docker image in CI because of this ignore statement. 

## Possible Solutions
A temporary solution when using Docker Compose is available in PR https://github.com/huggingface/chat-ui/pull/1238, but it doesn't apply to Kubernetes deployments. Reverting this change—i.e., to `!.env.local`—would work for both Docker Compose and Kubernetes.

nsarrazin: I've added a few examples[here](https://github.com/huggingface/chat-ui/wiki/Deployment-examples#run-as-a-docker-image) but i'll be improving the docs for this.

rodrigobdz: Furthermore, including `.env.local` in `Dockerfile` requires this change also:

```diff
+ COPY --chown=1000 .env.local /app/.env.local
```

rodrigobdz: I faced several issues when deploying this to a Kubernetes cluster. I ended up passing somehow the `.envlocal` into the Docker image.

PR #1094, opened 2024-05-01
Web Search: Playwright, spatial parsing, markdown

Saghen:
### Context

The existing web search implementation naively scrapes and chunks content for passing to the LLM. A brief explanation of how it works:
- **Scraping**: Fetch the page's HTML, statically parse it with selectors
	- `document.querySelectorAll('p, table, pre, ul, ol')`
- **Chunking**: Concatenate the text of all elements with spaces, chunk the resulting text so that the length of each chunk is less than the maximum embedding length
- **Embedding**: Get sentence similarity for each chunk, pass top 8 chunks to the LLM

### Solution

- **Scraping**: Load the page into Playwright and perform spatial parsing
  - The spatial parser uses a clustering technique based on the position to find the primary content. So i.e., you might end up with a cluster for the header, footer, primary content and sidebar. Heuristics, such as text density, find the critical cluster which should contain the primary content
  - Metadata scraping: title, description, site name, author, updated at, created at
- **Conversion to markdown/chunking:**
  - Convert the resulting list of HTML elements into a tree like `h1 [h2 [p p blockquote] h2 [h3 [...] ] ]`
  - Convert the HTML elements into their markdown equivalents
- **Chunking:** Treat each markdown element as a chunk. Split elements where `element.text.length > embeddingMaxLength` based on sentence boundaries
- **Embedding:**
  - Get sentence similiarty for each markdown element
  - Get top 8 chunks and their surrounding context (based on the tree from conversion to markdown)
  - For each website, include the scraped metadata before the chunks

~40% reduction in total content
~30% speed up when doing CPU embedding
~1s increased web search time on average when using GPU embedding


mishig25:
for the ones who are testing, make sure to run
```
npm ci
npx playwright install
```
to get the necessary deps before running

Aaditya-Sahay:
@Saghen Since we are using playwright as a library, we should add ``@playwright/browser-chromium`` to our list of dependencies so it automatically installs for people . See [here](https://playwright.dev/docs/library#key-differences)

Saghen:
For nix users, pin the playwright version to the latest version in nixpkgs (currently 1.40.0) via `npm i playwright@1.40.0`. Then launch a nix shell with the following config:

```nix
{ pkgs ? import <nixpkgs> { } }:
pkgs.mkShell {
  nativeBuildInputs = with pkgs; [ playwright-driver.browsers ];

  shellHook = ''
    export PLAYWRIGHT_BROWSERS_PATH=${pkgs.playwright-driver.browsers}
    export PLAYWRIGHT_SKIP_VALIDATE_HOST_REQUIREMENTS=true
  '';
}
```

gary149:
Let's go with JS disabled by default :)

README.md
```diff
@@ -170,6 +170,8 @@ You can enable the web search through an API by adding `YDC_API_KEY` ([docs.you.
 
 You can also simply enable the local google websearch by setting `USE_LOCAL_WEBSEARCH=true` in your `.env.local` or specify a SearXNG instance by adding the query URL to `SEARXNG_QUERY_URL`.
```
mishig25, line 172: in the websearch readme, should we add somewhere that `npx playwright install` is needed to run the websearch?

src/lib/server/isURLLocal.ts
```diff
@@ -1,26 +1,48 @@
...
+		const urlObj = new URL(url);
+		return isURLLocal(urlObj);
+	} catch (e) {
+		// assume local if URL parsing fails
+		return true;
+	}
+}
 
-	return isLocal;
+// TODO: move this to a generic url helper
```
mishig25, line 40: handle the todo

src/lib/server/preprocessMessages.ts
```diff
@@ -13,11 +13,9 @@ export async function preprocessMessages(
 	return await Promise.all(
 		structuredClone(messages).map(async (message, idx) => {
 			const webSearchContext = webSearch?.contextSources
-				.map(({ context }) => context)
-				.flat()
-				.sort((a, b) => a.idx - b.idx)
-				.map(({ text }) => text)
-				.join(" ");
+				.map(({ context }) => context.trim())
+				.join("\n\n----------\n\n");
```
mishig25, line 17: why is `.join("\n\n----------\n\n")` used over `.join(" ");` ?

src/lib/server/websearch/scrape/types.ts
```diff
@@ -0,0 +1,5 @@
+export type SerializedHTMLElement = {
```
mishig25, line 1: nit. stylisitically, in this case, we have been using `interface` over `type`

src/lib/server/websearch/scrape/playwright.ts
```diff
@@ -0,0 +1,57 @@
...
+	};
+	const ctx = await browser.newContext(options);
+	const blocker = await PlaywrightBlocker.fromPrebuiltAdsAndTracking(fetch).then((blker) => {
+		const mostBlocked = blker.blockFonts().blockMedias().blockFrames().blockImages();
+		if (env.WEBSEARCH_DISABLE_JAVASCRIPT) return mostBlocked.blockScripts();
+		return mostBlocked;
+	});
+	return Object.freeze({ ctx, blocker });
+}
+let playwrightService: Promise<{ ctx: BrowserContext; blocker: PlaywrightBlocker }>;
```
mishig25, line 43: mybe put this line before `initPlaywrightService` function for the clarity and add a comment that it is a singleton

src/lib/server/websearch/markdown/types.ts
```diff
@@ -0,0 +1,57 @@
...
+	BlockQuote = "BLOCKQUOTE",
+	CodeBlock = "CODE_BLOCK",
+
+	UnorderedList = "UNORDERED_LIST",
+	OrderedList = "ORDERED_LIST",
+	UnorderedListItem = "UNORDERED_LIST_ITEM",
+	OrderedListItem = "ORDERED_LIST_ITEM",
+}
+
+type BaseMarkdownElement = {
```
mishig25, line 14: can it be generic like `BaseMarkdownElement<T = MarkdownElementType>` and then instead of `type ParagraphElement = BaseMarkdownElement & { type: MarkdownElementType.Paragraph };`, we can use it like `type ParagraphElement = BaseMarkdownElement<MarkdownElementType.Paragraph>`

src/lib/server/websearch/markdown/fromHtml.ts
```diff
@@ -0,0 +1,98 @@
+import { collapseString, sanitizeString } from "./utils/nlp";
+import { stringifyHTMLElements, stringifyHTMLElementsUnformatted } from "./utils/stringify";
+import { MarkdownElementType, tagNameMap, type HeaderElement, type MarkdownElement } from "./types";
+import type { SerializedHTMLElement } from "../scrape/types";
+
+type ConversionState = {
```
mishig25, line 6: same comment as https://github.com/huggingface/chat-ui/pull/1094#discussion_r1596919054

src/lib/server/websearch/markdown/utils/stringify.ts
```diff
@@ -0,0 +1,72 @@
...
+
+/** Grabs all text content directly, ignoring HTML tags */
+export function stringifyHTMLElementsUnformatted(elem: (SerializedHTMLElement | string)[]): string {
+	return elem.map(stringifyHTMLElementUnformatted).join("");
+}
+
+/** Grabs all text content directly, ignoring HTML tags */
+function stringifyHTMLElementUnformatted(elem: SerializedHTMLElement | string): string {
+	if (typeof elem === "string") return elem;
+	return elem.content.map(stringifyHTMLElementUnformatted).join("");
```
mishig25, line 71: is joining with `""` rather than `" "` (a space) right move here ?

Saghen: yeah because this is most often used in cases where we're combining text nodes together which already include the spacing

src/lib/server/preprocessMessages.ts
Saghen: The content is formatted in markdown so I thought it might be helpful for the model to know when it's receiving a different source but not sure if it's helping that much

README.md
Saghen: That shouldn't be needed anymore since I added `@playwright/browser-chromium` to the `package.json` but I can't test that it works on my system

.env
```diff
@@ -27,6 +27,7 @@ SEARXNG_QUERY_URL=# where '<query>' will be replaced with query keywords see htt
 
 WEBSEARCH_ALLOWLIST=`[]` # if it's defined, allow websites from only this list.
 WEBSEARCH_BLOCKLIST=`[]` # if it's defined, block websites from this list.
+WEBSEARCH_JAVASCRIPT=false # CPU usage reduces by 60% on average by disabling javascript. Enable to improve website compatibility
```
mishig25, line 30: btw, is this variable still used? Couldn't find its usage from a search

src/lib/types/WebSearch.ts
```diff
@@ -14,33 +15,31 @@ export interface WebSearch extends Timestamps {
...
+export interface WebSearchScrapedSource extends WebSearchSource {
+	page: WebSearchPage;
+}
+export interface WebSearchPage {
+	title: string;
+	siteName?: string;
+	author?: string;
+	description?: string;
+	createdAt?: string;
+	modifiedAt?: string;
```
mishig25, line 30: they should be of type: `Date`

mishig25: actually, you can make WebSearchPage extends

https://github.com/huggingface/chat-ui/blob/492c51fff022ba40410056c0889b974f14a83471/src/lib/types/Timestamps.ts#L1-L4

and delete fields `createdAt` & `modifiedAt`

src/lib/server/websearch/markdown/utils/stringify.ts
```diff
@@ -0,0 +1,72 @@
...
+// --- Markdown Elements ---
+
+/** Converts markdown element to a string with formatting */
+export function stringifyMarkdownElement(elem: MarkdownElement): string {
+	const content = elem.content.trim();
+	if (elem.type === MarkdownElementType.Header) return `${"#".repeat(elem.level)} ${content}\n\n`;
+	if (elem.type === MarkdownElementType.BlockQuote) {
+		return `${"> ".repeat(elem.depth)}${content}\n\n`;
+	}
+	if (elem.type === MarkdownElementType.CodeBlock) return `\`\`\`\n${content}\n\`\`\`\n\n`;
```
mishig25, line 13:
```suggestion
	if (elem.type === MarkdownElementType.CodeBlock) return `\`\`\`\${elem.codeLanguage}n${content}\n\`\`\`\n\n`;
```

nit, it is possible to get the code language information like this?

src/lib/server/websearch/markdown/utils/stringify.ts
```diff
@@ -0,0 +1,72 @@
...
+		return `${order}. ${content}\n`;
+	}
+
+	return `${content}\n\n`;
+}
+
+// ----- HTML Elements -----
+
+/** Ignores all non-inline tag types and grabs their text. Converts inline tags to markdown */
+export function stringifyHTMLElements(elem: (SerializedHTMLElement | string)[]): string {
```
mishig25, line 32:
```suggestion
export function stringifyHTMLElements(elems: (SerializedHTMLElement | string)[]): string {
```

`elems` rather than `elem` would be a better argument name since it is an array

src/lib/server/websearch/markdown/utils/stringify.ts
```diff
@@ -0,0 +1,72 @@
...
+		const href = elem.attributes.href;
+		if (!href) return elem.content.map(stringifyHTMLElement).join("");
+		return `[${elem.content.map(stringifyHTMLElement).join("")}](${href})`;
+	}
+
+	return elem.content.map(stringifyHTMLElement).join("");
+}
+
+/** Grabs all text content directly, ignoring HTML tags */
+export function stringifyHTMLElementsUnformatted(elem: (SerializedHTMLElement | string)[]): string {
```
mishig25, line 64: same comment as https://github.com/huggingface/chat-ui/pull/1094#discussion_r1598099428

src/lib/server/websearch/markdown/types.ts
```diff
@@ -0,0 +1,55 @@
...
+	UnorderedList = "UNORDERED_LIST",
+	OrderedList = "ORDERED_LIST",
+	UnorderedListItem = "UNORDERED_LIST_ITEM",
+	OrderedListItem = "ORDERED_LIST_ITEM",
+}
+
+interface BaseMarkdownElement<T = MarkdownElementType> {
+	type: T;
+	content: string;
+	parent: HeaderElement | null;
```
mishig25, line 17:
```suggestion
	parent?: HeaderElement;
```

we can make `parent` optional unless having an actual `NULL` value is preferred

src/lib/server/websearch/markdown/tree.ts
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Converts HTML elements to Markdown elements and creates a tree based on header tags
+ * For example: h1 [h2 [p p blockquote] h2 [h3 [...] ] ]
+ **/
+export function htmlToMarkdownTree(
+	title: string,
+	htmlElements: SerializedHTMLElement[],
+	maxCharsPerElem: number
+): HeaderElement {
+	let parent: HeaderElement = {
```
mishig25, line 16:
```suggestion
	let root: HeaderElement = {
```
I think `root` could be a better var name here

src/lib/server/websearch/markdown/fromHtml.ts
```diff
@@ -0,0 +1,98 @@
...
+import type { SerializedHTMLElement } from "../scrape/types";
+
+interface ConversionState {
+	defaultType:
+		| MarkdownElementType.Paragraph
+		| MarkdownElementType.BlockQuote
+		| MarkdownElementType.UnorderedListItem
+		| MarkdownElementType.OrderedListItem;
+	listDepth: number;
+	blockQuoteDepth: number;
```
mishig25, line 13:
```suggestion
	depth?: number; // to be used for list depth or blockquote depth
```

wdyr about this typing?

src/lib/utils/url.ts
```diff
@@ -0,0 +1,8 @@
+export function isURL(url: string) {
```
mishig25, line 1: nit: the filename can be `isURL.ts`

src/lib/server/websearch/scrape/types.ts
mishig25: I think this should be `interface` & was left from https://github.com/huggingface/chat-ui/pull/1094/commits/492c51fff022ba40410056c0889b974f14a83471

src/lib/server/websearch/scrape/parser.ts
```diff
@@ -0,0 +1,552 @@
+import type { SerializedHTMLElement } from "./types";
+
+type DBSCANOptions<T> = {
```
mishig25, line 3: same comment as https://github.com/huggingface/chat-ui/pull/1094#discussion_r1598141472

src/lib/server/websearch/scrape/playwright.ts
```diff
@@ -0,0 +1,58 @@
...
+	devices,
+	type Page,
+	type BrowserContextOptions,
+} from "playwright";
+import { PlaywrightBlocker } from "@cliqz/adblocker-playwright";
+import { env } from "$env/dynamic/private";
+
+// Singleton initialized by initPlaywrightService
+let playwrightService: Promise<{ ctx: BrowserContext; blocker: PlaywrightBlocker }>;
+async function initPlaywrightService() {
```
mishig25, line 13:
```suggestion
let playwrightService: Promise<{ ctx: BrowserContext; blocker: PlaywrightBlocker }>;

async function initPlaywrightService() {
```

src/lib/server/websearch/scrape/scrape.ts
```diff
@@ -0,0 +1,37 @@
...
+	);
+	return Promise.race([timeout, page.evaluate(spatialParser)])
+		.then(({ elements, ...parsed }) => ({
+			...parsed,
+			markdownTree: htmlToMarkdownTree(parsed.title, elements, maxCharsPerElem),
+		}))
+		.catch((cause) => {
+			throw Error("Parsing failed", { cause });
+		})
+		.finally(() => page.close());
```
mishig25, line 36: should we `clearTimeout` in the `finally` ?

src/lib/types/WebSearch.ts
Saghen: They're actually strings because we don't know what format the dates will be in when we parse them from the page. For now we just grab the value in the meta elems

https://github.com/huggingface/chat-ui/pull/1094/files#diff-d4fd570f6480c02bb93d280a06da3a741b497b41c9878984197a3f9ec005030eR522-R528

src/lib/server/websearch/markdown/utils/stringify.ts
Saghen: eventually yes! but for now, there's no easy way to detect it that I know of. It'll likely involving looking at the class names on the pre/code elems

src/lib/server/websearch/markdown/types.ts
Saghen: yeah I preferred `null` here because I felt it was more explicit in the lack of a parent than `undefined` which may indicate to a reader that the parent is optional

src/lib/server/websearch/markdown/fromHtml.ts
Saghen: It's possible to have lists inside of block quotes so it's good to keep these separate

.env
Saghen: wait a minute, it was named wrong in the code :joy: fixed

src/lib/server/websearch/scrape/scrape.ts
Saghen: switched it to use the timeout util

mishig25 (approved):
lgtm! Great work 🔥

PR #1095, opened 2024-05-02
Ensures more consistent hyperparameters to Phi-3-mini-4k-instruct

gugarosa:
We are still trying to get to the root cause of the model de-railing after long conversations. We haven't been able to reproduce it locally (through scripts or chat-ui), so we are tackling some additional possibilities here.

### DIFF
--- a/chart/env/prod.yaml
+++ b/chart/env/prod.yaml
@@ -242,6 +242,7 @@ envVars:
           "chatPromptTemplate": "<s>{{preprompt}}{{#each messages}}{{#ifUser}}<|user|>\n{{content}}<|end|>\n<|assistant|>\n{{/ifUser}}{{#ifAssistant}}{{content}}<|end|>\n{{/ifAssistant}}{{/each}}",
           "parameters": {
             "stop": ["<|end|>", "<|endoftext|>", "<|assistant|>"],
+            "temperature": 0.7,
             "max_new_tokens": 1024,
             "truncate": 3071
           },


nsarrazin (commented):
Hey! Thanks for tackling this. I tried those changes locally but still get the same issue. 😅  (When hitting the prod inference API)

I'm going to check internally if it could be something on our API since you can't reproduce the issue with TGI.

(Also if you remove `preprompt` it means we can't use the model for assistants, but preprompt should be empty by default, so I think it should be fine to keep)

gugarosa:
Thanks @nsarrazin!

I am also suspecting that it could be related to the API because I am also able to produce the repetitive/gibberish output on the model's page widget. However, it is unclear to me what it could be 😢

nsarrazin:
I'll be asking internally to figure out what could be the issue. 🤔 In your tests, what inference solution were you using, also TGI?

nsarrazin:
Gonna merge this already just so we have the temperature update 😄

nsarrazin (approved):

gugarosa:
Thanks so much @nsarrazin!

Exactly! I launched a simple TGI with:

```bash
model=microsoft/Phi-3-mini-4k-instruct
volume=$PWD/data

docker run --gpus all \
    --shm-size 1g \
    -p 8080:80 \
    -v $volume:/data ghcr.io/huggingface/text-generation-inference:latest \
    --model-id $model \
    --trust-remote-code
```

And deployed the chat-ui using the endpoint generated above. For the chat-ui configuration, I used the same one defined by the `prod.yaml`.

gugarosa:
Could it be that the served production model is using some "outdated" files or a previous revision we had in the HF repository?

PR #1112, opened 2024-05-06
Fix context of logger

### LINKED ISSUE #1111: Logger is incorrectly called
In the current implementation, an error occurs when an error is thrown and the callback is handled by pinos logger. The issue arises because pinos log function requires a specific `this` context to operate correctly. When is's not the case, it results in the following error:

```
/app/node_modules/pino/lib/tools.js:59
      if (typeof this[msgPrefixSym] === 'string' && msg !== undefined && msg !== null) {
                     ^

TypeError: Cannot read properties of undefined (reading 'Symbol(pino.msgPrefix)')
    at LOG (/app/node_modules/pino/lib/tools.js:59:22)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```
Similar to [an issue in Pinos repo](https://github.com/pinojs/pino/issues/569#issuecomment-449363296)

jorado:
This pull request fixes #1111 by using an arrow function instead.

### DIFF
--- a/src/lib/server/database.ts
+++ b/src/lib/server/database.ts
@@ -133,22 +133,22 @@ export class Database {
 				{ sessionId: 1, updatedAt: -1 },
 				{ partialFilterExpression: { sessionId: { $exists: true } } }
 			)
-			.catch(logger.error);
+			.catch(e => logger.error(e));
 		conversations
 			.createIndex(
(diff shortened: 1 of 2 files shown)


nsarrazin (approved):
Nice thank you for the fix! 🚀

PR #1128, opened 2024-05-10
Use JWT email as login user

lijunle:
I am hosting Chat-UI after Cloudflare Tunnel and using Cloudflare Access for SSO.

After SSO, Cloudflare will inject a cookie `CF_Authorization`. Its value is a JWT. We could use the email after decoding the JWT value as login user.

The conversation will be kept for the same login user.

Besides, I don't think it makes sense to show the "Sign Out" button while decoding login user from JWT. So, that button is hidden (with a tricky way).

### DIFF
--- a/.env
+++ b/.env
@@ -6,6 +6,7 @@ MONGODB_DB_NAME=chat-ui
 MONGODB_DIRECT_CONNECTION=false
 
 COOKIE_NAME=hf-chat
+COOKIE_JWT_NAME=CF_Authorization
 HF_TOKEN=#hf_<token> from https://huggingface.co/settings/token
 HF_API_ROOT=https://api-inference.huggingface.co/models
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,6 +24,7 @@
 				"ip-address": "^9.0.5",
 				"jsdom": "^22.0.0",
 				"json5": "^2.2.3",
+				"jwt-decode": "^4.0.0",
 				"marked": "^12.0.1",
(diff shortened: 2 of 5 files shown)


nsarrazin:
Hi! Thanks for the contribution, big fan of cloudflare tunnels 😁 

I pushed a commit to fix the user creation and make sure everything passes the tests. If it still works for you let me know and I'll be happy to merge it!

lijunle:
@nsarrazin I am looking at Open-WebUI and they are using HTTP header instead of cookie to retrieve the user email. It looks like that is flexible to be compatible with Cloudflare Tunnel, Tailscale, OAuth0 and other SSO. Do you think that is a better option?

https://docs.openwebui.com/tutorial/sso#cloudflare-tunnel-with-cloudflare-access

nsarrazin:
Feel free to take a quick look at the PR if you have the time, we won't be using it for huggingchat but because this touches the login code i'd like to get some other pairs of eyes on the PR :hugs:

coyotte508 (approved):
Nice to see a small diff like this for a feature add :)

My only question is do we need to set `hfUserId: email` but looks good

evalstate:
Made a couple of extra changes in my branch as wanted User to be persisted so assistant functionality remained, and kept sign out to point at "<cloudflare>/access/logout".  Removed this line 
https://github.com/huggingface/chat-ui/blob/d67b28042b9c7d37fc6a20a055be4a83bd8ae059/src/hooks.server.ts#L158 
as looks like server session id is being incorrectly set temporarily.

PR #1129, opened 2024-05-10
[fix] Assistant websearch text/markdown source

mishig25:
## Description

When you combine assistant websearch with TEI, there was an error. (However, it was/is working fine with local models through transformers.js).

Although we have 

https://github.com/huggingface/chat-ui/blob/b2bee6290278fb70bba65fd2df8227508facc313/src/lib/server/embeddingEndpoints/tei/embeddingEndpoints.ts#L63, there was still an error (i.e. TEI endpoint was rejecting the input citing it as `invalid json`) when a text value was a text concat value (i.e. rather than `"Hello World"`, the actual value was `"Hello " + "World"`). Therefore, we needed to add https://github.com/huggingface/chat-ui/blob/b2bee6290278fb70bba65fd2df8227508facc313/src/lib/server/websearch/parseWeb.ts#L35-L37

### on main

image

### this PR
Screenshot 2024-05-10 at 14 50 55

### DIFF
--- a/src/lib/server/websearch/parseWeb.ts
+++ b/src/lib/server/websearch/parseWeb.ts
@@ -32,7 +32,9 @@ export async function parseWeb(url: string) {
 		r.headers.get("content-type")?.includes("text/plain") ||
 		r.headers.get("content-type")?.includes("text/markdown")
 	) {
-		return r.text();
+		const text = await r.text();
+		// JSON.stringify is needed to turn string concatenation into a single string (ex: "Hello, " + "world!" -> "Hello, world!")
+		return JSON.stringify(text);
 	} else {
 		throw new Error("Unsupported content type");
 	}


src/lib/server/websearch/parseWeb.ts
```diff
@@ -32,7 +32,9 @@ export async function parseWeb(url: string) {
 		r.headers.get("content-type")?.includes("text/plain") ||
 		r.headers.get("content-type")?.includes("text/markdown")
 	) {
-		return r.text();
+		const text = await r.text();
+		// JSON.stringify is needed to turn string concatenation into a single string (ex: "Hello, " + "world!" -> "Hello, world!")
```
coyotte508, line 36:
```shell
> JSON.stringify('"Hello, " + "world!"')
'"\\"Hello, \\" + \\"world!\\""'
```

It doesn't give `"Hello, world!"`

Not sure I understand where the concat happens

mishig25: example:
```ts
const text = await r.text();
console.log({text})
// {text: "Hello " + "world"}
```

and TEI endpoint was failing with `Invalid json`

```ts
let text = r.text();
text = JSON.stringify(text)
console.log({text})
// {text: "Hello  world"}
```

TEI endpoint works fine

coyotte508: will let @nsarrazin review I don't understand it

PR #1137, opened 2024-05-13
Add custom metrics for messages and conversations

antonioramos1:
Adding two new custom metrics, this will improve the understanding of how users are interacting with chat-ui. Similar to https://github.com/huggingface/chat-ui/pull/1013 but using Prometheus since we have gone down this route.

I am not very skilled in js, so let me know if you would like me to change anything.

It would be really nice to also have metrics to track exceptions but I will leave that to someone more skilled 😄 

Thanks!


nsarrazin:
I used your changes in #1186 so closing this, thank you for the contribution :smile:

PR #1150, opened 2024-05-17
Update examples for Yi

xianbaoqian:
Proposed 3 examples related to naming, search and travel planning. Let me know if you have more interesting ideas!

Regarding the second item that queries the internet, I wonder if there is a way to automatically enable internet search for that example?


julien-c:
> I wonder if there is a way to automatically enable internet search for that example

would make sense imo!

PR #1162, opened 2024-05-23
fix: parse ISS if returned in OIDC flow

muscionig:
## Description 

This PR fixes a bug where `chat-ui` does not parse the `iss` if the OIDC provider returns it in the `url` (for example, when using `keycloak>23`). 

Let me know if you would like me to open an issue for this as well.

### Minor Changes 

1. optional parse of `iss` from `url.searchParams.entries()`
2. updated `getOIDCUserData` to pass the `iss`

Tests:
1. [x] `npm run test`
2. [x] tested with a OIDC provider that does not return `iss` (previous behavior is maintained, as `iss` is undefined, the same if condition from the `openid-client` library is evaluated.)
3. [x] tested with a OIDC provider that returns `iss` and working


nsarrazin:
Thanks for the contribution!

PR #1163, opened 2024-05-24
Add error message for tools

Saghen:
Adds an indicator in the UI when the tool fails to run as well as providing the error message for debugging. Also fixes an issue with web search infinitely loading after erroring.

image


nsarrazin (approved):
Works well in local testing!

PR #996, opened 2024-04-09
Function calling

nsarrazin:
<details>
   <summary> **Screenshots** </summary>

</details>

* Probably only works well with command R+, requires a tokenizer with chat prompt templates that has multiple prompt templates for tools
* Hardcoded a few tools
    * websearch
    * text 2 img 
    * calculator
    * url fetcher (injects url content as plaintext)

PR #1164, opened 2024-05-24
Fix message update migration not applying

Saghen:
Fixes the message update migrations needed as part of #996 not applying to the tools objects.

### DIFF
--- a/src/lib/migrations/routines/04-update-message-updates.ts
+++ b/src/lib/migrations/routines/04-update-message-updates.ts
@@ -150,7 +150,7 @@ const updateMessageUpdates: Migration = {
 	_id: new ObjectId("5f9f4f4f4f4f4f4f4f4f4f4f"),
 	name: "Convert message updates to the new schema",
 	up: async () => {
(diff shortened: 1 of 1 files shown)


gary149 (approved):

gary149:
feel free to merge directly @mishig25

src/lib/migrations/routines/04-update-message-updates.ts
```diff
@@ -150,7 +150,7 @@ const updateMessageUpdates: Migration = {
 	_id: new ObjectId("5f9f4f4f4f4f4f4f4f4f4f4f"),
```
mishig25, line 150: don't we have to change the id? 
Otherwise, since migration with this id was already recorded as done, it will not run. No?

mishig25: or should we create new migration `src/lib/migrations/routines/06-xyz.ts` ?

Saghen: Yeah good call, updated the id so it will run again. The first run didn't work so this should be safe

PR #1168, opened 2024-05-24
Simplify tool error handling

Saghen:
Continues #1163 by having tools simplify throw errors rather than returning an explicit error or success status. Also introduced a util for stringifying unknown errors so we'll hopefully find less `[Object object]` errors downstream from libraries which don't use the native `Error` object

### DIFF
--- a/src/lib/server/tools/calculator.ts
+++ b/src/lib/server/tools/calculator.ts
@@ -22,14 +22,10 @@ const calculator: BackendTool = {
 			const query = blocks[blocks.length - 1].replace(/[^-()\d/*+.]/g, "");
 
 			return {
-				status: ToolResultStatus.Success,
 				outputs: [{ calculator: `${query} = ${vm.runInNewContext(query)}` }],
 			};
-		} catch (e) {
-			return {
-				status: ToolResultStatus.Error,
-				message: "Invalid expression",
-			};
+		} catch (cause) {
+			throw Error("Invalid expression", { cause });
 		}
 	},
 };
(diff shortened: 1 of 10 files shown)


mishig25:
the CI is read. For example, this line https://github.com/huggingface/chat-ui/blob/ada8a936583603fcb741ae182ef9e05af7ac209b/src/lib/server/tools/calculator.ts#L1 is no longer needed

src/lib/server/tools/calculator.ts
```diff
@@ -22,14 +22,10 @@ const calculator: BackendTool = {
...
-				status: ToolResultStatus.Success,
 				outputs: [{ calculator: `${query} = ${vm.runInNewContext(query)}` }],
 			};
-		} catch (e) {
-			return {
-				status: ToolResultStatus.Error,
-				message: "Invalid expression",
-			};
+		} catch (cause) {
+			throw Error("Invalid expression", { cause });
```
mishig25, line 28:
```suggestion
		} catch (err) {
			throw Error("Invalid expression", { cause: err });
```

mishig25:
there is a merge conflict, afterwards LGTM !

mishig25 (approved):
Fixed the merge conflicts. LGTM! Merging it.

### simulated a prod error

on `main`
Screenshot 2024-05-28 at 10 45 18

on `this branch`
Screenshot 2024-05-28 at 10 43 16

one small follow-up PR we can do is: like https://github.com/huggingface/chat-ui/pull/1188, we can improve the UI when there is an error (as an example, like red background color on error msg etc. cc: @gary149)

PR #1170, opened 2024-05-24
Clear playwright singleton on close

Saghen:
Fixes an issue where the playwright context will remain closed because we don't attempt to recreate it after it closes. This solves it by clearing the singleton on close.


mishig25 (approved):
lgtm !

PR #1184, opened 2024-05-27
fix(playwright): close context after each requests

rtrompier:


Saghen:
Continuing this in #1187

PR #1190, opened 2024-05-28
Auto-growing input for editing existing messages

Saghen:
image

Changes the message input to use a `<span>` with `contenteditable` and `role="textarea` (for accessibility). This allows the textarea to auto-expand to the content and avoids relayout on edit. I've found it improves the UX for editing larger prompts

### DIFF
--- a/src/lib/components/chat/ChatMessage.svelte
+++ b/src/lib/components/chat/ChatMessage.svelte
@@ -198,7 +198,7 @@
 	$: if (editMode) {
 		tick();
 		if (editContentEl) {
-			editContentEl.innerText = message.content;
+			editContentEl.textContent = message.content;
 			editContentEl?.focus();
 		}
 	}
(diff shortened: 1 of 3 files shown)


src/lib/components/chat/ChatMessage.svelte
```diff
@@ -410,29 +410,25 @@
...
+				>
+					<span
+						role="textbox"
+						tabindex={1}
+						class="w-full whitespace-break-spaces text-wrap break-words break-words rounded-xl px-5 py-3.5 text-gray-500 outline-none *:h-max dark:text-gray-400"
+						class:bg-gray-100={editMode}
+						class:dark:bg-gray-800={editMode}
+						bind:this={editContentEl}
+						on:keydown={handleKeyDown}
+						contenteditable={editMode}>{message.content.trim()}</span
```
gary149, line 429: I can paste html and it renders:

image


I would stay with a`<textarea>` and do the elastic height like we already do on https://github.com/huggingface/chat-ui/blob/e9c846acd33f806b30498f67140bac9579c1c33b/src/lib/components/chat/ChatInput.svelte#L46

PR #1201, opened 2024-05-28
feat: allow disabling metrics server, consolidate exit handlers

Saghen:
The metrics server causes issues in development due to the Sveltekit not allowing us to hook into restarts (as far as I could tell). This allows disabling the metrics server with `METRICS_ENABLED=false` and a bit of documentation to recommend this setting for development (maybe it should be default?). I've also consolidated the exit handler code to make signal handling more consistent


nsarrazin (approved):
Swapped things so that the metrics server is off by default and on in prod, other than that LGTM! Thanks for tackling this 🚀

PR #1204, opened 2024-05-28
feat: mime types on tools, show upload btn on demand

Saghen:
Adds a mimetypes property on the tools used by the frontend to control the `UploadBtn`'s `accept` property. If there are no active mime types, the upload button will not be shown. Adds an enable/disable all button to the tools menu

image

### DIFF
--- a/src/lib/components/ToolsMenu.svelte
+++ b/src/lib/components/ToolsMenu.svelte
@@ -60,10 +60,12 @@
 				<button
 					class="ml-auto text-xs underline"
 					on:click|stopPropagation={() => setAllTools(!allToolsEnabled)}
-					>{#if allToolsEnabled}
+				>
+					{#if allToolsEnabled}
 						Disable all
 					{:else}
-						Enable all{/if}
+						Enable all
+					{/if}
 				</button>
 			</div>
 			{#each $page.data.tools as tool}
--- a/src/lib/components/ToolsMenu.svelte
+++ b/src/lib/components/ToolsMenu.svelte
@@ -22,6 +22,8 @@
 			tools: Object.fromEntries($page.data.tools.map((tool: ToolFront) => [tool.name, value])),
 		});
 	}
+	$: allToolsEnabled = activeToolCount === $page.data.tools.length;
+	$: allToolsDisabled = activeToolCount === 0;
 </script>
 
 <details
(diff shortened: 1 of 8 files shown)


src/lib/components/ToolsMenu.svelte
```diff
@@ -50,6 +56,10 @@
 					>
 				{/if}
 			</div>
+			<div class="flex space-x-4 text-sm text-gray-500">
+				<button on:click={() => setAllTools(true)}>Enable all</button>
+				<button on:click={() => setAllTools(false)}>Disable all</button>
+			</div>
```
mishig25, line 62:
```suggestion
			<div class="flex space-x-4 text-sm text-gray-500">
				<button
					disabled={allToolsEnabled}
					class:opacity-50={allToolsEnabled}
					on:click={() => setAllTools(true)}>Enable all</button
				>
				<button
					disabled={allToolsDisabled}
					class:opacity-50={allToolsDisabled}
					on:click={() => setAllTools(false)}>Disable all</button
				>
			</div>
```

and define after `$: activeToolCount` definition
```ts
	$: allToolsEnabled = activeToolCount === $page.data.tools.length;
	$: allToolsDisabled = activeToolCount === 0;
```

for a better UX

src/lib/components/UploadBtn.svelte
```diff
@@ -3,6 +3,7 @@
 
 	export let classNames = "";
 	export let files: File[];
+	export let mimeTypes: string[] = ["*/*"];
```
mishig25, line 6:
```suggestion
	export let mimeTypes: string[];
```

`mimeTypes` prop always has valid value. No ?

src/lib/components/ToolsMenu.svelte
gary149: didn't try but I guess it should use <input type="checkbox"> with indeterminate state
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate_state_checkboxes

Saghen: image
image

thanks Mishig, looks great imo!

mishig25 (approved):
lgtm !

src/lib/components/ToolsMenu.svelte
gary149: could be smaller I can tweak it a bit 👀

nsarrazin:
Works great! I extended support to the dropzone & onpaste handler in [this PR](https://github.com/huggingface/chat-ui/pull/1222)

gary149 (approved):

PR #1205, opened 2024-05-28
feat: skip tools when all disabled

Saghen:
Skips requesting the model pick tools when all of the tools have been disabled in the UI.

### DIFF
--- a/src/lib/server/textGeneration/index.ts
+++ b/src/lib/server/textGeneration/index.ts
@@ -19,6 +19,8 @@ import { generate } from "./generate";
 import { mergeAsyncGenerators } from "$lib/utils/mergeAsyncGenerators";
 import type { TextGenerationContext } from "./types";
 import type { ToolResult } from "$lib/types/Tool";
+import { toolHasName } from "../tools/utils";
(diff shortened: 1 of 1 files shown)


src/lib/server/textGeneration/index.ts
```diff
@@ -61,7 +61,7 @@ async function* textGenerationWithoutTitle(
 
 	if (model.tools && !conv.assistantId) {
 		const tools = pickTools(toolsPreference, Boolean(assistant));
-		toolResults = yield* runTools(ctx, tools, preprompt);
+		if (tools.length > 0) toolResults = yield* runTools(ctx, tools, preprompt);
```
nsarrazin, line 64: Wouldn't `tools.length` always be 1 at a minimum ? Since [directlyAnswer](https://github.com/huggingface/chat-ui/blob/c52d8aaa958b184dab27ec1ebfe63e9f3d1de75b/src/lib/server/tools/directlyAnswer.ts#L5-L8) is always enabled and is hidden from the user. I think we'd need to filter that one out too

nsarrazin (approved):

PR #1206, opened 2024-05-28
feat: support file pasting in chat input

Saghen:
Adds support for pasting files as an alternative to using the file picker or dropzone. Checks that the input has been focused before accepting the paste

### DIFF
--- a/src/lib/components/chat/ChatInput.svelte
+++ b/src/lib/components/chat/ChatInput.svelte
@@ -7,6 +7,7 @@
 	export let maxRows: null | number = null;
 	export let placeholder = "";
 	export let disabled = false;
+	export let files: File[] = [];
 
 	let textareaElement: HTMLTextAreaElement;
 	let isCompositionOn = false;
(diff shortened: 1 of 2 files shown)


src/lib/components/chat/ChatInput.svelte
```diff
@@ -35,6 +36,23 @@
...
+
+			for (const item of items) {
+				if (item.kind !== "file") continue;
+				const file = item.getAsFile();
+				if (file) files = [...files, file];
+			}
+		};
+		document.addEventListener("paste", handlePaste);
+		return () => document.removeEventListener("paste", handlePaste);
+	});
```
gary149, line 55:
```suggestion
		function handlePaste (event: ClipboardEvent) => {
			if (!textareaElement.contains(document.activeElement)) return;

			const items = event.clipboardData?.items;
			if (!items) return;

			for (const item of items) {
				if (item.kind !== "file") continue;
				const file = item.getAsFile();
				if (file) files = [...files, file];
			}
		};
```

Then forward the event from ChatInput with `on:paste` and just use <ChatInput on:paste={handlePaste} />

mishig25: https://github.com/huggingface/chat-ui/pull/1212

nsarrazin (approved):
Tested locally, works great

PR #1213, opened 2024-05-29
Add super basic admin feature

nsarrazin:
Only lets you unfeature/delete assistants for now.

Was needed because deleting assistants requires deleting it from every user's settings which was impossible to do manually

Screenshot from 2024-05-29 11-49-05

Currently admins are created by setting `isAdmin` manually on the user document in the DB


coyotte508 (approved):

mishig25 (approved):
lgtm !

PR #1215, opened 2024-05-29
Fix migration object id

nsarrazin:
`new ObjectId(6)` does not return the same value every time :sweat_smile: 

Replaced it with `new ObjectId("000000000006")` which does


nsarrazin:
Might need to clean up some migration results documents in the DB

Saghen (approved):

PR #1216, opened 2024-05-29
Let users access the API using their HF token

nsarrazin:
Simple PR that lets you use bearer authentication against the API if the flag `USE_HF_TOKEN_IN_API` is set to true.  (Currently false while we flesh out API & limits)

This stores token hashes in a cache with a TTL of 5min. If the cache misses, we check against `whoami` API, grab the user id and check it against our user DB.

Caveat is that the user needs to have signed up once on huggingchat before being able to use the API. (since we don't have access to the user DB of the hub) But I think that's fine


nsarrazin:
no rush at all on the review, not on the roadmap just a nice to have I wanted to tackle :smile:

coyotte508:
> Caveat is that the user needs to have signed up once on huggingchat before being able to use the API. (since we don't have access to the user DB of the hub) But I think that's fine

We should have enough info in the whoami response to create the user I think

Also we probably want to make the backend calls to API inference with the token. In that case you can add a `X-Required-Access: inference-api` header to any calls (including whoami) to the hub to make sure the token has access to the inference api.

Note that we'll also want to use the OAuth access token to make calls to the inference backend (will open an issue)

PR #1228, opened 2024-05-30
Fix error logging

wdhorton:
Logging for chat-ui was switched to Pino in https://github.com/huggingface/chat-ui/pull/1086. For error logging, Pino wants the object first, and then the string message. Otherwise you don't get the information from the error object. For example see this thread: https://github.com/pinojs/pino/issues/673

There's already one example in the repo of correct usage: https://github.com/huggingface/chat-ui/blob/main/src/lib/server/textGeneration/tools.ts#L95


nsarrazin (approved):
Oof 😬 Thanks for catching that!

PR #1230, opened 2024-05-30
Document Parser RAG

Saghen:
Adds support for RAG to the document parser. The implementation on the document-parser-rag side simply splits into 512 character chunks, embeds and performs cosine similarity between the embeddings and the queries

Known issues:

- [ ] Space supports multiple files and queries but the tool does not
- [ ] Command R+ frequently generates poor queries due to lack of context


nsarrazin:
Closing since this was replaced in tools refacto

PR #1234, opened 2024-05-31
Fix: tools in endpointVertex now optional

SebSchroen:
The googleSearchRetrieval leads to very odd and unnatural behavior when chatting with Gemini. Making the argument optional and removing it for certain use cases fixes the issue. 

Behavior with googleSearchRetrieval:

Gemini_Pro_with_googleSearchRetrieval


nsarrazin (approved):

nsarrazin:
Thanks for the contribution!

PR #1237, opened 2024-06-01
Delete chat entries

krakenftw:
wanted to know if i am doing this correctly and how would i access `messageId` inside `deleteBranch` action, Thanks 😄


krakenftw:
I figured it out 👍🏻

PR #1239, opened 2024-06-01
Fix#1210

krakenftw:
Tried adding the delete branch feature, Does not work.
It shows a replace error when i click on the delete button, any help would be appreciated


nsarrazin:
Hey so you leaked your token in `.env` and it's visible even in [the old commit](https://github.com/huggingface/chat-ui/pull/1239/commits/43266250624773504c1722b34a45c60286abad5c#diff-e9cbb0224c4a3d23a6019ba557e0cd568c1ad5e1582ff1e335fb7d99b7a1055dL11), so you should really remove it in your profile ASAP. 

You should add overrides to `.env.local` not in `.env` in the future, that one is gitignored

krakenftw:
Yeah i realised it later and revoked it. can you help me with the issue on deleting branch?

omesh-omg (changes requested):
put your MONGODB_URL and HF_TOKEN by creating a .env.local file in the root of the repository

PR #1243, opened 2024-06-03
Npm: Remove updateProdEnv from script from package.json

rodrigobdz:
The script itself and the documentation about it were deleted in https://github.com/huggingface/chat-ui/pull/1107 already.

### Removed

- Script `updateProdEnv` from `package.json`


nsarrazin (approved):
Thanks for the catch!

PR #1250, opened 2024-06-05
Community tools

nsarrazin:
# Community tools

## Features
* New `/tools` page with list of tools
  * Show official tools & community tools
  * See active tools, trending tools & your own tools
  * Filter by name, top of all time, trending
  * See number of tool runs
* Click on a tool to see details
  * Edit/delete tool if you're the owner
  * View tool spec if you're not the owner
  * Share with direct link
  * Activate/deactivate tool
* Create a new tool
  * Enter huggingface space URL to get input list
  * Automatic prefilling of description & default values from API
  * Automatic parsing of output to extract text & file
  * Support file input with restricted MIME type
  * Support optional & fixed value tool inputs
  * Get a warning if tool output is not supported (audio, video, etc.)
 
> [!IMPORTANT]
> The feature is currently feature flagged, so your user in the db must have `isEarlyAccess: true` in order to see any part of the feature. This will let us test in production internally before releasing to the public. See #1365 for more details

### Main window
Screenshot from 2024-07-31 16-27-56

### `/tools` page
Screenshot from 2024-07-31 16-36-10

### `/tools/[toolId]` page
Screenshot from 2024-07-31 16-34-51

### New tool/ tool edit form
Screenshot from 2024-07-31 16-34-43


nsarrazin:
Overall this should be good to go @gary149, the feature is fully hidden behind a feature flag so we can deploy & test internally for now

mishig25:
First round of review on UX:
1. `filter by name` is not working. For example, you can go to URL http://localhost:5173/chat/tools?sort=trending&q=parser where I expect the tools that only have substring `parser` in their name to be present. But all the tools are present.  image
2. On medium sized screen, tool titles that are long like `Image Generation` get cut image
3. There should be limit on tool name and tool description 
image

4. when a space does not have a function that can be called as a tool, we should show a label that says: `Space does not have a callable function`. Example you can try with is: `predatortoabuse/Realistic-Image-Generator-Model` image

5. durin toolmcreation, save btn should be disabled until all the fields are valid (i.e. have values) 
image

6. I can enter `string` into `float` values. I guess we should not ? image

7. `cancel` btn in tools should go back to previous page. Right now, it is redirecting to homepage  image

nsarrazin:
Overall I think I should improve the inputs in the tool creation form. Use number input fields for `Int` and `Float` types, a checkbox for `boolean` types and a radio with a list of options for `Literal['optionA', 'optionB']` types. That should cover 99% of input types. Will tackle this

nsarrazin:
@mishig25 I implemented your feedback, should be better now! 

I also refactored the input selection field for default values, it now only accepts floats for float inputs, integers for int inputs, shows a checkbox for boolean selection and a dropdown select for `Literal` types with multiple choices

nsarrazin:
Going to merge this, since we have feature flags so it will be invisible to everyone and I can ensure everything is working well in prod.

PR #1251, opened 2024-06-05
Fix max upload size in huggingchat

nsarrazin:
We were running into issues with the file upload mechanism. The old way was to encode the file as a `base64` string and POST a request with a JSON body, including the files. This worked OK locally but the body got too big in prod and the request would get denied with a 413. 

So I switched to using `FormData` instead (like we do for the assistant avatar since that seemed to work ok).

 In order to keep the PR simple, everything except the files is still a stringified JSON inside of `form.get('data')` but I moved the files to `form.getAll("files")`

### DIFF
--- a/src/lib/utils/messageUpdates.ts
+++ b/src/lib/utils/messageUpdates.ts
@@ -59,18 +59,28 @@ export async function fetchMessageUpdates(
 	const abortController = new AbortController();
 	abortSignal.addEventListener("abort", () => abortController.abort());
 
+	const form = new FormData();
+
+	const optsJSON = JSON.stringify({
+		inputs: opts.inputs,
+		id: opts.messageId,
+		is_retry: opts.isRetry,
+		is_continue: opts.isContinue,
+		web_search: opts.webSearch,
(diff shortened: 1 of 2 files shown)


nsarrazin:
cc @alak I think this will break the iOS app so I want to coordinate with you to minimize downtime. I don't think the iOS app supports file upload yet, so the main thing for you would be to wrap the [request parameters](https://github.com/huggingface/chat-ui/pull/1251/files#diff-278e19521eb0ad173b98387709ab1c22e0b5cddf90aec08ba7a403db146bd845R64-R71) in a stringified JSON, and add that to a [FormData](https://github.com/huggingface/chat-ui/pull/1251/files#diff-278e19521eb0ad173b98387709ab1c22e0b5cddf90aec08ba7a403db146bd845R79). 

Make sure to also [change the header](https://github.com/huggingface/chat-ui/pull/1251/files#diff-278e19521eb0ad173b98387709ab1c22e0b5cddf90aec08ba7a403db146bd845L64) in the POST request

PR #1252, opened 2024-06-06
feat(chart): service account

rtrompier:


chart/env/prod.yaml
```diff
@@ -8,7 +8,14 @@ nodeSelector:
...
   - key: CriticalAddonsOnly
     operator: Equal
-    
+
+serviceAccount:
+  enabled: true
+  create: true
+  name: huggingchat-prod
+  annotations:
+    "eks.amazonaws.com/role-arn": "arn:aws:iam::707930574880:role/hub-prod-huggingchat"
```
XciD, line 17: maybe we should hardcode that in argo ?

nsarrazin (approved):
lgtm, do we need to also change the secret before merging & deploying this ?

PR #1257, opened 2024-06-08
Refactor logger error messages for consistency.

dnzzzl:
Corrected the use of `logger.error(obj:any,msg:string)` which enabled correct logging of errors and ease of debugging.


nsarrazin (approved):
Thanks, makes sense to me!

PR #1274, opened 2024-06-11
fix(scrape parser): Cannot read properties of 'undefined' during page parsing

neven4:
Fix of the error that occurs during parsing of the web search results

image


nsarrazin:
Thanks for the contribution! :rocket:

PR #1294, opened 2024-06-19
Auto open app

itsbhatt:


nsarrazin (changes requested):
Hi, thanks for the contribution! I think having the browser open automatically is a good addition, however I don't think we should expose the dev server to the local network by default. 

Depending on your environment this can be a bit of a security hazard, so I would prefer if users were explicit about it. If you revert that part then the rest looks good to me!

itsbhatt:
@nsarrazin Done.

nsarrazin (approved):
Thanks for the contribution!

Issue #1113, opened 2024-05-06
uploading of image is not accessible


vojtapolasek:
Hello,
I am a screen reader user. When chatting with some models on hf.co/chat, the screen reader announces that there is a "file drop zone". It is located near to the edit box. This makes me believe taht I should be able to upload a file there. I would like to try out various models for image description... but I can't seem to be able to upload a file. Could you please check accessibility of controls for file uploads?
Thank you.

nsarrazin:
Hello!

I think this is an accessibility issue on our side. The file upload component exists, but should only be visible to the user when specific models flagged as multimodal are selected.

We currently do not have such models active on HuggingChat so the file dropzone should be hidden. Seems like the screen reader is still picking up on it, so it is probably not hidden properly. That's probably why you can't upload a file 😄 

If someone wants to tackle this, feel free! Otherwise I'll try to do it at some point.

PR #1295, opened 2024-06-19
fix(screen-reader): disable FileDropZone area-label for models without file upload

neven4:
Closes: https://github.com/huggingface/chat-ui/issues/1113


nsarrazin (approved):
Working great, thanks for the fix! :fire:

Issue #1298, opened 2024-06-21
Option to turn on/off playwright (adblocker).


calycekr:
There are many problems due to playwright in offline environment. I have to wait for playwright fetch timeout before the application is fully loaded.

`ERROR (1399872): Failed to initialize PlaywrightBlocker from prebuilt lists`

I need an option to turn it on and off.
Please add an option to turn it off, not to deal with errors.

nsarrazin:
Makes a lot of sense to me! Let me know if someone wants to tackle this, I think putting an env variable `PLAYWRIGHT_ADBLOCK=true` by default in the `.env` which could be set to `false` should work.

krakenftw:
can i work on this? @nsarrazin

nsarrazin:
Sure thing @krakenftw

nsarrazin:
Just using an env variable in `.env` yes

krakenftw:
@calycekr Yeah, i did add a  boolean env, and disabling `blocker` based on that env. It has'nt been merged yet tho.
Hoping a response from @nsarrazin

nsarrazin:
thanks for noticing @calycekr should be fixed with https://github.com/huggingface/chat-ui/commit/df67a2c0ba5140d68241589e1f73da03cc3c3843 ! let me know if not

PR #1301, opened 2024-06-24
feat: playwright-adblocked toggle #1298

krakenftw:
added env `PLAYRIGHT_ADBLOCK` which enables/disables the playright-adblock.
please review and let me know if that correct or i do need some changes


krakenftw:
@nsarrazin can you please review for #1298

krakenftw:
What else is left here to merge? @nsarrazin

src/lib/server/websearch/scrape/playwright.ts
```diff
@@ -64,7 +64,7 @@ export async function withPage<T>(
 
 	try {
 		const page = await ctx.newPage();
-		await blocker.enableBlockingInPage(page);
+		process.env.PLAYWRIGHT_ADBLOCKER && (await blocker.enableBlockingInPage(page));
```
nsarrazin, line 67:
```suggestion
		(/true/i).test(process.env.PLAYWRIGHT_ADBLOCKER) && (await blocker.enableBlockingInPage(page));
```

Since if you set `PLAYWRIGHT_ADBLOCKER=false` it will actually be `"false"` which evaluates to true since it's not an empty string

Other than than LGTM!

nsarrazin (changes requested):

nsarrazin:
Thanks for the contrib! My bad the regex failed the type checks so I rolled it back 😬

krakenftw:
All good, thanks for the merge : )

Issue #1314, opened 2024-06-27
"Sign in with 🤗 Hugging Face" button for login modal can look strange at narrow widths


Mardak:
Relatively minor issue with left side of this screenshot shows the beginning and end of the button text overflowing:
<img width="1198" alt="narrow hugging chat login" src="https://github.com/huggingface/chat-ui/assets/438537/67f79795-c932-471d-9667-f0b0c57f0629">

The current code happens to have newline/whitespace that could be preserved or somehow shorten the text?
https://github.com/huggingface/chat-ui/blob/70e1e30dc333e0506b07e3a1477f3c706e8f6080/src/lib/components/LoginModal.svelte#L38-L44

julien-c:
good catch, a PR is welcome!

Mardak:
Specifically for the Firefox chatbot sidebar use case, we might end up enforcing a minimum width of say 420px and avoid this issue for huggingchat that only appears around 350px.

PR #1315, opened 2024-06-27
Fix "login" button text wrap on mobile

neven4:
Closes: https://github.com/huggingface/chat-ui/issues/1314


neven4:
Also, I can create a separate `LoginButton` component, but I'm not sure at what point you decide to extract code into a separate component. Is 2 repetitions enough?

nsarrazin (approved):
Thanks for the contrib! Looks great, will deploy 🚀

Issue #1302, opened 2024-06-24
Assistant feature: Send user query as part of template variable GET request


ethayu:
Trying to integrate RAG as an assistant. Thinking of using a template variable that makes a GET request (with the prompt as the request body), to get the relevant documents as context. Is this possible (i.e. there is a special variable in the system prompt page for the user query), or is there a better way of doing this?

ethayu:
Found this: https://github.com/huggingface/chat-ui/pull/949, but seems to only support `time` and `date`. Would be great if something like `{{ query }}` was supported, and secondly we could query {{url=123.123.123.12:8000/{{ query }}}}, or something like that.

ethayu:
https://github.com/huggingface/chat-ui/issues/611 would also make it doable.

PR #1322, opened 2024-07-01
add option to send user message as part of request body in assistant …

ethayu:
Add support for to send user message/query within assistant invocations, as referenced in #1302.


ethayu:
@nsarrazin please let me know your thoughts

ethayu:
@nsarrazin about keeping the old functionality:
my changes don't remove any existing functionality - it just makes it such that if a new env variable `ENABLE_ASSISTANTS_POST` is set to `true`, any call to `{{url=https://example.com}}` will be a POST request instead of a GET request, with the user prompt sent as the message body.

nsarrazin:
> it just makes it such that if a new env variable ENABLE_ASSISTANTS_POST is set to true, any call to {{url=https://example.com}} will be a POST request instead of a GET 

Yep, my point was that if we use a different keyword like `url|post` instead of setting it with an env variable then we can support both GET and POST in the same config :smile:

ethayu:
Yes, that actually makes more sense. I’ll do that.

nsarrazin:
Hi @ethayu do you still want to take a look at this? Just curious no pressure :smile:

ethayu:
@nsarrazin sorry, I've been a little swamped with school recently! I'll get to it around thanksgiving break, if nobody else has taken care of it by then

ethayu:
@nsarrazin done! please take a look 😄

ethayu:
@nsarrazin lint issue fixed 😂

ethayu:
@nsarrazin Could you kick off the checks please?

ethayu:
@nsarrazin any updates? Would like to get this in before I have to leave for school.

nsarrazin:
Thanks a lot @ethayu ! Taking a look at this now.

nsarrazin:
@ethayu I've added back the `url` option so it doesn't break legacy assistants on deployed versions of chat-ui. I've also removed the code that sends POST request from the browser, it wasn't working anyway most of the time because of CORS.


The rest looks great thanks for the PR! Will be merging this now.

nsarrazin (approved):

PR #1328, opened 2024-07-05
add android id in the hf manifest

alak:


nsarrazin:
Screenshot from 2024-07-05 14-22-46


Added a small button that shows up when `beforeinstallprompt` fires, you can close it with the X, and if you click on it elsewhere it should trigger the native app install prompt flow

kinda hard to test the android flow locally tbh, any clue on how to do that?

I can make the button show up by running this in the devtools console of my browser 
```
event = new Event('beforeinstallprompt')
window.dispatchEvent(event)
```

but the native install flow doesn't work afterwards

gary149:
This seems complicated and not sure it will work very well. let's **just use the user agent to detect if the user is on android** and display a dismissable banner on top of the app if it is the case imo.

nsarrazin:
here it is dark/light with portrait & landscape
| Screenshot from 2024-07-11 14-44-56 | Screenshot from 2024-07-11 14-45-45 |
| --- | --- |


we use the following logic to display it
* is hugging chat
* is android user agent
* is not android app user agent
* `modalLastSeen` is not defined OR it's been >72 hours  (date stored in localStorage)

src/routes/+layout.svelte
```diff
@@ -126,6 +129,16 @@
 	$: mobileNavTitle = ["/models", "/assistants", "/privacy"].includes($page.route.id ?? "")
 		? ""
 		: data.conversations.find((conv) => conv.id === $page.params.id)?.title;
+
+	let showAndroidModal =
+		browser && // only show on browser
+		isHuggingChat && // only show on huggingchat
+		navigator.userAgent.toLowerCase().includes("android") && // if it's android
+		!navigator.userAgent.includes("co.huggingface.chat_ui_android") && // but not the android app
+		(!localStorage.getItem("huggingChatLastSeenAndroidModal") || //and the user hasn't seen it yet
```
gary149, line 138: I think it's fine to display it everytime (it's the case on ios for example)

gary149 (approved):
Appart from my little comment on showing it every time I think we can merge.

nsarrazin (approved):

alak:
lgtm

PR #1333, opened 2024-07-09
perf: add prod deployment resources

jcudit:
- adds more replicas as a temporary measure to mitigate against crashes
- reduces cpu and mem requests to what we've observed in practice over the last week


nsarrazin (approved):

jcudit:
@nsarrazin feel free to send this out when convenient.  I'm not authorized to merge 😄

nsarrazin:
Thanks for tackling this! @jcudit

PR #1350, opened 2024-07-18
UX: Dropzone

gary149:
Dropzone shows when you drag on window


nsarrazin (approved):
Looks a lot better! 🚀

Issue #1286, opened 2024-06-14
Feature Request: Manage model through UI instead of .env.local


TriEyeMonster:
Since the website will be running in a production environment, it will be better that admin is able add or remove model through UI instead of a production deployment for a minor .env.local update.

nsarrazin:
Yeah that would be definitely be nice, I think it was on our roadmap a while ago but never got prioritized. If there's strong demand we could try to tackle this! It would make life a lot easier for self-hosting users.

TriEyeMonster:
Yeah, this is a essential feature if we want to make it more product ready.

widyaageng:
+1 for prioritisation.

neven4:
@nsarrazin how you want to handle `endpoints` array in the modal? Currently, we have a multiple types (I think it will be hard to unify them).

Do we want to have an additional mongo collection for each endpoint type (`AWS`, `openai`, `anthropic`), or just store types and json and validate it from the frontend side?

PR #1362, opened 2024-07-25
feat(embeddingModel): add embedding model into mongodb

neven4:
Hi there,

I decided to create the first PR which migrate config management from `.env` file to UI.

How do I see it:
- Migrate the `EmbeddingModel`
- Migrate the `Model`
- Create UI for `EmbeddingModel` and `Model`

WDYT?

Part of https://github.com/huggingface/chat-ui/issues/1286

### DIFF
--- a/scripts/populate.ts
+++ b/scripts/populate.ts
@@ -14,7 +14,7 @@ import type { User } from "../src/lib/types/User";
 import type { Assistant } from "../src/lib/types/Assistant";
 import type { Conversation } from "../src/lib/types/Conversation";
 import type { Settings } from "../src/lib/types/Settings";
-import { defaultEmbeddingModel } from "../src/lib/server/embeddingModels.ts";
+import { getDefaultEmbeddingModel } from "../src/lib/server/embeddingModels.ts";
 import { Message } from "../src/lib/types/Message.ts";
 
 import { addChildren } from "../src/lib/utils/tree/addChildren.ts";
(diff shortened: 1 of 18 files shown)


KartikGS:
Will the default embedding model setting be inserted in the database every time the server restarts?

neven4:
> Will the default embedding model setting be inserted in the database every time the server restarts?

For now - yes, in order to support the current behavior and allow users to edit embedding models via `.env` file (until the UI for editing is created)

neven4:
> Hi, thanks for the contrib, this looks like a nice feature, will do a deeper review soon but just curious, how do you handle the API keys stored in the env variable? Do you copy these over in the DB?

Hi there, I hope I got a question right.

I'm not touching API keys at all, I just parse `TEXT_EMBEDDING_MODELS` from the `.env` file and populate database with it.

nsarrazin:
I meant for example the OpenAI embedding endpoint has an `apiKey` property [(permalink)](https://github.com/huggingface/chat-ui/blob/ae0d33653a0b1eb1b6cba5c6edcc7c434886190b/src/lib/server/embeddingEndpoints/openai/embeddingEndpoints.ts#L11-L12) would it get copied over into the DB ? :eyes:

neven4:
Oh, good catch, I missed that part. In this case we need to encrypt/decrypt api keys using some encryption key 👍

neven4:
Hi @nsarrazin, I added an encryption for sensitive data 👍

PR #1365, opened 2024-07-26
feat(auth): Add admin & early access flags based on orgs membership

nsarrazin:
This PR adds a check on the login callback, so that if a user is part of a specific hub org they will get access to early access or admin features automatically.

This can be used to feature flag new features or give moderation rights automatically without going through the DB. 

Caveat is that the check only occurs on login, so flags are only updated on log-in, not when reusing a session. I think this is an OK tradeoff and if we need to force update flags on someone we can always either do it manually or force a logout by deleting matching sessions.


nsarrazin:
cc @coyotte508 I 

tried to keep it simple, no rush at all but let me know if you think this makes sense as an implementation! :smile:

coyotte508 (approved):
It's nice!

For the caveat, maybe we can switch to oauth-based sessions now that we have refresh tokens on HF. (so that we always have a valid oauth token)

Or if it's HF-only, we can use the `/api/users/.../overview` endpoint

julien-c:
what would be the impact on the measurement of DAUs?



> maybe we can switch to oauth-based sessions

coyotte508:
> what would be the impact on the measurement of DAUs?
> 
> > maybe we can switch to oauth-based sessions

I don't think any?

The user would be redirected to signin once, which would be seemless most likely

The sessions currently expire after two weeks of inactivity, and the oauth refresh tokens last 3 months, so we could renew the refresh token after one month and there would be no perceived difference

nsarrazin:
Opened #1377 to track the idea of using refresh tokens for the auth flow, will merge this PR for now as I think it can be tackled separately!

Issue #1372, opened 2024-07-27
Only the first few pixels of the tools bar buttons are accessible


dlavrenuek:
### Description

While playing around with the chat application I did notice some visual inconsistencies. The tools buttons displayed at the bottom right of each assistant's message overlap following user message which has a higher z-index and are cut off. Following is a video demonstrating the behavior and some screenshots:

https://github.com/user-attachments/assets/07d443ae-82a9-46d0-808d-d5ff92865a7a

<img width="1221" alt="chat-ui-buttons-overlap" src="https://github.com/user-attachments/assets/272c0956-1a60-46b3-89d0-f5bcd37f6f29">

This is because the space between the assistant's and user's messages is less than the toolbar height + offset. The space (gap 6 or 7 depending on screen size) is defined in https://github.com/huggingface/chat-ui/blob/fc2c114798d30662de88518d2181c546c8c4caff/src/lib/components/chat/ChatWindow.svelte#L217 and reduced by 2 (padding bottom 4 and negative margin bottom 6) due to negative margin in https://github.com/huggingface/chat-ui/blob/78e00251b731e85a97a785839d9d3720608aba48/src/lib/components/chat/ChatMessage.svelte#L237

Proposed solution: increase the gap between chat messages to be at least the size of buttons

#### Further visual bugs

The top and bottom distance from user's message to assistant's message is different. This is caused by ChatMessage adding a negative bottom margin (as described above) and thus reducing the gap defined in ChatWindow

<img width="529" alt="chat-ui-different-gap" src="https://github.com/user-attachments/assets/93543e8a-02e7-4e32-8ab0-fdfe2d234680">

The Copy button has a different color than the rest of buttons. This is because the color is overwritten for the copy icon itself altho the correct color is passed to the button component: https://github.com/huggingface/chat-ui/blob/745e51e28bab8c380564efb32056a3508ee94f11/src/lib/components/CopyToClipBoardBtn.svelte#L46

Suggestion: move the color classes from the icon to button https://github.com/huggingface/chat-ui/blob/745e51e28bab8c380564efb32056a3508ee94f11/src/lib/components/CopyToClipBoardBtn.svelte#L38

The copy button is wider than the rest. This is caused by additional margin to compensate that the icon/button width is 2px smaller.

<img width="345" alt="Bildschirmfoto 2024-07-27 um 21 17 13" src="https://github.com/user-attachments/assets/e5827c48-c799-47b3-857b-aef898358b5d">

### Context

[...]

PR #1373, opened 2024-07-27
Fix cut off buttons in chat message

dlavrenuek:
Fixes https://github.com/huggingface/chat-ui/issues/1372

### Description

From the linked issue following bugs are fixed:
- Cut off tool bar buttons
- Different spacing between messages (top/bottom)

Does not fix (due to having not enough context about why it was implemented this way):
- Copy button width
- Copy button color


nsarrazin (approved):
Thanks for the contribution! Appreciate the improvement here :smile:

PR #1379, opened 2024-07-30
fix(env) enable LLM_SUMMARIZATION

zhang-yi-chi:
env.LLM_SUMMARIZATION is not working because env var is string type.


nsarrazin (approved):
Appreciate the fix thanks!

PR #1396, opened 2024-08-07
Add steps to install from source for llama.cpp

Vaibhavs10:
ref: https://x.com/dnouri/status/1821197502760030686?s=46

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ Install llama.cpp w/ brew (for Mac):
 # install llama.cpp
 brew install llama.cpp
 ```
+
 or [build directly from the source](https://github.com/ggerganov/llama.cpp/blob/master/docs/build.md) for your target device:
 
 ```
--- a/README.md
+++ b/README.md
@@ -37,9 +37,21 @@ You can quickly start a locally running chat-ui & LLM text-generation server tha
 
 **Step 1 (Start llama.cpp server):**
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,9 +37,21 @@ You can quickly start a locally running chat-ui & LLM text-generation server tha
 
 **Step 1 (Start llama.cpp server):**
 
+Install llama.cpp w/ brew (for Mac):
```
julien-c, line 40: brew also exists for linux no?

julien-c (commented):
oooh nice

README.md
Vaibhavs10: It does, however, it only instals CPU only build, so would be quite slow. So recommended way would still be to `clone` + `make` imo.

Vaibhavs10:
CI is 🟢

dnouri:
I just found that the docs at llama.cpp have some really useful links for how to install on my platform.  Maybe you could add a link?

https://github.com/ggerganov/llama.cpp/blob/master/docs/build.md

Vaibhavs10:
Yes! great suggestion @dnouri - I added the suggestions here in line 47:

```
or [build directly from the source](https://github.com/ggerganov/llama.cpp/blob/master/docs/build.md) for your target device:
```

nsarrazin (approved):
Looks great thanks!

Issue #1387, opened 2024-08-06
CopyToClipBoardBtn in ChatMessage.svelte has a bug?


calycekr:
https://github.com/huggingface/chat-ui/blob/6de97af071c69aa16e8f893adebb46f86bdeeaff/src/lib/components/chat/ChatMessage.svelte#L378-L384

When compared to other components, classNames is the only difference here.
When rendered, the icon appears faint in the browser.
Is there a reason for this, or is it a bug?

https://github.com/huggingface/chat-ui/blob/6de97af071c69aa16e8f893adebb46f86bdeeaff/src/lib/components/CopyToClipBoardBtn.svelte#L37-L51

It seems that the classNames of IconCopy is the cause of the faintness.

nsarrazin:
Yep clearly there's something off with the button styling, if someone wants to tackle it feel free!

evalstate:
OK, happy to submit a PR for this. 
Changes are to CopyToClipBoardBtn.svelte to set icon height consistently, and use prop passing for the style classes.

**BEFORE FIX**:
![image](https://github.com/user-attachments/assets/19709f8c-857b-4ef3-adbb-ad589af80c45)
![image](https://github.com/user-attachments/assets/5ac47d6d-55a3-4dca-8476-5bd81fe30542)


**AFTER FIX**:
Normal Dark:
![image](https://github.com/user-attachments/assets/e636cd56-8996-4cff-9726-9bc1d459214e)
Normal Light:
![image](https://github.com/user-attachments/assets/87b01340-85b8-4987-8584-5bf9897032c1)

Hover Dark:
![image](https://github.com/user-attachments/assets/4054ab5f-a3d7-4e36-a8ee-7d0a3cd1555b)
Hover Light:
![image](https://github.com/user-attachments/assets/599830d9-f63d-406e-b927-dc06db4d0ee1)

nsarrazin:
Great thanks for tackling this! Feel free to open a PR and i'll take a look

evalstate:
Can't create a PR; fork is here: https://github.com/barrensconsulting/llmindset-hug-ui/tree/fix/copy-button-styling

nsarrazin:
Closing since this is solved, thanks @evalstate !

PR #1403, opened 2024-08-11
Fix "Copy" button styling. Relates to issue 1387 

evalstate:
Fix to copy button styling identified in Issue #1387 . Copy button under "ChatMessage" now consistent with other Icons. The component is also used as an icon in "CodeBlock.svelte". The styling for CodeBlock has been kept as it was.

PR recreated due to user error.


nsarrazin (approved):
Looking great in local testing, thanks for the fix!

evalstate:
More than welcome - I left the CodeBlock styling as it was - the background for CodeBlocks are always dark, but there is a small change in tone between light/dark modes.

Issue #926, opened 2024-03-12
AWS credentials resolution for Sagemaker models


nason:
chat-ui is excellent, thanks for all your amazing work here!

I have been experimenting with a model in Sagemaker and am having some issues with the model endpoint configuration. It currently requires credentials to be provided explicitly. This does work, but the ergonomics are not great for our use cases:
- in development, my team uses AWS SSO and it would be great to use our session credentials and not need to update our MODELS environment variable manually every time our sessions refresh
- in deployments, we would want to use an instance or task execution role to sign requests

In my investigation I found this area of code https://github.com/huggingface/chat-ui/blob/eb071be4c938b0a2cf2e89a152d68305d4714949/src/lib/server/endpoints/aws/endpointAws.ts#L22-L37, which uses the `aws4fetch` library that only support signing with explicitly passed AWS credentials.

I was able to update this area of code locally and support AWS credential resolution by switching this to use a different library [`aws-sigv4-fetch`](https://github.com/zirkelc/aws-sigv4-fetch) like so:

```ts
try {
	createSignedFetcher = (await import("aws-sigv4-fetch")).createSignedFetcher;
} catch (e) {
	throw new Error("Failed to import aws-sigv4-fetch");
}

const { url, accessKey, secretKey, sessionToken, model, region, service } =
	endpointAwsParametersSchema.parse(input);

const signedFetch = createSignedFetcher({
	service,
	region,
	credentials:
		accessKey && secretKey
			? { accessKeyId: accessKey, secretAccessKey: secretKey, sessionToken }
			: undefined,
});

// Replacer `aws.fetch` with `signedFetch` below when passing `fetch` to `textGenerationStream#options`
```

My testing has found this supports passing credentials like today, or letting the AWS SDK resolve them through the default chain.

Would you be open to a PR with this change? Or is there a different/better/more suitable way to accomplish AWS credential resolution here?

nsarrazin:
Yes I think that would be a great addition, especially since it's an added feature that doesn't require breaking changes 😄 Feel free to open a PR and I'll have a look!

PR #1419, opened 2024-08-20
Dynamic credential resolution for AWS endpoints

nason:
If AWS credentials are provided, they will continue to be used. With this change, they become optional parameters and the AWS SDK will attempt to use the default resolution chain if they are not provided. More details on the default resolution behavior can be found at https://github.com/zirkelc/aws-sigv4-fetch?tab=readme-ov-file#usage and https://www.npmjs.com/package/@aws-sdk/credential-provider-node

See https://github.com/huggingface/chat-ui/issues/926 for more context (and sorry for the delay in opening this PR!), I've been using this patch for a while now and have been able to utilize dynamic and temporary keys instead of static credentials to invoke my sagemaker endpoints


src/lib/server/endpoints/aws/endpointAws.ts
```diff
@@ -8,32 +8,45 @@ export const endpointAwsParametersSchema = z.object({
...
+	secretKey: z
+		.string({
+			description:
+				"An AWS Access Key Secret. If not provided, the default AWS identity resolution will be used",
+		})
+		.min(1)
+		.optional(),
 	sessionToken: z.string().optional(),
 	service: z.union([z.literal("sagemaker"), z.literal("lambda")]).default("sagemaker"),
-	region: z.string().optional(),
```
nsarrazin, line 15: making the `region` required could be a breaking change for some existing configs no? what is the reasoning for this change? 👀

nason: Great catch, I did not mean to commit this -- just an artifact from my testing 🌵

Re-added the .optional()

PR #1420, opened 2024-08-20
Websearch using Bing API

harisyammnv:
I have created this PR if someone prefers using BING search
Please let me know if there are any issues that i have overlooked in the current implementation

Thanks

### DIFF
--- a/.env
+++ b/.env
@@ -27,7 +27,6 @@ SERPSTACK_API_KEY=#your serpstack api key here
 SEARCHAPI_KEY=#your searchapi api key here
 USE_LOCAL_WEBSEARCH=#set to true to parse google results yourself, overrides other API keys
 SEARXNG_QUERY_URL=# where '<query>' will be replaced with query keywords see https://docs.searxng.org/dev/search_api.html eg https://searxng.yourdomain.com/search?q=<query>&engines=duckduckgo,google&format=json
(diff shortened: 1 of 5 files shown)


.env
```diff
@@ -27,6 +27,8 @@ SERPSTACK_API_KEY=#your serpstack api key here
 SEARCHAPI_KEY=#your searchapi api key here
 USE_LOCAL_WEBSEARCH=#set to true to parse google results yourself, overrides other API keys
 SEARXNG_QUERY_URL=# where '<query>' will be replaced with query keywords see https://docs.searxng.org/dev/search_api.html eg https://searxng.yourdomain.com/search?q=<query>&engines=duckduckgo,google&format=json
+BING_SEARCH_URL=#https://api.bing.microsoft.com/v7.0/search
```
nsarrazin, line 30: are there a lot of 3rd party search providers that use the bing API format? if it's only bing i'd prefer keeping only the `BING_SUBSCRIPTION_KEY` and hardcoding the `BING_SEARCH_URL` to avoid top-level env var clutter

harisyammnv: sure it is the same url for everyone i can remove that and hardcode it

nsarrazin: thanks!

harisyammnv: updated it

nsarrazin (approved):
Thanks for the contrib!

PR #1447, opened 2024-09-03
feat(tools): Basic tool support for OpenAI models

gladish:


harisyammnv:
really helpful

nsarrazin (commented):
Type checks seems to fail with 

```
Property 'correlationKey' is missing in type '{ name: string; parameters: Record<string, string>; }' but required in type 'ToolCall'.ts(2741)
```

could you take a look ?

nsarrazin (approved):
Thanks for the PR! I fixed a couple things and made it a bit more robust but I believe it's ready to go now. I'll merge but if you see issues with this please let me know :rocket:

PR #1448, opened 2024-09-04
Move Google GenAI API key to env file instead

altbdoor:
- Updated the documentation with the latest links from Google AI Studio
- Fixed the incorrect documentation on `safetyThreshold`
- Fixed the `safetySettings` to exclude unspecified category, as it is not considered as a valid input category in Google's API


nsarrazin (approved):
Appreciate the change! It's much nicer to have the API key in a separate variable :rocket:

PR #1469, opened 2024-09-13
pluralize

julien-c:


nsarrazin (approved):
thanks for the catch! :smile:

julien-c:
(you could steal a helper function for this in the Hub's code btw)

Issue #1471, opened 2024-09-16
Remap enter button in mobile keyboards to add new line


alr86:
I can't go to the new line in HuggingChat with my phone, and when I click on the enter button, the message is sent.
It's not the same as other chatbots.

PR #1475, opened 2024-09-18
Added enter to insert new line on mobile instead of default.

natecard:
Fix [#1471](https://github.com/huggingface/chat-ui/issues/1471) for mobile users to add a new line to their messages.

### DIFF
--- a/src/lib/components/chat/ChatInput.svelte
+++ b/src/lib/components/chat/ChatInput.svelte
@@ -19,16 +19,27 @@
 	function handleKeydown(event: KeyboardEvent) {
 		// submit on enter
 		if (event.key === "Enter" && !event.shiftKey && !isCompositionOn) {
-			event.preventDefault();
-			// blur to close keyboard on mobile
-			textareaElement.blur();
-			// refocus so that user on desktop can start typing without needing to reclick on textarea
-			if (isDesktop(window)) {
-				textareaElement.focus();
+			//  using isDesktop to check if the user is on a mobile device
+			if (!isDesktop(window)) {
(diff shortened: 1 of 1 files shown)


nsarrazin:
Hey, thanks a lot for opening a PR for this. Will review it!

src/lib/components/chat/ChatInput.svelte
```diff
@@ -19,14 +19,24 @@
...
+				const end = textareaElement.selectionEnd;
+				value = value.substring(0, start) + "\n" + value.substring(end);
+				textareaElement.selectionStart = textareaElement.selectionEnd = start + 1;
+			} else {
+				event.preventDefault();
+				// blur to close keyboard on mobile
+				textareaElement.blur();
+				// refocus so that user on desktop can start typing without needing to reclick on textarea
+				if (isDesktop(window)) {
+					textareaElement.focus();
```
nsarrazin, line 36: I think that if statement doesn't need to be there anymore right ? since you're already in the `else` condition here

nsarrazin (commented):
Left a small comment but otherwise should be good!

src/lib/components/chat/ChatInput.svelte
natecard: Thanks, you are right I did not catch removing that. Fixed and updated.

nsarrazin (approved):
I modified a few things, mostly i'm now detecting virtual keyboards instead of relying on viewport width. This should help detecting touchscreens with wider screens as well, other than that lgtm! 🚀

PR #1487, opened 2024-09-25
Copy to clipboard Error solve

KingNish24:
On google chrome and edge [As i tested on both of them]. The share URL is not copying to clipboard. So, this PR resolves that issue.

### DIFF
--- a/src/lib/shareConversation.ts
+++ b/src/lib/shareConversation.ts
@@ -4,30 +4,43 @@ import { share } from "./utils/share";
 import { page } from "$app/stores";
 import { get } from "svelte/store";
 import { getShareUrl } from "./utils/getShareUrl";
+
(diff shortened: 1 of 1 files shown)


nsarrazin:
PR didn't work for me on chrome in local testing 😅 I ended up fixing it here: https://github.com/huggingface/chat-ui/pull/1497

Currently deploying to production, let me know if it works there for you otherwise i can take another look

KingNish24:
> PR didn't work for me on chrome in local testing 😅 I ended up fixing it here: #1497
> 
> Currently deploying to production, let me know if it works there for you otherwise i can take another look

Your PR worked for me as well. 
Perhaps mine didn't work for you due to maybe different version of Chrome. 😅

KingNish24:
@nsarrazin Thanks for making it better.

Issue #1404, opened 2024-08-12
Cohere endpoint type doesn't support tools


nsarrazin:
## Bug description

The cohere endpoint type doesn't work well with our current tools setup and errors with `cannot specify both message and tool_results in multistep mode`

## Steps to reproduce

1. Set a `.env.local` with a model using cohere endpoint type, and set `tools: true` on that model.
2. Start a conversation with tool enabled

### Logs

```
 Error: {"message":"invalid request: cannot specify both message and tool_results in multistep mode"}
at eval (C:/Users/achar/Documents/aditya/chat-ui/src/lib/server/endpoints/cohere/endpointCohere.ts:93:17)
at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
at async eval (C:/Users/achar/Documents/aditya/chat-ui/src/lib/server/endpoints/cohere/endpointCohere.ts:70:18)
at async Module.generate (C:/Users/achar/Documents/aditya/chat-ui/src/lib/server/textGeneration/generate.ts:8:20)
at async textGenerationWithoutTitle (C:/Users/achar/Documents/aditya/chat-ui/src/lib/server/textGeneration/index.ts:54:3)
at async Module.mergeAsyncGenerators (C:/Users/achar/Documents/aditya/chat-ui/src/lib/utils/mergeAsyncGenerators.ts:13:34)
at async Module.textGeneration (C:/Users/achar/Documents/aditya/chat-ui/src/lib/server/textGeneration/index.ts:24:3)
at async Object.start (C:/Users/achar/Documents/aditya/chat-ui/src/routes/conversation/[id]/+server.ts:329:26) {
[cause]: BadRequestError: BadRequestError
Status code: 400
Body: {}
at CohereClient. (C:\Users\achar\Documents\aditya\chat-ui\node_modules\cohere-ai\Client.js:112:31)
at Generator.next ()
at fulfilled (C:\Users\achar\Documents\aditya\chat-ui\node_modules\cohere-ai\Client.js:31:58)
at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
statusCode: 400,
body: ReadableStream { locked: false, state: 'closed', supportsBYOB: false }
```

### Config

```
MODELS=`[
  {
    "name": "command-r-plus",
    "tools": true,
    "displayName": "Command R+",
    "endpoints": [{
      "type": "cohere",
    }]
  }
]`
```

PR #1490, opened 2024-09-25
Fix: Cohere Endpoint - Add Support for Tools

dvirmor:
This pull request addresses the issue described in https://github.com/huggingface/chat-ui/issues/1404#issue-2460503883, where attempting to use tools with the Cohere endpoint results in the following error: `cannot specify both message and tool_results in multistep mode`.

Since the Chat UI project handles the tool usage process, the multistep feature provided by the Cohere endpoint is unnecessary in this context. Therefore, I've modified the implementation to enforce a single API call, which resolves the issue.

**Tested on:**
- **OS**: Ubuntu 22.04 
- **Node.js Version**: v22.9.0

### DIFF
--- a/src/lib/server/endpoints/cohere/endpointCohere.ts
+++ b/src/lib/server/endpoints/cohere/endpointCohere.ts
@@ -62,7 +62,7 @@ export async function endpointCohere(
 				});
 
 				stream = await cohere.chatStream({
-					force_single_step: true,
+					forceSingleStep: true,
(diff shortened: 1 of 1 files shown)


dvirmor:
Hi @nsarrazin,
I just fix the linting.
In addition, I run `npm run test`, `npm run check` and `npm run lint` to be sure the workflow will pass

nsarrazin (approved):

PR #1496, opened 2024-09-27
Add support for model version in Vertex AI

pocman:
Bumping the version of Gemini from gemini-1.5-pro-001 to gemini-1.5-pro-002 should be completly transparent for the user.
Right now, this is not the case as the Gemini model version is strongly coupled with the HuggingChat model.

This PR add supports for `extraBody.model_version` in endpointVertexParametersSchema


src/lib/server/endpoints/google/endpointVertex.ts
```diff
@@ -16,6 +16,7 @@ export const endpointVertexParametersSchema = z.object({
 	model: z.any(), // allow optional and validate against emptiness
 	type: z.literal("vertex"),
 	location: z.string().default("europe-west1"),
+	extraBody: z.record(z.any()).optional(),
```
ArthurGoupil, line 19: nit but if we're able to strictly type it, lets do it
```suggestion
	extraBody: z.object({ model_version: z.string() }).optional(),
```

ArthurGoupil (approved):

nsarrazin (approved):
Thanks for the contrib! 

Long term we probably want to move the `extraBody` and `defaultHeaders` arguments out from endpoint-specific implementations and make it accessible to all endpoint types but that shouldn't be a blocker for this PR 🚀

pocman:
@nsarrazin I'm not sure I can merge the PR. Could you do it when you believe it's good to go ?

Issue #1492, opened 2024-09-25
Scrollbar is not pressable at the right edge of the viewport


rotemdan:
## Bug description

When the the mouse is located at the right edge of the viewport, the scroll bar is not responsive to clicks (pressing the left mouse button would cause text selection instead, which is not expected behavior).

It requires slight positioning to the left to get it pressed.

This is an issue with many different websites, but I've found a simple way to solve it here.

## Steps to reproduce

1. Position the mouse at the right edge of the viewport
2. Try to press and drag the scrollbar

## Screen capture

![scrollbar-press-at-right-edge](https://github.com/user-attachments/assets/a4bebd6d-86d4-45f7-93fc-77e85965bcdb)

### Specs

- **OS**: Windows 11
- **Browser**: Brave Version 1.70.117 Chromium: 129.0.6668.59 (Official Build) (64-bit)
- **chat-ui commit**: `v0.9.2` (web)

### Temporary fix using a custom stylesheet

1. Install the [Stylus extension](https://chromewebstore.google.com/detail/stylus/clngdbkpkpeebahjckkjfobafhncgmne)

2. Apply this stylesheet to https://huggingface.co/chat/

```css
@-moz-document url-prefix("https://huggingface.co/chat/") {
    div.scrollbar-custom {
        margin-right: 0px !important;
    }
}
```

The stylesheet simply zeros the right margin for the element that includes the scrollbar. The margin is otherwise `4px`, which causes the problem.

### Screen capture after applying stylesheet

![scrollbar-press-at-right-edge-after-custom-stylesheet](https://github.com/user-attachments/assets/9d4e99d5-e3b7-4dae-a60e-e9dc1a93b3d9)

### Thanks

I've been frustrated by this issue for a while now. I'm glad I've found a temporary workaround. 

I would appreciate the issue being fixed for all users. Sometimes what looks like a minor UI fix can make a surprisingly large improvement in the overall user experience.

nsarrazin:
Hi, thanks for reporting an issue! Looks like you already have a fix, if you want to open it as a PR i'll happily review it :rocket:

rotemdan:
The workaround I gave basically overrides some CSS properties with a browser extension.

In order to fix this properly, it should modify the cause of why the margin is added in the first place.

The particular HTML class that adds the margin seems to be `.mr-1`:

![Screenshot_10](https://github.com/user-attachments/assets/155e267f-c9a4-49e6-b3ad-7d60510cec1b)

[...]

PR #1498, opened 2024-09-27
Remove right margin from custom scrollbar

rotemdan:
Here's a basic pull request for an attempt at fixing #1492

You can use it as a starting point. It's not tested (tried to run `npm run preview` locally but it failed due to lack of MongoDB, seems like getting this to work isn't trivial for me and it'd take some more time and effort than I have right now).

I simply removed the tailwind class `mr-1`, which means:

```css
.mr-1 {
  margin-right: .25rem;
}
```

From instances of `custom-scrollbar` in these source files:

`src/lib/components/chat/ChatWindow.svelte`
`src/routes/assistants/+page.svelte`
`src/routes/models/+page.svelte`
`src/routes/tools/+page.svelte`

You'll need to test it yourself to ensure it actually works.

### DIFF
--- a/src/lib/components/chat/ChatWindow.svelte
+++ b/src/lib/components/chat/ChatWindow.svelte
@@ -235,7 +235,7 @@
 		/>
 	{/if}
 	<div
-		class="scrollbar-custom mr-1 h-full overflow-y-auto"
+		class="scrollbar-custom h-full overflow-y-auto"
 		use:snapScrollToBottom={messages.length ? [...messages] : false}
 		bind:this={chatContainer}
 	>
--- a/src/routes/assistants/+page.svelte
+++ b/src/routes/assistants/+page.svelte
@@ -114,7 +114,7 @@
 	{/if}
 </svelte:head>
 
-<div class="scrollbar-custom mr-1 h-full overflow-y-auto py-12 max-sm:pt-8 md:py-24">
+<div class="scrollbar-custom h-full overflow-y-auto py-12 max-sm:pt-8 md:py-24">
 	<div class="pt-42 mx-auto flex flex-col px-5 xl:w-[60rem] 2xl:w-[64rem]">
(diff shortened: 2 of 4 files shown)


rotemdan:
I've managed to start it locally using a docker container for the MongoDB instance.

The local instance doesn't actually work (getting errors like `Authorization header is correct, but the token seems invalid`) but I could paste a large amount of text to the chat window to see how the scrollbars behave.

Based on what I've seen, it seems the change does apply correctly.

**Just note** that Chromium has an unrelated issue with scrollbars: on my PC, for a zoom level of 125%, the scrollbar isn't pressable at the right edge, but with other zoom levels (100%, 150%, etc.) it does work. I've seen this happen on many other websites. It could be some internal issue with rounding of fractional pixels, or something of that sort. I'm not sure if it's actually reported on the Chromium tracker.

rotemdan:
Issue has now returned. I'm not sure exactly when, but I think even yesterday it still worked correctly.

Looking at the updated source code, the files, where I applied changes to remove the mr-1 classes from tags, have changed significantly.

I see the `mr-1` and `mr-1.5` classes added in many different places. The code looks very different from when I made the changes previously.

Figuring out what changes to make to fix the issue was hard enough the last time. I don't know if I could go through it again. As a maintainer of many different open-source repositories myself, I'd never ask a user for a second pull request on an issue they correctly fixed earlier, if the reason it broke again is not related to them.

I reopened #1492.

As you can see, I personally really care about this particular behavior. I noticed it very quickly, and it's frustrating to have it back again.

rotemdan:
Turns out the return of the issue was actually a [Chromium bug I activated by accidentally changing to a 125% zoom level](https://github.com/huggingface/chat-ui/issues/1492#issuecomment-2507910042). It wasn't coming from the website. I closed the issue again.

PR #1500, opened 2024-09-30
feat: add content disposition to file endpoint

nsarrazin:
see https://huggingface.slack.com/archives/C03Q18WK18T/p1722639880087169

### DIFF
--- a/src/routes/conversation/[id]/output/[sha256]/+server.ts
+++ b/src/routes/conversation/[id]/output/[sha256]/+server.ts
@@ -47,6 +47,9 @@ export const GET: RequestHandler = async ({ locals, params }) => {
 			"Content-Type": mime ?? "application/octet-stream",
 			"Content-Security-Policy":
 				"default-src 'none'; script-src 'none'; style-src 'none'; sandbox;",
+			"Content-Disposition": `attachment; filename="${sha256.slice(0, 8)}.${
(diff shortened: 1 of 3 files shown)


src/routes/conversation/[id]/output/[sha256]/+server.ts
```diff
@@ -47,6 +47,9 @@ export const GET: RequestHandler = async ({ locals, params }) => {
 			"Content-Type": mime ?? "application/octet-stream",
 			"Content-Security-Policy":
 				"default-src 'none'; script-src 'none'; style-src 'none'; sandbox;",
+			"Content-Disposition": `attachment; filename="${sha256.slice(0, 8)}.${
+				mime?.split("/")[1] ?? "bin"
```
Comment (coyotte508) [inline src/routes/conversation/[id]/output/[sha256]/+server.ts:51]: I'm not sure this is accurate always, see https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types for extension/mime type matching

coyotte508 (approved):
The `mime-types` module has an `extension` helper, if there's not already a module we have that has something similar

src/routes/conversation/[id]/output/[sha256]/+server.ts
Comment (nsarrazin) [reply src/routes/conversation/[id]/output/[sha256]/+server.ts]: fixed it using `mime-types`

PR #1508, opened 2024-10-08
fix: Remove ; in Vertex AI example configuration in Readme.md and add model_version as config

SebSchroen:
Using the example configuration for multimodality in Vertex AI models lead to an error when building, I removed the ; which lead to the error. 

Additionally, the Vertex Endpoint uses the following arguments to create the endpoint for getGenerativeModel:

```
export function endpointVertex(input: z.input<typeof endpointVertexParametersSchema>): Endpoint {
	const { project, location, model, apiEndpoint, safetyThreshold, tools, multimodal, extraBody } =
		endpointVertexParametersSchema.parse(input);

...
		const generativeModel = vertex_ai.getGenerativeModel({

			model: extraBody?.model_version ?? model.id ?? model.name,
---
}
```

Using model.id > model.name is counter-intuitive if you have variations of the same model using different tools or data stores. I suggest utilizing the extraBody parameter to allow updating the model version separately as introduced by @pocman. The final example config then reads:

```
MODELS=`[
//...
    {
       "name": "gemini-1.5-pro",
       "displayName": "Vertex Gemini Pro 1.5",
       "multimodal": true,
       "endpoints" : [{
          "type": "vertex",
          "project": "abc-xyz",
          "location": "europe-west3",
          "extraBody": {
          "model_version": "gemini-1.5-pro-preview-0409",
          },

          // Optional
          "safetyThreshold": "BLOCK_MEDIUM_AND_ABOVE",
          "apiEndpoint": "", // alternative api endpoint url,
          "tools": [{
            "googleSearchRetrieval": {
              "disableAttribution": true
            }
          }],
          "multimodal": {
            "image": {
              "supportedMimeTypes": ["image/png", "image/jpeg", "image/webp"],
              "preferredMimeType": "image/png",
              "maxSizeInMB": 5,
              "maxWidth": 2000,
              "maxHeight": 1000,
            }
          }
       }]
     },
]`

```


nsarrazin (approved):
thanks for the fix!

PR #1510, opened 2024-10-10
feat: Update featured workflow to allow user to submit featured requests

nsarrazin:
## Why

Currently we review all tools & assistants without checking if the user wants it to be reviewed. This means that we waste a lot of time on private stuff, tools that don't work, etc.

With this, the user can now ask to be featured, which will set a flag on the assistant and trigger a slack notification. Then the admin has two buttons to approve or deny a request. Once a request is denied the user won't be able to retrigger the flow. Admins can at any time reset an assistant or a tool to private (to un-feature something).

[(slack link)](https://huggingface.slack.com/archives/C052EKHD1U0/p1728468364907289?thread_ts=1728464686.106719&cid=C052EKHD1U0)

## How

The  `featured: boolean` flag was replaced by `review: "PRIVATE"|"PENDING"|"APPROVED"|"DENIED"`, with the appropriate migrations for tools & assistants. 

Buttons were added for admins to deny/approve and reset featured status. A button was added for user to request to be featured, with a confirm native modal to avoid misclicks.

Slack notifications were added for featured requests.

### user
image
image

### admin

image

### DIFF
--- a/scripts/populate.ts
+++ b/scripts/populate.ts
@@ -20,6 +20,7 @@ import { Message } from "../src/lib/types/Message.ts";
 
 import { addChildren } from "../src/lib/utils/tree/addChildren.ts";
 import { generateSearchTokens } from "../src/lib/utils/searchTokens.ts";
+import { ReviewStatus } from "../src/lib/types/Review.ts";
 
 const rl = readline.createInterface({
 	input: process.stdin,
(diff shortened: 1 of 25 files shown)


nsarrazin:
cc @mishig25 I think the approval pipeline for assistant would have to be updated it should only run for assistants with  `review: "PENDING"` and it can set it to either `review: "DENIED"` or `review: "APPROVED"`

(it should also make things a bit simpler since you don't need to keep track of which assistants have already been checked and can just check which ones are pending)

PR #1511, opened 2024-10-10
feat(endpoints): Add conv ID to headers passed to TGI

nsarrazin:
Adds conv ID as a custom header `ChatUI-Conversation-ID`  when calling TGI for better caching


OlivierDehaene (approved):
Thanks!

PR #1538, opened 2024-10-22
fix(modal): fix freeze after modal close

neven4:
Quite often, my whole interface freezes after I close the modal, but with tools, it happens to me every time.

It seems like Svelte's issue with animation: https://github.com/sveltejs/kit/issues/5434

Steps to reproduce:
- click 'globe' icon in tools
- close modal
- interface will freeze

https://github.com/user-attachments/assets/f717687a-2f45-4c17-a057-9f6df73ec793


nsarrazin (approved):
Love it, thanks a lot for the fix! :rocket:

PR #1539, opened 2024-10-22
feat: Add a model config option to disable system prompts

### LINKED ISSUE #1432: Add a model config option to disable system prompts

Some models don't support system prompts and if you try to use the `system` role in their chat template they throw an error. We should add a config option in the model definition to specify that system prompts are not supported. In that case we 

## Implementation idea

add a `systemPrompt: true|false` in the `modelConfig` which will trigger logic to filter out `role: 'system'` messages and preprend the content of the first one to the first user message

Solves #1386

Mounayer:
## Description

Closes #1432 

Some models don't support system prompts, therefore, a new `modelConfig` option has been added: `systemRoleSupported `

By default, the `systemRoleSupported ` option is `true`, where everything behaves as is. However, if the `systemRoleSupported ` option is set to `false`, all `role: "system"` messages are filtered out. And the very first `role: "system"` message encountered, will have its content preprended to the very first user message.

Brief example, let's assume the following is our `messages` array, before processing:

```json
[
    { "content": "Welcome to our service!", "role": "system" },
    { "content": "Hi, how can I help you today?", "role": "system" },
    { "content": "I need help with my account.", "role": "user" },
    { "content": "Certainly! What issue are you experiencing?", "role": "system" },
    { "content": "I forgot my password.", "role": "user" },
    { "content": "Please reset your password using the 'Forgot Password' link.", "role": "system" },
    { "content": "Thank you!", "role": "user" }
]
```

After we apply the filtering, and preprending the first `system` message's content into the first user message, we get:

```json
[
    { "content": "Welcome to our service!\nI need help with my account.", "role": "user" },
    { "content": "I forgot my password.", "role": "user" },
    { "content": "Thank you!", "role": "user" }
]
```

This keeps the existing functionality untouched, while giving the option of using a wider range of models, even those that do not support `system` role messages!

Please let me know any thoughts or suggestions that may arise!

### DIFF
--- a/src/hooks.server.ts
+++ b/src/hooks.server.ts
@@ -19,6 +19,9 @@ import { refreshConversationStats } from "$lib/jobs/refresh-conversation-stats";
 
 // TODO: move this code on a started server hook, instead of using a "building" flag
 if (!building) {
+	// Set HF_TOKEN as a process variable for Transformers.JS to see it
+	process.env.HF_TOKEN ??= env.HF_TOKEN;
+
 	logger.info("Starting server...");
 	initExitHandler();
 
--- a/src/lib/server/models.ts
+++ b/src/lib/server/models.ts
@@ -68,7 +68,7 @@ const modelConfig = z.object({
 	unlisted: z.boolean().default(false),
 	embeddingModel: validateEmbeddingModelByName(embeddingModels).optional(),
 	/** Used to enable/disable system prompt usage */
-	systemPrompt: z.boolean().default(true),
+	systemRoleSupported: z.boolean().default(true),
 });
 
 const modelsRaw = z.array(modelConfig).parse(JSON5.parse(env.MODELS));
@@ -117,7 +117,7 @@ async function getChatPromptRender(
 			role: message.from,
 		}));
 
-		if (!m.systemPrompt) {
+		if (!m.systemRoleSupported) {
 			const firstSystemMessage = formattedMessages.find((msg) => msg.role === "system");
 			formattedMessages = formattedMessages.filter((msg) => msg.role !== "system");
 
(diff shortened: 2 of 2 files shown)


Mounayer:
> Thanks for the contribution! i think this is very much needed for some kind of models like gemma & the o1 models.
> 
> I think the name of the argument is maybe a bit confusing sorry I know I'm the one who wrote it in the initial issue 😬
> 
> I think people might assume it's a field to set a system prompt. What do you think of `m.systemRoleSupported` instead ?

Super happy to contribute!

Honestly, I agree. It can be a little confusing, I've updated the config option name to be `systemRoleSupported` instead. I think now it's more intuitive as to what it does!

nsarrazin (approved):
Thanks for the contribution!

burkhat:
@Mounayer I've tried this configuration parameter to use o1-mini within huggingface chat-ui.
I've used following image: ghcr.io/huggingface/chat-ui:0.9.4 and following model configuration:

`MODELS=[{
{
  "id": "o1-mini",
  "name": "o1-mini",
  "displayName": "o1-mini",
  "systemRoleSupported": false,
  "parameters": {
      "temperature": 0.3,
      "max_new_tokens": 4096,
  },
  "description": "OpenAI's flagship model. Deployed on Azure. With vision support.",
  "modelUrl": "https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/models?tabs=global-standard%2Cstandard-chat-completions#o-series-models",
  "websiteUrl": "https://openai.com/index/openai-o1-mini-advancing-cost-efficient-reasoning/",
  "logoUrl": "https://openai.com/favicon.ico",
  "tools": false,
  "multimodal" : true,
  "endpoints": [
      {
          "type": "openai",
          "baseURL": "https://<url>.openai.azure.com/openai/deployments/o1-mini",
          "defaultQuery": {
                  "api-version": "2024-08-01-preview"
              },
          "defaultHeaders": {
            "api-key": "${OPENAI_API_KEY}"
          },
      }
  ]
}]`

But with this parameter I still got following error message with o1-mini:
"400 Unsupported value: 'messages[0].role' does not support 'system' with this model"

Any hints how to fix it?

PR #1541, opened 2024-10-24
Add speech recognition to chat interface

flozi00:


nsarrazin (commented):
Love the idea of this PR, thanks for the contrib! I tried it a bit and I think it could be a great addition.

Couple points:
- We should check if the browser supports webgpu, I tried it on firefox and it failed silently initially because it's not supported by default. If that's the case maybe we should hide the icon?
- Would be nice to have some kind of download indicator for the transcription model, maybe a small download bar below the mic icon when you press it? Otherwise you get no feedback until it's done.
- We already have a few loading animations, do you think we could be using these instead? (`src/lib/components/icons/IconLoading.svelte` and `	import EosIconsLoading from "~icons/eos-icons/loading";`)
- Sometimes I stop recording and record again and it seems to transcribe the previous recording only. Overall this is super cool but maybe the audio chunks should be cleared when you start a new recording ?

flozi00:
@nsarrazin could you please take this PR ? I won't have time to finish it sadly

nsarrazin:
No worries thanks for starting it!

dagoatzmcclusky:
great job! It's a very great, must-hav feature in huggingchat. We need this PR to be merged asap!

PR #1544, opened 2024-10-25
[feat] Add support for endpoint web sources

ArthurGoupil:
Some endpoints like VertexUI provides web sources used to generate the message.

This PR add the basis for any endpoint to add sources. They just have to provide `webSource` in the endpoint `output`.
Vertex implementation is done as an example.

https://github.com/user-attachments/assets/93dc4a26-fc08-45e5-96dc-a47a1e4eda43


nsarrazin (approved):
Looks good to me! Vertex endpoint is 🔥 thanks to you 😁

PR #1118, opened 2024-05-07
Web Search: Inline citations

Saghen:
Continuation of the work on #1094, don't merge until that's merged! Adds support for inline citations by prefixing each source with `[[idx]]` and requesting that the model include that in its output. It then replaces any instances of these during runtime with a link to source in superscript.

image

PR #1551, opened 2024-11-05
feat: add sources for websearch

nsarrazin:
continuation of #1118, branch had a lot of conflicts

image


nsarrazin:
doesnt work well with tools, investigating. works well for models that dont have tools enabled

nsarrazin:
works well enough without tools, will merge for now, failure mode is usually just that it forgets to put citations so it's not really breaking

PR #1554, opened 2024-11-07
[fix] Add placeholder param for sources icon

ArthurGoupil:
For sources display, we use Google API (for instance `https://www.google.com/s2/favicons?sz=64&domain_url=github.com`). But when for some reason the hostname (`new URL(link).hostname`) is not well computed, we end up with a broken link `https://www.google.com/s2/favicons?sz=64&domain_url=` and the icon is replaced by the `alt` attributes, which can lead to very weird display:
image

So this fix sets the hostname to `placeholder` when not well computed `https://www.google.com/s2/favicons?sz=64&domain_url=placeholder`

### DIFF
--- a/src/lib/components/chat/ChatMessage.svelte
+++ b/src/lib/components/chat/ChatMessage.svelte
@@ -335,7 +335,8 @@
 						>
 							<img
 								class="h-3.5 w-3.5 rounded"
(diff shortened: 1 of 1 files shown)


pocman (approved):

nsarrazin (approved):
Nice, thanks for the fix

1atAlcone (approved):

Issue #1552, opened 2024-11-05
Repetition penalty passed as frequency penalty in OpenAI-compatible providers


tommasobonomo:
## Bug description

I'm setting the repetition penalty parameter to a set value $p$ on a vLLM-deployed model through chat-ui's `.env.local`. When inspecting vLLM's logs, I find that the repetition penalty is the default value and instead the frequency penalty is set to $p$, the original value I wanted for the repetition penalty. 

I found the cause in the codebase: https://github.com/huggingface/chat-ui/blob/d6dc46a3980478887a825d73a2041a44303e21c4/src/lib/server/endpoints/openai/endpointOai.ts#L169

According to [OpenAI's](https://platform.openai.com/docs/api-reference/chat/create#chat-create-frequency_penalty) and [vLLM's](https://docs.vllm.ai/en/latest/dev/sampling_params.html) docs, repetition and frequency penalties have a different scale, although they are conceptually similar. The first has a neutral default value of 1.0, whilst the second has a neutral value of 0.0.

This bug introduces a pretty hefty `frequency_penalty = 1.0` when the user has no intention of setting a penalty (`repetition_penalty = 1.0`).

## Steps to reproduce

Deploy a model through any OpenAI-compatible endpoint and set the parameter `repetition_penalty = 1.0` in the model parameters in `.env.local`.

### Config

```
MODELS=`[
    {
	"name":"HuggingFaceTB/SmolLM2-1.7B-Instruct",
        "endpoints": [
            {
                "type":"openai",
                "baseURL":"http://vllm_server:8000/v1",
                "completion":"completions"
            }
        ],
        "parameters": {
            "repetition_penalty": 1.0,
        },
    }
]`
```

## Notes

As noted above, the issue stems from this line in the codebase: https://github.com/huggingface/chat-ui/blob/d6dc46a3980478887a825d73a2041a44303e21c4/src/lib/server/endpoints/openai/endpointOai.ts#L169

Repetition and frequency penalty should probably be parsed separately by chat-ui.

PR #1557, opened 2024-11-11
Fix/oai parameters 1552

evalstate:
Fix for issue #1552  (incorrect setting of presence/frequency (repetition) penalties for OAI models).
Also allowed temperature up to 2 in zod (some models allow it).


nsarrazin (approved):
Thanks!

PR #1561, opened 2024-11-12
feat(migrations): clean up empty conversations from db

nsarrazin:
removes conversations that have no matching user or valid sessions

also removes old conversations that have no messages in them

### DIFF
--- a/src/lib/migrations/routines/09-delete-empty-conversations.ts
+++ b/src/lib/migrations/routines/09-delete-empty-conversations.ts
@@ -1,0 +1,47 @@
+import type { Migration } from ".";
+import { collections } from "$lib/server/database";
+import { ObjectId } from "mongodb";
+import { logger } from "$lib/server/logger";
+
+export async function deleteConversations(
+	collections: typeof import("$lib/server/database").collections
+) {
+	let deleteCount = 0;
+	const { conversations, sessions } = collections;
+
+	// First delete conversations with no messages that are older than 1 hour
+	const emptyResult = await conversations.deleteMany({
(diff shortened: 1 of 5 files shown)


nsarrazin:
cc @coyotte508 if you have some time, feel free to look at `src/lib/migrations/routines/09-delete-empty-conversations.ts` [(link)](https://github.com/huggingface/chat-ui/blob/04a111a53c935cb3b0e27cebd8bedb609d087664/src/lib/migrations/routines/09-delete-empty-conversations.ts)

There are some tests to make sure the migration works as intended too but wanted to check with you before committing, no rush though :smile:

src/lib/migrations/routines/09-delete-empty-conversations.ts
```diff
@@ -0,0 +1,77 @@
...
+const BATCH_SIZE = 1000;
+const DELETE_THRESHOLD_MS = 60 * 60 * 1000;
+
+export async function deleteConversations(
+	collections: typeof import("$lib/server/database").collections
+) {
+	let deleteCount = 0;
+	const { conversations, sessions } = collections;
+
+	const validSessionIdsArray = await sessions.distinct("sessionId");
```
coyotte508, line 15: Not sure how it's going to work with such a large collection. Especially if you want to include the huge array in the query afterwards.

You may be better off running that check manually - eg get conversations without users 1000 by 1000 and then do `db.dessions.find({_id: {$in: ...}})` and see which Ids are missing, and delete those conversations

nsarrazin: Oh thanks, yes that would make more sense.

PR #1564, opened 2024-11-14
feat: infinite scrolling list of conversations

nsarrazin:
should be working now! infinite loading 30 convs at a time


src/lib/components/InfiniteScroll.svelte
```diff
@@ -0,0 +1,47 @@
+<script lang="ts">
+	import { onMount } from "svelte";
+	import { createEventDispatcher } from "svelte";
```
gary149, line 3:
```suggestion
	import { onMount, createEventDispatcher } from "svelte";
```

Issue #1567, opened 2024-11-14
Support explicit configuration and discovery of JWS signing algorithm when RS256 is not used by the OP.


drewtoto:
## Describe your feature request

Support explicit configuration of `id_token_signed_response_alg` in the *openid-client* `issuer.Client` configuration.  Currently, the properties configured when instantiating a `issuer.Client` are:

From `src/lib/server/auth.ts`:
```
{
  client_id: ...,
  client_secret: ...,
  redirect_uris: ...,
  response_types: ...,  
  [custom.clock_tolerance]: ...,
}
```

## Screenshots (if relevant)

N/A

## Implementation idea

Pass configuration object containing `id_token_signed_response_alg` to `issuer.Client` in `getOIDCClient`.  Allow this property to be explicitly set via environment variable or obtained from `id_token_signing_alg_values_supported` field in the issuer metadata during OP discovery when `RS256` is not included by the OP.

Proposed data structure to pass to `issuer.Client`:
```
{
  client_id: ...,
  client_secret: ...,
  redirect_uris: ...,
  response_types: ...,  
  [custom.clock_tolerance]: ...,
  id_token_signed_response_alg: ...,
}
```

drewtoto:
Issue resolved by PR #1568.

PR #1568, opened 2024-11-14
feat: explicit config of id token alg (#1567)

drewtoto:
- support to configure `id_token_signed_response_alg` field in the configuration object passed to `issuer.Client` in `src/lib/server/auth.ts`.

- allow `id_token_signed_response_alg` to be set from environment variable `OIDConfig.ID_TOKEN_SIGNED_RESPONSE_ALG` or obtained via OP metadata during issuer discovery when `RS256` is not included.


nsarrazin:
Hi, out of curiosity why is this needed? Not opposed to adding it, but I'm interested in what the use case is.

drewtoto:
Hi nsarrazin,

I have a project using OIDC auth with an on-prem hosted Authentik (https://goauthentik.io/) as the IdP.   In our case, the provider in Authentik only supports **HS256** for `id_token_signing_alg_values_supported` (see `.well-known/openid-configuration` below).

I only discovered this situation after debugging **chat-ui** and finding an `RPError` thrown by the **openid-client** module (here https://github.com/panva/openid-client/blob/v5.6.5/lib/client.js#L911).

I could not find another way around this issue, and implementing this optional parameter solved my case.   If I missed something and this situation is solved another way, that would be great as well.  I don't know how widespread this case is, but I did not find very much related to it, so I thought it might be worth sharing.


`.well-known/openid-configuration`:
```
{
  "issuer": "http://idp.site.com:9000/application/o/myapp/",
  "authorization_endpoint": "http://idp.site.com:9000/application/o/authorize/",
  "token_endpoint": "http://idp.site.com:9000/application/o/token/",
  "userinfo_endpoint": "http://idp.site.com:9000/application/o/userinfo/",
  "end_session_endpoint": "http://idp.site.com:9000/application/o/myapp/end-session/",
  "introspection_endpoint": "http://idp.site.com:9000/application/o/introspect/",
  "revocation_endpoint": "http://idp.site.com:9000/application/o/revoke/",
  "device_authorization_endpoint": "http://idp.site.com:9000/application/o/device/",
  "response_types_supported": [
    "code",
    "id_token",
    "id_token token",
    "code token",
    "code id_token",
    "code id_token token"
  ],
  "response_modes_supported": [
    "query",
    "fragment",
    "form_post"
  ],
  "jwks_uri": "http://idp.site.com:9000/application/o/myapp/jwks/",
  "grant_types_supported": [
    "authorization_code",
    "refresh_token",
    "implicit",
    "client_credentials",
    "password",
    "urn:ietf:params:oauth:grant-type:device_code"
  ],
  "id_token_signing_alg_values_supported": [
    "HS256"
  ],
  "subject_types_supported": [
    "public"
  ],
  "token_endpoint_auth_methods_supported": [
    "client_secret_post",
    "client_secret_basic"
  ],
  "acr_values_supported": [
    "goauthentik.io/providers/oauth2/default"
  ],
  "scopes_supported": [
    "openid",
    "email",
    "profile",
    "offline_access"
  ],
  "request_parameter_supported": false,
  "claims_supported": [
    "sub",
    "iss",
    "aud",
    "exp",
    "iat",
    "auth_time",
    "acr",
    "amr",
    "nonce",
    "email",
    "email_verified",
    "name",
    "given_name",
    "preferred_username",
    "nickname",
    "groups"
  ],
  "claims_parameter_supported": false,
  "code_challenge_methods_supported": [
    "plain",
    "S256"
  ]
}

```

nsarrazin (approved):
Hi :wave: sorry for the delay on this, I updated the code a bit to ensure it passes all checks. Let me know if it still works for you!

I also made sure to add `ID_TOKEN_SIGNED_RESPONSE_ALG` in the zod object. For now it's an optional string. Do you think we could narrow it down to some literals option ? Not super familiar with the spec so let me know if you think it's possible.

drewtoto:
Yes, this looks great.  I think the optional string works fine, as I suspect most users won't even need to worry about this, and when a user does, they'll be familiar enough to know which algorithms are supported.  However, if your preference is to enumerate validate algorithm options, that should be possible as well.  I'll provide some info here for you to review so you can decide how you want to proceed. Below are also a few links to spec docs that go directly to relevant sections if you (or others reading this) are interested.

Also, Golang *go-jose* project has implemented signature constants like [this](https://github.com/go-jose/go-jose/blob/15bc4c2ac4575ad865f078390db61d44530f985d/shared.go#L105) (string constants), which can be used as a reference.

- [OpenID Authorization Code Flow (ID Token section)](https://openid.net/specs/openid-connect-core-1_0.html#CodeIDToken)
- [OpenID Implicit Flow (ID Token section)](https://openid.net/specs/openid-connect-core-1_0.html#ImplicitIDToken)
- [JWT (RFC 7519 section 8) requirements](https://www.rfc-editor.org/rfc/rfc7519.html#section-8)
- [JOSE signature and encryption algorithms](https://www.iana.org/assignments/jose/jose.xhtml#web-signature-encryption-algorithms)

nsarrazin:
Hi! I think it's fine like this, like you said if users change this we can expect them to know what they're doing haha

Looks good to me, merging now :rocket:

PR #1571, opened 2024-11-18
Anthropic PDF Beta Support

evalstate:
Anthropic PDF support (beta) : https://docs.anthropic.com/en/docs/build-with-claude/pdf-support

- Bump Anthropic SDK to latest version
- To enable support (Sonnet 3.5 only), update model configuration: 

```
      "multimodalAcceptedMimetypes": ["image/*","application/pdf"],
      "endpoints": [
        {
          "defaultHeaders": {"anthropic-beta": "pdfs-2024-09-25"},
```


nsarrazin:
Thanks for the contrib @evalstate ! Ideally chat-ui endpoints that support PDF would support more than one message having PDFs.

Do you think this is something that can be resolved on Anthropic side or is this just the expected behaviour?

evalstate:
I'm 99% sure it's an error on their side: The API throws a hard 500 Internal Server Error rather than a more graceful failure. There's nothing in the docs that indicate this would be unsupported (and handling multiple PDFs in one message anywhere in the conversation is fine too).

I've also updated locally to use the BetaBase64PDFBlock from the "Beta" types in their API.

Their own claude.ai front-end doesn't suffer from this, but I think their file handling for the app goes through a different path (and supported PDFs long before the public beta).

Shall I leave this in draft for the moment in case anyone wants it while I find out more?

nsarrazin:
Yes feel free to come back once you know more! If this is intended behaviour we'll figure out something but would be nice if this was supported indeed.

evalstate:
The Anthropic API no longer throws a 500 Error in this scenario, so proposing merge. Currently Claude doesn't seem to "see" PDFs that are in later messages, raised via the Anthropic feedback route as I expect this will get fixed their end.

nsarrazin (approved):
Thanks for the contribution!

evalstate:
This feature is now out of beta. I've just tested, and you no longer need the `anthropic-beta` beta. The issue with Claude not "seeing" PDFs in follow-on messages is also fixed (at Anthropics end).

```
      "multimodalAcceptedMimetypes": ["image/*","application/pdf"],
```

PR #1576, opened 2024-11-19
fix: docker build

nsarrazin:


gianpaj:
@nsarrazin are you sure it's all correct?

I see this args twice
```dockerfile
ARG APP_BASE=
ARG PUBLIC_APP_COLOR=blue
ENV BODY_SIZE_LIMIT=15728640
```

PR #1582, opened 2024-11-22
feat(huggingchat): add tools support for Qwen 2.5 72B

nsarrazin:
Also refactors the tool parsing slightly to handle tool call signatures with different naming schemes


nsarrazin:
Added already

PR #1585, opened 2024-11-23
Small fix for Scroll Previous Button

evalstate:
One liner to update the query selector for the Scroll to Previous button


nsarrazin (approved):
Woops my bad, broke it when I changed the data attributes, thanks for the fix!

PR #1586, opened 2024-11-23
fix: double cookie refresh in server hooks

evalstate:
refreshSessionCookie() is called twice for POST requests. 
Removed the call which sends the _server_ id to the client.


nsarrazin (approved):
Appreciate the fix, looks like this file could be cleaned up a bit :sweat_smile:

Issue #1599, opened 2024-11-28
LaTeX is wrongly inserted inside inline code blocks


BafS:
## Bug description

LaTeX (KaTeX) is inserted inside inline code when `$` is used.

## Steps to reproduce

- Use any model
- "write a quick sort in php"

## Screenshots

<img width="704" alt="image" src="https://github.com/user-attachments/assets/24ba78db-15ac-4489-8a43-aeb001ba5671">

### Logs

(nothing interesting)

### Specs

- **OS**: Macos
- **Browser**: Firefox
- **chat-ui commit**: v0.9.4

nsarrazin:
Hi, thanks for the report, I can reproduce. Let me see what I can do!

PR #1601, opened 2024-11-28
Feat/new_renderer_markdown

nsarrazin:


nsarrazin:
I skipped the failing tests for now, they cover edge cases we do not support yet anyway (like #1599) , they will need to be supported but I want the testing setup merged :smile:

PR #1604, opened 2024-11-29
Capture screen like similar to Claude

harisyammnv:
I tried to get a simpler version of Screen capture similar to the one in Claude here
@nsarrazin : your comments would be appreciated here


nsarrazin (approved):
Hi @harisyammnv I merged the latest changes to the tool UI to your branch and updated a few things to make it work there. 

The icon now shows whenever we support uploading image MIME types. So tools that accept images and multimodal models are both supported.

cc @gary149 for review, if we want to add this in prod or not

src/lib/components/icons/iconScreenshot.svelte
gary149, line 1: Capitalized: `IconScreenshot.svelte`

src/lib/components/chat/ChatInput.svelte
```diff
@@ -250,6 +252,27 @@
 						</label>
 					</HoverTooltip>
 				</form>
+				{#if mimeTypes.includes("image/*")}
+					<HoverTooltip label="Capture screenshot" position="top">
```
gary149, line 256: positionning could be better: 
image

src/lib/components/chat/ChatWindow.svelte
```diff
@@ -414,65 +414,67 @@
 					</div>
 				{/if}
 			</div>
-			<form
-				tabindex="-1"
-				aria-label={isFileUploadEnabled ? "file dropzone" : undefined}
-				on:submit|preventDefault={handleSubmit}
-				class="relative flex w-full max-w-4xl flex-1 items-center rounded-xl border bg-gray-100 focus-within:border-gray-300 dark:border-gray-600 dark:bg-gray-700 dark:focus-within:border-gray-500
+			<div class="flex w-full items-center gap-2">
+				<form
```
gary149, line 418: did test extensively but I would prefer not changing the elements wrapping here.

nsarrazin: yeah not sure why that was there, reverted the changes in that file!

src/lib/components/chat/ChatInput.svelte
nsarrazin: fixed

src/lib/utils/screenshot.ts
```diff
@@ -0,0 +1,37 @@
...
+		// Stop all tracks
+		stream.getTracks().forEach((track) => track.stop());
+
+		// Convert to base64
+		return canvas.toDataURL("image/png");
+	} catch (error) {
+		console.error("Error capturing screenshot:", error);
+		throw error;
+	}
+}
```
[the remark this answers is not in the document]
nsarrazin: had to move stuff around since we defined stream in the `try` block but should be good now

PR #1610, opened 2024-12-02
Remove unneeded QwQ System prompt

mishig25:
Removing unneeded QwQ System prompt as HF Inference API automatically adds the default system prompt as defined in [chat_template](https://huggingface.co/Qwen/QwQ-32B-Preview/blob/main/tokenizer_config.json#L197)

### DIFF
--- a/chart/env/prod.yaml
+++ b/chart/env/prod.yaml
@@ -139,7 +139,7 @@ envVars:
       },
       {
         "name": "Qwen/QwQ-32B-Preview",
-        "preprompt": "You are a helpful and harmless assistant. You are Qwen developed by Alibaba. You should think step-by-step.",
+        "preprompt": "",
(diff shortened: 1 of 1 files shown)


chart/env/prod.yaml
```diff
@@ -139,7 +139,7 @@ envVars:
       },
       {
         "name": "Qwen/QwQ-32B-Preview",
-        "preprompt": "You are a helpful and harmless assistant. You are Qwen developed by Alibaba. You should think step-by-step.",
+        "preprompt": "",
```
mishig25, line 142: @nsarrazin or can I remove this field entirely?

nsarrazin: You can just remove it entirely!

nsarrazin (approved):

PR #1618, opened 2024-12-07
feat(openai): added support for o1 reasoning models

KeganHollern:
## Summary

1. Fixing a bug with `systemRoleSupported` for the OpenAI endpoint. 
2. Now using `max_completion_tokens` over `max_tokens` when querying `chat_completions` endpoint.

The bugfix is necessary as `o1-preview` and `o1-mini` models do not support "system" message types.
The token configuration change was necessary as `o1-preview` and `o1-mini` no longer support the, _now deprecated_, `max_tokens` field.

## Considerations

I added a new endpoint configuration field called `useCompletionTokens` and defaulted it to `true`. 
This would have the effect of changing the default `chat_completions` call to use `max_completion_tokens` over `max_tokens`.
There may be unintended consequences for users, as this could increase their overall token usage per query (as the default would change). 
It may be prudent to start with this field set to `false`, and later switch to `true` as part of a planned deprecate of `max_tokens` all together. 

Another consideration is that my change for system messages will leave a `"role": "user", "content": ""` entry. One thing we may want is to drop the system message entirely if content is blank  and system messages are not supported.

## Reference

For those wanting to introduce o1 reasoning models based on this PR (future readers), here is an example implementation of `o1-mini`:

```json
{
      "name": "o1-mini",
      "displayName": "o1 mini test",
      "multimodal": false,
      "description": "ChatGPT o1-mini",
      "systemRoleSupported": false,
      "parameters": {
        "max_new_tokens": 2048,
      },
      "endpoints" : [{
        "type": "openai",
      }]
}
```

### DIFF
--- a/docs/source/configuration/models/providers/openai.md
+++ b/docs/source/configuration/models/providers/openai.md
@@ -53,6 +53,25 @@ MODELS=`[{
 }]`
 ```
 
+We also support models in the `o1` family. You need to add a few more options ot the config: Here is an example for `o1-mini`:
+
+```ini
+MODELS=`[
+  {
+      "name": "o1-mini",
+      "description": "ChatGPT o1-mini",
+      "systemRoleSupported": false,
+      "parameters": {
+        "max_new_tokens": 2048,
+      },
+      "endpoints" : [{
+        "type": "openai",
+        "useCompletionTokens": true,
+      }]
+  }
+]
+```
+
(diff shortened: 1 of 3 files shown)


evalstate:
this is cool - i had a crack at this a while ago ( https://github.com/huggingface/chat-ui/discussions/1540#discussioncomment-11208660 ) at which point streaming wasn't supported so i had to do some extra messing around (i capture token usage too).

nsarrazin:
Hi! Thanks for the contrib. I made a couple changes:
- Added an example in the doc
- Made the parameter false by default (like you said, probably better to not change the default behaviour for now)
- Fixed the linting

Overall this looks great, tried it locally and it seems to work well. Will merge once the checks are done. :rocket:

PR #1629, opened 2024-12-16
Add support for Amazon Nova

tarekbecker:
This pull requests adds support for Amazon Nova models. The API is slightly different. No need to specify the `type` etc: https://docs.aws.amazon.com/nova/latest/userguide/complete-request-schema.html

In addition, it fixes the hard coded "image/jpeg" to use the mime type from the file processor.


nsarrazin:
Hi! Thanks for the contribution. I took the liberty of removing some `console.log` that were left in the changed code. I've also replaced the implicit check to see if `model.id` contains nova with a more explicit `isNova` flag.

Let me know if those changes still work for you and if so LGTM! :rocket:

nsarrazin (approved):

tarekbecker:
LGTM. Thanks for the fixes.

PR #1648, opened 2025-01-14
Added currentDate variable to Dynamic Prompt

evalstate:
Allow the use of {{currentDate}} in an assistant Dynamic Prompt, which inserts the date in the format "Tuesday, 14 January 2025".

Not sure if anyone else wants this? - and implementation could probably be more robust (e.g. able to use from normal system prompts, not require ASSISTANT_RAG). Happy to take guidance on improving implementation.

### DIFF
--- a/src/lib/components/AssistantSettings.svelte
+++ b/src/lib/components/AssistantSettings.svelte
@@ -96,7 +96,7 @@
 		: false;
 
 	let tools = assistant?.tools ?? [];
-	const regex = /{{\s?(get|post|url|currentDate)(=.*?)?\s?}}/g;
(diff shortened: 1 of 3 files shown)


nsarrazin:
Hi, I think this is a handy addition, I'll test it to be sure it works well but I like the idea.

Maybe long-term I'll move the preprompt processing from being assistant specific to a more generic spot indeed so it can be used with model system prompts as well, that would be nice.

nsarrazin (approved):
Renamed the tag to {{today}} but other than that the feature works great, thanks for the contrib!

PR #1670, opened 2025-01-29
Rm unused prop

mishig25:
Removes the unused `model` prop from `ChatMessage`. Please let me know if it is correct


nsarrazin (approved):
ha nice, thanks for the catch

PR #1686, opened 2025-02-07
fix: bedrock prepareMessages parameters

lgarbarini:
#1629 accidentally broke non-Nova models, fixes the incorrect parameter


nsarrazin (approved):
Woops sorry for the regression, thanks for contributing a fix :rocket:

Issue #1689, opened 2025-02-07
Enhance toast visibility and readability


ishatalkin:
## Describe your feature request

This feature request aims to improve user experience by enhancing the toast display. Specifically, the timeout duration for the toast is requested to be increased to ensure it stays visible long enough for users to read, even if the message is lengthy. Additionally, a maximum width is requested to be set for the toast to prevent showing it on the full screen, making it easier to read long messages.

nsarrazin:
Closed with #1690

PR #1690, opened 2025-02-07
feat: Enhance toast display to improve user experience (#1689)

ishatalkin:
1. Updated the Toast component to add a maximum width for the message to prevent it from stretching across the full screen.
2. Increased the timeout duration for error toasts from 3 seconds to 10 seconds to ensure long messages are readable.


nsarrazin (approved):
Appreciate it, thanks :rocket:

PR #1703, opened 2025-02-08
feat(api): migrate login and logout to API routes

nsarrazin:
- Replaced form-based login/logout with fetch-based API routes
- Updated hooks and components to use new `/login` and `/logout` endpoints. 

a POST on /login will return the authorization url, the callback function is the same, except the load function is now a GET server endpoint.


nsarrazin:
cc @cyrilzakka I guess this could have some impact :smile:

PR #1717, opened 2025-02-13
fix: move FinalAnswer outside of reasoning processing blocks

brians1196:
moving final answer outside of reasoning block so the ui displays all tokens for non reasoning models


nsarrazin (approved):
Thanks for catching this, seems to work great!

PR #1730, opened 2025-02-25
Update openai.md

JeanBaptiste-dlb:
I was able to set multimodal: 'true' when using a open ai backend running Qwen2-VL in a vllm openai like backend


nsarrazin (commented):
Thanks for the contrib!

nsarrazin (approved):

PR #1747, opened 2025-03-07
Adding Deepseek R1 to HuggingChat UI config

### LINKED ISSUE #1724: HuggingChat should use the original Deepseek R1 instead of the Distilled Qwen 32 B parameter version.
## Bug description

The point is, when I use the distilled 32B Qwen version of deepseek, it significantly hallucinated like gpt-4o in many mid-tier tasks. Instead, when I switched back to the original R1 model on HuggingFace, the model ddid not hallucinate.

## Steps to reproduce

Maybe the devs should replace deepseek r1-distill-32b-qwen model with the original deepseek-r1 model. The latter is proved to be more capable at handling more complex tasks and does not hallucinate like the r1-distill-32b-qwen model.
Please fix this problem as soon as possible, the current deepseek r1-distilled-32b model is now unusable after this type of error.

## Screenshots

Image

The model hallucinates every time i prompt it to write a vietnamese essay.

Image

And this too.... :(

### Specs

- **OS**: Ubuntu 24.04 LTS
- **Browser**: Vivaldi

## Notes

Since the current deepseek-r1-distill-32b model is based on the qwq-32b-preview model in which the latter is already available on huggingchat, so why having deepseek-r1-distill-32b model anyway? It's best to replace it with the original deepseek-r1.

gmanskibiditoilet:
Fixes #1724 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -924,7 +924,7 @@ Example configuration for DeepSeek R1 (token-based):
 
 ```json
 {
-	"name": "deepseek-ai/DeepSeek-R1-Distill-Qwen-32B",
+	"name": "deepseek-ai/DeepSeek-R1",
 	// ...
 	"reasoning": {
 		"type": "tokens",
--- a/chart/env/prod.yaml
+++ b/chart/env/prod.yaml
@@ -132,10 +132,10 @@ envVars:
       },
       {
         "name": "deepseek-ai/DeepSeek-R1-Distill-Qwen-32B",
-        "modelUrl": "https://huggingface.co/deepseek-ai/DeepSeek-R1-Distill-Qwen-32B",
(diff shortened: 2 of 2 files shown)


nsarrazin:
Hey so like I said before [here](https://github.com/huggingface/chat-ui/issues/1724#issuecomment-2688917093) this is not something that can be done simply by updating the HuggingChat config. The model is not hosted in our API right now so this would just give an error in chat. We keep the config open to be transparent with our users but that doesn't mean anyone can just change it.

We are currently working on provisioning resources for hosting R1 but given the size of the user base on HuggingChat and the size of the model this requires significant planning so the best thing you can do is just to give visibility to your request in [the models discussion page](https://huggingface.co/spaces/huggingchat/chat-ui/discussions/372).

PR #1764, opened 2025-03-21
Fix system message handling to preserve user-configured system prompts

IAmStoxe:
## Problem
Previously, when a preprompt was provided, it would completely replace any user-configured system message instead of being prepended to it.

## Solution
- Added explicit detection of existing system messages
- When a system message exists and preprompt is defined, prepend preprompt to the existing content
- When a system message exists and no preprompt is defined, preserve the user's message unchanged
- When no system message exists, create one with either preprompt or empty string
- Maintained handling for models that don't support system roles

## Testing
| Scenario | User System Prompt | Preprompt | Result |
|----------|-------------------|-----------|--------|
| 1 | ❌ No | ❌ No | System message with empty content is created |
| 2 | ❌ No | ✅ Yes | System message with preprompt content is created |
| 3 | ✅ Yes | ❌ No | User's system prompt is preserved unchanged |
| 4 | ✅ Yes | ✅ Yes | Preprompt + "\n\n" + User's system prompt |

## Breaking Changes
None - this change preserves all existing functionality while fixing the prompt handling logic.

### DIFF
--- a/src/lib/server/endpoints/openai/endpointOai.ts
+++ b/src/lib/server/endpoints/openai/endpointOai.ts
@@ -198,25 +198,37 @@ export async function endpointOai(
 			toolResults,
 			conversationId,
 		}) => {
+			// Format messages for the chat API, handling multimodal content if supported
 			let messagesOpenAI: OpenAI.Chat.Completions.ChatCompletionMessageParam[] =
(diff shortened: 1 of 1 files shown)


nsarrazin (approved):
Thank you!

Issue #1774, opened 2025-03-26
Improve quickstart setup experience


nsarrazin:
We would like to improve the experience of running chat-ui locally and reduce friction in the setup.

Current pain points:
- Having to setup your DB
- Having to point to an inference server
- Having to deal with either docker or git + npm

The goal is to reduce the friction with the following steps:
- [x] Make the DB optional, use an in-process mongo server if it's not set (#1773)
- [x] Make it possible to download and run models directly from chat-ui (with [node-llama-cpp](https://node-llama-cpp.withcat.ai/)?) #1778 
- [x] Allow admin login from CLI à la jupyter notebook (#1789 )
- [ ] Add an onboarding UI for setting up new models from the app
- [ ] Provide an easy setup like `npx @huggingface/chat-ui` or `pip`
- [ ] Update the README so it focuses on quickstarting and move the rest of the doc to the docs page

arch-btw:
Completely agree with this, I was on the verge of opening an issue asking if this would be possible, looks like you already thought of it.

For us that don't use Docker, the setup gets pretty difficult. I'm also concerned about maintaining my instance and losing data if I were to migrate all of it to a different device due to the complexity.

I like the idea of making the DB optional, maybe something like optional SQLite (which is embeddable, no separate server needed) might even simplify things further? I haven't checked to see if something like this is possible already, though.

The final thing I wanted to suggest is that maybe GitHub Actions could auto-build releases. 

Excited about these changes! Looks like the first one is done already :)

nsarrazin:
So the DB is now embedded! @arch-btw  If you run the app without setting MONGODB_URL there should be a `db` folder that gets created where the DB data is added. If you backup that folder you should be able to recreate your data on any device.

> The final thing I wanted to suggest is that maybe GitHub Actions could auto-build releases.

Curious what build output you would be interested in if you're not using docker, do you mean the output of `npm run build`?

nsarrazin:
No worries! I think we might be able to distribute the node app as a binary in the future actually, that's a good idea 😄

mehdiamiriorg90-droid:
JWT IP

PR #1778, opened 2025-03-31
feat: add a local endpoint type for inference directly from chat-ui

nsarrazin:
Part of #1774 

- [x] Run models locally from .gguf file
- [x] Auto-download model if not stored locally 
- [x] Use GPU if available
- [x] Get chat template from .gguf file
- [x] Show every .gguf in `models/` as a model if `MODELS` is undefined
- [ ] Handle batching & multiple model inference at once more gracefully


nsarrazin:
Works well you can do something like

```
MODELS=`[{
  "name": "HuggingFaceTB/SmolLM2-1.7B-Instruct",
  "parameters": {
    "stop_sequences": ["<|im_end|>", "<|endoftext|>"]
  },
  "endpoints": [{"type": "local", "modelPath": "hf:HuggingFaceTB/SmolLM2-1.7B-Instruct-GGUF:Q4_K_M"}]
}]`
```
it will automatically use your GPU if available and download models to the `models` folder if not available locally. 

It's still super rough as it doesn't handle running out of memory gracefully so I'm still working on dealing with this better. 

I also want to automatically expose any .gguf files in the `models/` folder as a model in chat-ui without having to set the `MODELS` env var

nsarrazin:
Merging for now, it works well in local testing! Will update the docs to explain this when I'm done with the quick setup.

PR #1787, opened 2025-04-07
Add `toolId` to ToolCall and use it in tool results

keligggg:
Add Tool id to Open ai tools calls.
Allows tools to work, especially for the Mistral AI api, via the open ai library.


nsarrazin (approved):
Thanks for the contribution!

PR #1790, opened 2025-04-10
Fix auto-scroll issue with message streaming in Svelte 5

### LINKED ISSUE #1766: Auto Scrolling not working during streaming responses
## Bug description

When the LLM response is streamed, the scrollbar does not automatically move down as the response content is generated and appended to the bottom of the screen. This results in the user only seeing the initial portion of the response and needing to manually scroll down to view the complete answer. The expected behavior is that the scroll position should automatically follow the generated content, ensuring the latest output is always visible.

## Steps to reproduce

1. Type "Code a snake game" in Huggingface Chat
2. Long response

### Specs

- **Browser**: I tested Chrome, Edge, Mobile Browsers
- **chat-ui commit**: Huggingface Chat

## Notes

Is `snapScrollToBottom.ts` working?

calycekr:
In Svelte 5, actions only update when the bound value changes by reference. Previously, snapScrollToBottom was bound to `messages`, which didn't trigger updates when only the content of a message changed during streaming.

Updated binding to `messages.map(message => message.content)` to ensure the action reacts to content changes, restoring expected scroll behavior.

Fix #1766

### DIFF
--- a/src/lib/components/chat/ChatWindow.svelte
+++ b/src/lib/components/chat/ChatWindow.svelte
@@ -248,7 +248,7 @@
 <div class="relative min-h-0 min-w-0">
 	<div
 		class="scrollbar-custom h-full overflow-y-auto"
-		use:snapScrollToBottom={messages.map(message => message.content)}
+		use:snapScrollToBottom={messages.map((message) => message.content)}
 		bind:this={chatContainer}
 	>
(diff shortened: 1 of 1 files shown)


nsarrazin (approved):
Works great! Thanks :fire:

PR #1802, opened 2025-04-22
feat: allow storing env variable in DB

nsarrazin:
This is a prerequisite for building a good onboarding UI.

We add a new collection in the DB that acts as a key-value store representing env variables.

You can do things like

```
npm run config add PUBLIC_APP_NAME 'my new app name'
```
or
```
npm run config add ENABLE_ASSISTANTS true
```

still needs a check to ensure the config is not stale and updates if needed on requests.


nsarrazin:
Only thing broken is the db access during test. working on a fix.

PR #1813, opened 2025-05-06
New `InferenceClient` endpoint type

nsarrazin:
This will become the default endpoint type. It already lets us enable tool calling on a lot of models where we didn't have it before. 

Making sure everything works for all the prod models and switching to it in this PR. 

- meta-llama/Llama-3.3-70B-Instruct
    - Tool: ✅
    - Normal: ✅
- Qwen/Qwen3-235B-A22B
    - Tool: ✅
    - Reasoning ✅
    - Normal: ✅
- Qwen/Qwen2.5-72B-Instruct
    - Tool ✅
    - Normal ✅
- CohereForAI/c4ai-command-r-plus-08-2024
    - Tool ✅
    - Normal ✅
- deepseek-ai/DeepSeek-R1-Distill-Qwen-32B
    - Tool ✅
    - Reasoning ✅
    - Normal ✅
- nvidia/Llama-3.1-Nemotron-70B-Instruct-HF
    - Normal ✅
- Qwen/QwQ-32B
    - Tool ✅
    - Reasoning ✅
    - Normal ✅
- google/gemma-3-27b-it (something is wrong with the multimodal image processing)
    - Tool ❌
    - Image ❌
    - Normal ✅
- mistralai/Mistral-Small-3.1-24B-Instruct-2503 (seems to be an endpoint config issue) 
    - Tool ❌
    - Normal ❌
- Qwen/Qwen2.5-VL-32B-Instruct (seems to be an inference proxy issue)
    - Tool ❌
    - Image ❌
    - Normal ❌
- microsoft/Phi-4
    - Tool ✅
    - Normal ✅
- NousResearch/Hermes-3-Llama-3.1-8B
    - Tool ✅
    - Normal ✅


src/lib/server/endpoints/hfinference/endpointHfInference.ts
```diff
@@ -0,0 +1,294 @@
...
+		};
+
+		toolChoices.push(toolDefinition);
+	}
+
+	return toolChoices;
+}
+
+export const endpointHfInferenceParametersSchema = z.object({
+	type: z.literal("hfinference"),
```
julien-c, line 83: i would name it either `inference-providers` (the name of the product) or `InferenceClient` (the name of the library class)

julien-c: otherwise it's confusing with `provider="hf-inference"`

src/lib/server/models.ts
```diff
@@ -380,6 +306,8 @@ const addEndpoint = (m: Awaited<ReturnType<typeof processModel>>) => ({
 						return endpoints.tgi(args);
 					case "local":
 						return endpoints.local(args);
+					case "hfinference":
```
julien-c, line 309:
```suggestion
					case "inference-providers":
```

gmanskibiditoilet:
alas, it is a great moment to enable community tools here. we have been waiting for it at last.

dagoatzmcclusky:
image

please note that since this PR has been launched, it somehow caused the entire chat-ui client to crash. please fix this.

dagoatzmcclusky:
image

please note that since this PR has been launched just recently, it somehow caused the entire chat-ui client to crash right immediately on startup.

nsarrazin:
It was a delay between the moment where prod env variables are updated and the container is updated. old containers were trying to use the new config resulting in a 500.

gmanskibiditoilet:
> It was a delay between the moment where prod env variables are updated and the container is updated. old containers were trying to use the new config resulting in a 500.

alright, thanks for the notice!

gmanskibiditoilet:
image

but there's another problem. when we tried to add a prompt to one of the chatbots, the bot tried reasoning for a while but then it somehow crashed again. (the chatbot we used was qwq-32b)

gmanskibiditoilet:
please don't forget to update the prod env variables and the containers that run them at the same time to prevent any further misshaps or unexpected errors.

Issue #1550, opened 2024-11-01
Add full-text search in chat history


kadykov:
## Describe your feature request

Allow users to search for specific keywords or phrases within the chat history, making it easier to find and recall previous conversations.

## Screenshots (if relevant)

An example of the search bar placement could be found in #1079

## Implementation idea

One possible implementation could be to use a library to index the chat history data. This would allow for efficient and scalable search functionality. The search bar could be added to the chat history interface, and when a user enters a search query, it would send a request to the search index to retrieve relevant results. The results could be displayed in a dropdown list or a separate search results page, with links to the original chat messages.

## Previous proposals and why this one is different

I'm aware that a similar proposal was made in the past #243, but it was rejected in favor of using the browser's page search functionality (ctrl + F). However, I'd like to argue that page search does not provide the same functionality as a dedicated full-text search in chat history. Here's why:

- Page search is limited to the currently loaded chat history and previous chat names, whereas a dedicated search would allow users to search across the entire conversation history, even if it's not currently loaded on the page.
- Page search does not provide any contextual information, such as the date and time of the message, or the conversation, whereas a dedicated search could provide this information and make it easier for users to understand the context of the search results.

Given these differences, I believe that a dedicated full-text search in chat history is a valuable feature that would greatly improve the user experience, and I'd like to propose it again for consideration.

Personally, I tend to create a new chat for each small problem to keep the LLM focused on what's important. As a result, I end up with too many chats with similar names, which makes the browser page search nearly useless.

siddharthgoel88:
I second this 👍 This would be a great addition. ChatGPT also allows to search `⌘ K` combination.

kramosss:
Hi, 

I'm thinking of working on this feature if it's still available and I just wanted to confirm and check my understanding.

So far I've looked through the codebase about where this feature would be implemented and also did some research about how to get started on this feature. I think in the codebase, where the main changes would take place for this feature are in src/lib/components/chat/ChatInput.svelte.

I'm thinking to tackle this, this feature could be broken down into two parts.

-First in src/lib/components/chat/ChatInput.svelte an event similar to chatGPT's command + k would be implemented and a UI to show user input and relevant results. 

-From there, I'm thinking I could use the fuzzy search algorithm based on user input to match with results from chat log stored in the database for that user and also have to make the relevant api calls. 

As mentioned before, this feature would work every similar to chatGPT's command + k feature which I didn't know about, but is highly useful.

kramosss:
[...]

PR #1823, opened 2025-05-14
Add first version of the pop up search (Ctrl/Command  + K) to search entire conversation history 

hongantran3804:
**Description:**

- This PR introduces a first version of the popup search feature, developed by me and @kramosss, as discussed in the [issue](https://github.com/huggingface/chat-ui/issues/1550). The feature allows users to press Ctrl / Cmd + K to quickly search through their previous conversations. We’re submitting this initial version to gather feedback from other contributors and improve it further.

**How It Works:**
- When the user triggers the search popup, we make an API call to fetch all conversations.

- On the frontend, we filter the conversations to show only those that include the search query.

- Matching conversations are then displayed in the pop-up interface.

**Question / Feedback Requested:**
- We’re currently filtering on the frontend for simplicity and responsiveness. However, we're aware of trade-offs:

- Frontend filtering avoids server load but may lead to slow performance or memory issues for users with many conversations.

- Backend filtering would reduce payload size and improve performance, but it could increase server load since every user input change may trigger a new request.

**We’re unsure which strategy would scale better and offer the best user experience. Would it be better to:**

- Move filtering to the backend with debounced input to limit request frequency?

- Stick with frontend filtering and maybe add pagination or limit results?

We’d appreciate any guidance or suggestions on the best approach to handle this trade-off.


gmanskibiditoilet:
absolute great work!

gmanskibiditoilet:
any progress out there? we would love to hear out the news!

nsarrazin (approved):
Hi, thanks for the contribution! This is great, I took some time to make it fit better with the rest of the app, a little overview of what I did:

- I removed the custom icons, we try to use carbon wherever possible and only use custom icons when we can't find what we need. 
- I made use of the `NavConversationItem.svelte` component we had for displaying conversation links
- I **moved the search to the backend** using a text index on user conversations, this will reduce the load on client and network I/O. 
- I moved the `Search` component to the top layout and added a button to open the search in `NavMenu` ( I also kept the shortcut, nice idea!)
- Removed the passing of `messages` when returning conversations, to reduce I/O since the filtering is on the backend now.

I took the liberty to do this directly on the PR, let me know if you have any questions.

And again thanks for tackling this! :fire:

Issue #1767, opened 2025-03-25
Incorrect mobile device detection - touch screen laptops trigger mobile UI behavior


calycekr:
## Bug description

The current mobile device detection logic incorrectly identifies laptops with touch screen functionality as mobile devices, resulting in the mobile user interface being displayed. This impacts usability on laptops with touch screen.

Submitting messages via enter key doesn't work, etc.

## Steps to reproduce

1. Access the application on a laptop with a touch screen.
2. Observe the user interface behavior.
3. Submitting messages via enter key doesn't work.

## Notes

I think root causes are in `function isVirtualKeyboard()` in `ChatInput.svelte`
https://github.com/huggingface/chat-ui/blob/d238bc96b6bd2c517285013ec8af5d651694f5d0/src/lib/components/chat/ChatInput.svelte#L88-L101

#1635

nsarrazin:
Hi thanks for reporting, I've seen it reported before in the discussions page so it's good to track it here too. :)

Not sure what the best option is, the alternative is to use media queries based on screen width, for example on the ChatGPT interface that's what they do but the downside is that if the page width is small enough you also can't send with enter. 

Doing it based on touch points seemed like the better option but touch screens are an issue then. Maybe we could do both?

PR #1824, opened 2025-05-14
Fix incorrect touchscreen detection

zvun:
The UI erroneously detects a touchscreen laptop as a phone and assumes it has a virtual keyboard, which makes the Enter button not send a message but rather start a new line. This was mentioned in #1767.


nsarrazin (approved):
Thanks for the fix!

PR #1838, opened 2025-05-27
feat(reasoning): support end token search across multiple chunks

nsarrazin:
also includes an update to the reasoning code so we support searching for the end token across multiple chunks from the endpoint. (in case the end sentence is not a single token)


gmanskibiditoilet:
keep going man!

PR #1841, opened 2025-05-30
Search feature enhancements

vijay2904:
Features enhanced / added:

1. Modified query logic to return all the matching snippets (not just the first N) using infinite scroll, in sets of 5.
2. Added a short description to show the matching content with search keyword highlighted in **bold**.
3. Added padding to the search icon in the nav menu.
4. Added `ctrl/cmd k` to the end of the placeholder for users to know.

### DIFF
--- a/src/lib/components/InfiniteScroll.svelte
+++ b/src/lib/components/InfiniteScroll.svelte
@@ -44,7 +44,7 @@
 </script>
 
 <div bind:this={loader} class="flex animate-pulse flex-col gap-4">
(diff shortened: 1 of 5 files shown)


vijay2904:
Thanks for accepting my changes. MongDB is not my strong forte >_<.

I was able to learn from your modifications. Thank you so much.

If you think my contribution is worth it, please assign the issue to me before closing.

- [ ] By the way, I would like to write an article on my blog about this contribution with your permission.

PR #1863, opened 2025-07-04
fix: Ensure deterministic height adjustment in handleKeydown

calycekr:
This PR improves the fix introduced in #1854 by ensuring the textarea height adjustment completes deterministically after the submit event is dispatched.

Previously, there was a race condition where the textarea height adjustment could be scheduled but not applied before submission, leading to inconsistent UI behavior.

Changes:
- Improved the existing logic by `tick().then()` in `handleKeydown` ensuring DOM updates are applied after the submit event is dispatched.

Fixes: #1854


gary149:
ty reworked it

PR #1865, opened 2025-07-04
fix: reduce min pods following closure

rtrompier:


rtrompier:
For now, pods consume less than 2% of the CPU, and we have autoscaling based on the CPU if needed, so 12 instances is overkill IMO, but we can wait a bit before scaling down if you prefer.

nsarrazin:
No sounds good to me :smile:

PR #1884, opened 2025-10-02
fix CI

coyotte508:


coyotte508:
there is still an error about a dependency using `require` in ESM context (maybe `@gradio/client`) but most of the stuff is fixed

PR #1960, opened 2025-10-27
allow guest on home and shared convo

krampstudio:
https://github.com/user-attachments/assets/5d15f0b8-9844-4104-8852-7ccd5e9e47cf

### DIFF
--- a/src/lib/server/api/routes/groups/models.ts
+++ b/src/lib/server/api/routes/groups/models.ts
@@ -1,4 +1,5 @@
-import { Elysia } from "elysia";
+import { Elysia, status } from "elysia";
+import { refreshModels, lastModelRefreshSummary } from "$lib/server/models";
 import type { BackendModel } from "$lib/server/models";
 import { authPlugin } from "../../authPlugin";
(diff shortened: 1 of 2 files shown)


src/lib/components/NavConversationItem.svelte
```diff
@@ -60,6 +62,7 @@
 				title="Confirm delete action"
 				onclick={(e) => {
```
coyotte508: Maybe better to make the button disabled?

krampstudio:
> For api endpoints could send a 401, too (btw we can enable remote functions maybe, instead of api endpoints, [tried it with mongoku](https://github.com/huggingface/Mongoku/blob/master/src/api/servers.remote.ts) was nice, and easier to create a `checkAuth()` helper reused in most endpoints)

the API should already returned 401 for `POST`, `PUT` & `DELETE`  
https://github.com/huggingface/chat-ui/blob/90e32543d4a68f46923430d11fa24a03448df171/src/hooks.server.ts#L222-L231

but yes, let's change to something more robust and cleaner (I'll follow up)

PR #1988, opened 2025-11-19
mcp: tool results now properly rendered

ServeurpersoCom:
## Text content: `<pre>` + monospace (terminal-style)

tool-result-text

## Image content: base64 decoded as ``

tool-result-image

## Conforms to MCP spec 2025-06-18


ServeurpersoCom:
I tested the limits by asking the model to produce an SVG and display it. The MCP "view" command (Claude-style) shouldn't convert it to base64 image since it's more useful as source code, and the model (a small GPT-OSS-20B) had a stroke of genius: it encoded the SVG to base64 in the terminal (long line) and reformatted it as a base64 image in the context, which renders correctly. So I added a commit to properly wrap long lines.

Capture d’écran 2025-11-19 084218

PR #1994, opened 2025-11-24
Forward original MIME 

gary149:
The API now returns the content-type from the fetched response instead of always using text/plain. This allows clients to receive the correct content type for the requested resource.

### DIFF
--- a/src/lib/utils/loadAttachmentsFromUrls.ts
+++ b/src/lib/utils/loadAttachmentsFromUrls.ts
@@ -92,8 +92,7 @@ export async function loadAttachmentsFromUrls(
 				}
 
 				const forwardedType =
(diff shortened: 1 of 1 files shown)


src/lib/utils/loadAttachmentsFromUrls.ts
```diff
@@ -82,13 +91,17 @@ export async function loadAttachmentsFromUrls(
 					return;
 				}
 
+				const forwardedType =
+					response.headers.get("x-forwarded-content-type") ||
+					response.headers.get("x-original-content-type");
```
coyotte508:
```suggestion
				const forwardedType =
					response.headers.get("x-forwarded-content-type");
```

coyotte508 (approved):
one day those lint errors should be fixed :)

PR #1998, opened 2025-11-26
[FIX] Call MCP tools with audio as inputs

frascuchon:
When calling a configured Gradio MCP server tool that has audio references (URLs), the tool call constantly fails due to an input format issue. This PR handles audio inputs to allow calling remote Gradio MCP tools with referred audio URLs (direct audio file uploading is not implemented)

Additionally, this PR allows configuring the default MCP tool timeout by environment variable.


frascuchon:
Hi. Yes, I've tried locally. This PR attempts to address only the issue that occurs when calling an MPC tool with an audio file URL. Currently, the tool call is failing due to an unexpected input format. I did not work with multi-modal models.

PR #2119, opened 2026-02-13
fix: prevent SSRF via open redirect bypass in fetch-url endpoint

rtrompier:
## Summary

- The `/api/fetch-url` endpoint followed HTTP redirects automatically, allowing attackers to bypass URL validation by redirecting through an external HTTPS server (e.g. a Cloudflare Worker) to internal services (`http://127.0.0.1:5565/`). This exposed Prometheus metrics, process info, and enabled internal port scanning.
- Redirects are now followed manually with `redirect: "manual"`, validating each hop via `isValidUrl()` + DNS resolution (`assertResolvedUrlSafe()`) before following it.
- Added `ipaddr.js` dependency for robust IP range detection (handles IPv4-mapped IPv6, link-local, etc.) and DNS-based SSRF protection against DNS rebinding attacks.
- Redirect chain capped to 5 hops.

## Attack vectors blocked

| Vector | Protection |
|---|---|
| Direct internal URL (`http://127.0.0.1`) | `isValidUrl()` rejects non-HTTPS |
| Open redirect to internal URL | `redirect: "manual"` + per-hop validation |
| DNS rebinding (domain resolving to 127.0.0.1) | `assertResolvedUrlSafe()` DNS lookup |
| IPv4-mapped IPv6 (`::ffff:127.0.0.1`) | `ipaddr.js` range detection |
| Infinite redirect loop | Max 5 redirects |


package.json
```diff
@@ -81,6 +81,7 @@
 		"highlight.js": "^11.7.0",
 		"htmlparser2": "^10.0.0",
 		"ip-address": "^9.0.5",
+		"ipaddr.js": "^2.3.0",
```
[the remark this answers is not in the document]
rtrompier: Removed `ipaddr.js`, now using `ip-address` (already a dependency) with `Address4.isInSubnet()` for proper CIDR subnet matching. No more lockfile changes.

package-lock.json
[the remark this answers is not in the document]
rtrompier: Fixed — no dependency changes at all anymore, lockfile is untouched.

package.json
coyotte508: you can revert this then :) (and package-lock.json)

since not using `ipaddr.js` anymore

rtrompier: Done — removed `ipaddr.js` from `package.json` and reverted `package-lock.json` to main.

rtrompier: yes haha

PR #2187, opened 2026-03-16
fix: respect USE_USER_TOKEN config in OpenAI endpoint

balex42:
## Description
This PR fixes an authentication bug where the `USE_USER_TOKEN` environment variable is ignored across both standard and MCP OpenAI-compatible endpoint calls.

Currently, if a user session exists (e.g., via a custom OIDC provider), `locals.token` is unconditionally injected as the `Authorization: Bearer` header for LLM inference calls. This overwrites the configured `OPENAI_API_KEY`, causing a `401 Unauthorized` error when interacting with external OpenAI-compatible backends (like Ollama Cloud or vLLM) that do not recognize the OIDC token.

## Motivation and Context
The `.env` file explicitly documents that the user token should only be used for inference if `USE_USER_TOKEN=true`:
`# When set to true, user token will be used for inference calls`
`USE_USER_TOKEN=false`

However, the implementation was unconditionally applying `locals.token` if it existed. This PR updates the headers payload to respect the `USE_USER_TOKEN` configuration for both standard generation and tool use.

## Changes
Modified the following files to check `config.USE_USER_TOKEN === "true"` before injecting `locals.token` into the authorization header:
* `src/lib/server/endpoints/openai/endpointOai.ts` (Covers standard streaming and non-streaming completions)
* `src/lib/server/textGeneration/mcp/runMcpFlow.ts` (Covers MCP tool-enabled streaming and fallback non-streaming completions)

**Before:**
```typescript
...(locals?.token ? { Authorization: `Bearer ${locals.token}` } : {}),

```

**After:**

```typescript
...(config.USE_USER_TOKEN === "true" && locals?.token ? { Authorization: `Bearer ${locals.token}` } : {}),

```

## How Has This Been Tested?

1. Configured a custom OIDC provider (`OPENID_CLIENT_ID`, `OPENID_PROVIDER_URL`).
2. Configured an external OpenAI-compatible backend (`OPENAI_BASE_URL` and `OPENAI_API_KEY`).
3. Set `USE_USER_TOKEN=false`.
4. Logged in via OIDC.
5. Attempted to send a standard chat message, and a chat message requiring tool (MCP) use.

* **Without this PR:** The API calls fail with `401 Unauthorized` and `[mcp] flow failed, falling back to default endpoint` because the backend receives the OIDC token instead of the API key.
* **With this PR:** Both API calls succeed because the `OPENAI_API_KEY` is correctly utilized.


gary149:
ok (sorry for the delay)

PR #2237, opened 2026-04-15
fix: include orgs with a paid plan in billing dropdown

Charlie-Boyer:
## Problem

The billing org dropdown in HuggingChat settings filters orgs on `canPay === true` only. Orgs on a Team or Enterprise plan that haven't purchased additional prepaid credits have `canPay = false`, so they don't appear in the dropdown even though their plan includes monthly inference credits ($2/seat/month).

The Hub's inference widget uses `org.plan || canPay !== false`, so those same orgs show up fine there.

## Fix

Include orgs that have a paid plan (`org.plan` is truthy) in the billing orgs filter, matching the Hub's behavior. The `plan` field is already returned by `/oauth/userinfo`.


gary149 (approved):

gary149:
thanks (didn't test if you want to report if it's fixed in prod)

PR #2249, opened 2026-04-29
fix(auth): make sameSite cookie default depend on secure flag to fix HTTP deployments

### LINKED ISSUE #2243: Application can't send chats to local VLLM
## Bug description

chat-ui can't communicate with local VLLM. Possible this is only the case if plain HTTP is used, I don't have SSL certificates.

## Steps to reproduce

1. Run VLLM on your local PC. Let's say it's the default port, so it's running on http://localhost:8000/v1 for the API. My command below.
2. Use 'docker compose up' with the file given below
3. On your PC, connect to http://localhost:3000 to load. The webpage should load and the VLLM model should be detected properly.
4. Start a new chat and send something. Nothing happens.

VLLM command:
```
vllm serve $MODEL_PATH \
--served-model-name $MODEL_NAME \
--port 8000 \
--host 0.0.0.0 \
--max-model-len 262144 \
--max-num-seqs 4 \
--enable-prefix-caching \
--gpu-memory-utilization 0.6 \
--load-format fastsafetensors \
--enable-auto-tool-choice \
--tool-call-parser qwen3_coder \
--reasoning-parser qwen3 \
--max-num-batched-tokens 8192 \
--trust-remote-code \
--mm-encoder-tp-mode data \
-O3
```

chat-ui's docker-compose.yml
```
services:
  chat-ui:
    image: ghcr.io/huggingface/chat-ui-db:latest
    container_name: chat-ui
    init: true
    network_mode: host
    environment:
      - OPENAI_BASE_URL=http://localhost:8000/v1
      - OPENAI_API_KEY=not-needed
    volumes:
      - chat-ui-data:/data

volumes:
  chat-ui-data:
```

### Logs

I can see your app successfully fetching the model list and showing them in the UI settings page. 

However, when I try to send a question, nothing happens. This exception is thrown in my browser:
```
GET http://192.168.100.100:3000/api/v2/conversations/69e7f5df4ec7018555a490d2
Error: {"message":"You don't have access to this conversation. If someone gave you this link, ask them to use the 'share' feature instead."}`
```

No request is sent to VLLM other than /v1/models to get the model list according to VLLM logs.

There's also this error in the browser console log when I load the app which may be related:
```
Cookie “hf-chat” has been rejected because a non-HTTPS cookie can’t be set as “secure”
```

Could this be No Cookie -> All conversations denied? Is this because I don't use SSL? That's my business, isn't it?

### Specs

- **OS**: Ubuntu 24.04
- **Browser**: Firefox
- **chat-ui commit**: chat-ui-db b1407709b522, latest as of today

### Config

Refer to docker-compose.yml

anishesg:
## Problem

When running chat-ui over plain HTTP (e.g. a local VLLM setup without TLS), the browser rejects the `hf-chat` session cookie because it is set with `Secure; SameSite=None`. Browsers require `Secure` for `SameSite=None` cookies, and drop `Secure` cookies entirely over HTTP. The result is that the session is never established: the model list endpoint succeeds (no auth required) but every conversation request returns a 403 "You don't have access to this conversation" error.

The root cause is in `src/lib/server/auth.ts`: `sameSite` was computed before `secure`, so its default of `"none"` was hardcoded for non-dev environments regardless of whether `COOKIE_SECURE=false` was set. Even when a user explicitly sets `COOKIE_SECURE=false`, the cookie was still sent as `SameSite=None`, which browsers also reject without `Secure`.

## Fix

Reorder the computation in `auth.ts` so that `secure` is resolved first, then use it to derive the `sameSite` default: when `secure=false` (HTTP mode, dev, or `ALLOW_INSECURE_COOKIES=true`), `sameSite` defaults to `"lax"` instead of `"none"`. Both variables are now exported so the `logout` route can reuse them directly instead of duplicating the logic. The docs for the common-issues page are updated to reflect that only `COOKIE_SECURE=false` is required for HTTP deployments.

Fixes #2243

### DIFF
--- a/docs/source/configuration/common-issues.md
+++ b/docs/source/configuration/common-issues.md
@@ -6,12 +6,13 @@ This usually happens when running Chat UI over HTTP without proper cookie config
 
 **Recommended:** Set up a reverse proxy (NGINX, Caddy) to handle HTTPS.
 
-**Alternative:** If you must run over HTTP, configure cookies:
+**Alternative:** If you must run over HTTP, set the following env var:
 
 ```ini
 COOKIE_SECURE=false
-COOKIE_SAMESITE=lax
 ```
+
+This automatically sets `COOKIE_SAMESITE` to `lax`, which is the correct value for HTTP deployments. You can still override it explicitly with `COOKIE_SAMESITE=lax` if needed.
 
 Also ensure `PUBLIC_ORIGIN` matches your actual URL:
 
--- a/src/lib/server/auth.ts
+++ b/src/lib/server/auth.ts
@@ -57,15 +57,15 @@ export const OIDConfig = z
 
 export const loginEnabled = !!OIDConfig.CLIENT_ID;
 
-const sameSite = z
-	.enum(["lax", "none", "strict"])
-	.default(dev || config.ALLOW_INSECURE_COOKIES === "true" ? "lax" : "none")
-	.parse(config.COOKIE_SAMESITE === "" ? undefined : config.COOKIE_SAMESITE);
-
-const secure = z
+export const secure = z
 	.boolean()
 	.default(!(dev || config.ALLOW_INSECURE_COOKIES === "true"))
 	.parse(config.COOKIE_SECURE === "" ? undefined : config.COOKIE_SECURE === "true");
+
+export const sameSite = z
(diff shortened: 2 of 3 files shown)


gary149:
lgtm

PR #2274, opened 2026-05-18
[codex] improve smooth streaming flow

gary149:
## Summary

- Normalize stream transport padding before raw or smooth stream handling.
- Improve smooth streaming so long unbroken output is chunked, partial buffered text is flushed on source errors, and control updates can catch up to queued stream text.
- Simplify the conversation page by removing duplicate stream-token padding cleanup.

## Why

The existing smooth stream flow stripped server padding only after the smoother had already seen the tokens. That could make artificial padding affect chunk detection and pacing. The smoother also waited behind cosmetic delays even when tool/status/final updates were queued.

## Impact

Smooth streaming should feel more consistent across provider chunk shapes, avoid stalls on long unbroken text, and surface control updates without unnecessary delay. Raw mode still keeps provider-native stream chunking, now with transport padding normalized at the iterator boundary.

## Validation

- `npm test -- --run src/lib/utils/messageUpdates.spec.ts`
- `npm run check` (0 errors; existing `ChatMessage.svelte:274` a11y warning remains)
- `npx prettier --check src/lib/utils/messageUpdates.ts src/lib/utils/messageUpdates.spec.ts 'src/routes/conversation/[id]/+page.svelte'`

### DIFF
--- a/src/lib/utils/messageUpdates.spec.ts
+++ b/src/lib/utils/messageUpdates.spec.ts
@@ -131,6 +131,76 @@ describe("smoothStreamUpdates", () => {
 		expect(streamText(updates)).toBe("word ".repeat(40));
 	});
 
+	it("resumes smoothing after a control update catches up", async () => {
+		const source: MessageUpdate[] = [
+			{ type: MessageUpdateType.Stream, token: "word ".repeat(40) },
+			{ type: MessageUpdateType.Title, title: "done" },
+			{ type: MessageUpdateType.Stream, token: "more ".repeat(20) },
+		];
+		const delays: number[] = [];
+		const seen: Array<{ update: MessageUpdate; time: number }> = [];
+		let nowMs = 0;
+
+		for await (const update of smoothStreamUpdates(fromArray(source), {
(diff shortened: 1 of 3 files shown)


gary149:
Addressed the smoothing-rate review feedback in `d41850c6`:

- Catch-up stream emissions no longer contribute to the pacing window.
- The pacing window resets after a control update catches up, so later stream chunks resume smoothing.
- The immediate first stream chunk is excluded from pacing-rate accounting.
- Long-buffer splitting now uses a single-pass code-point splitter instead of repeated full-buffer counts.
- Added regression coverage for `Stream(long) -> Title -> Stream(more)` and first-chunk rate inflation.

Validation rerun:

- `npm test -- --run src/lib/utils/messageUpdates.spec.ts` (20 tests)
- `npm run check` (0 errors; existing `ChatMessage.svelte:274` a11y warning remains)
- `npx prettier --check src/lib/utils/messageUpdates.ts src/lib/utils/messageUpdates.spec.ts 'src/routes/conversation/[id]/+page.svelte'`

PR #2292, opened 2026-06-05
doc(readme): move closing parens to correct pos

LoneRifle:
Quick one-liner PR to the README that corrects where the closing parens is in the following phrase:

> any service that speaks the OpenAI protocol (llama.cpp server, Ollama, OpenRouter, etc.**)** will work by default ~)~.


gary149 (approved):
idk if it's automated but thanks for the contrib

PR #2296, opened 2026-06-08
fix(chat/effort): drive caret icon by dropdown state

LoneRifle:
## Problem
The current chat window is reliant on CarbonCaretDown, which may not be the most suitable depending on context.

In particular:
- the caret next to the model display name is slightly misleading, given that it opens the settings modal
- the caret used in ThinkingEffortChip could change depending on whether the dropdown menu is open

## Solution
Replace CarbonCaretDown with more suitable icons
- ~use a gear icon next to the model display name so it is clearer that tapping would bring up the settings modal~ submitted separately as #2300, using `lucide/settings-2`
- listen to thinking effort dropdown menu state, and use that to determine whether caret down (closed) or caret up (open) should be displayed

#### Before
before

#### After
after

### DIFF
--- a/src/lib/components/chat/ThinkingEffortChip.svelte
+++ b/src/lib/components/chat/ThinkingEffortChip.svelte
@@ -1,6 +1,7 @@
 <script lang="ts">
 	import { DropdownMenu } from "bits-ui";
 	import CarbonCaretDown from "~icons/carbon/caret-down";
+	import CarbonCaretUp from "~icons/carbon/caret-up";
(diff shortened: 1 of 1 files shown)


gary149:
idk this feels a bit weird to me tbh 
image

LoneRifle:
> idk this feels a bit weird to me tbh

I've reworked the PR so that it adds just the effort dropdown for now. Choice of icon for model id indicator resubmitted as #2300

PR #2377, opened 2026-06-22
feat(offline): impl client chat store, service worker, UX 

### LINKED ISSUE #2344: Feature request: add offline mode - read conversations without connectivity

Add a read-only **offline mode** so users can open the app and read their existing
conversations (and attachments) when connectivity is flaky or absent. The server
stays the source of truth — this is purely an additive client-side cache plus a
service worker for the app shell.

### Motivation
chat-ui already ships a web app manifest and is installable, but it has no offline
capability: every load fetches conversations from the server, so with no network the
app shell fails to load and no conversation content is available. Users who want to
retrieve an earlier answer, a snippet, or an uploaded file "in a pinch" (commute, plane,
dead zone) currently can't. For an installed PWA this is a notable gap.

##  Implementation idea
Two additions, both leaving the server as the authoritative store:

1. **Client-side conversation cache (IndexedDB).** Route conversation reads (the list
   and single-conversation `load` functions) through a small repository abstraction that
   mirrors fetched, server-confirmed conversations into IndexedDB (cache-aside /
   stale-while-revalidate). Population is **lazy** — only what the user actually views,
   plus the sidebar list — so there's no eager full sync or quota blow-up.
2. **Service worker for the app shell.** Use SvelteKit's built-in `$service-worker`
   (`build`/`files`/`version`) to precache the shell so the app loads with no network,
   reusing the existing manifest. Network-first for data/API; cache-first only for the
   shell.

When offline (or a fetch fails), reads fall back to the IndexedDB mirror; when online,
the server response refreshes the mirror.

### Consistency model
Read-only offline keeps this simple and avoids merge conflicts entirely:

- **Mirror holds server-confirmed state only.** In-flight/optimistic turns stay in the
  transient UI layer and are never written to the mirror until the server acks them
  (i.e. a response has begun) — so there is nothing for a refresh to "clobber."
- **Server always wins.** On reconnect the next load overwrites the mirror with server
  truth. No client-side merge logic.
- **Unsent turn (network drop before the server received it) is dropped** — it lived only
  in transient UI state, so it's lost on reload, identical to today's behavior. No
  failed-send outbox.
- **Ambiguous case reconciled by message id.** Messages already carry client-generated
  UUIDs; if the refreshed server copy already contains an in-flight turn's id, it was
  received (drop the local pending — no duplicate); otherwise it was genuinely unsent.

### Scope / non-goals
- ✅ Read-only retrieval of previously-viewed conversations and their attachments offline.
- ✅ Faster perceived loads online (render cached, revalidate in background).
- ❌ No offline composing, queued sends, or offline edits (rename/delete).
- ❌ No change to the persistence model — MongoDB/GridFS stay the source of truth.
- Note: multi-tab (two tabs sharing one IndexedDB) is handled with a `BroadcastChannel`
  notify; last-writer-wins.

### Attachments
Cached conversations reference GridFS-backed files. To render them offline, the service
worker can runtime-cache the file/output responses for viewed conversations (or the
repository can store the blobs). Exact approach TBD in the offline-mode PR.

### Proposed delivery — two PRs
To keep reviews focused:

1. **PR 1 — client conversation cache/repository (foundation).** Introduces the
   IndexedDB repository and routes conversation reads through it as a cache-aside layer.
   No behavior change beyond stale-while-revalidate. *Honest caveat:* this is largely
   invisible groundwork — its only standalone, user-visible gain is snappier loads — with
   the real payoff arriving in PR 2. Splitting it out keeps each diff small and reviewable.
2. **PR 2 — service worker + offline mode.** Adds the service worker (app-shell
   precache), the offline read fallback, attachment caching, and the offline-state UX.

### Dependencies / tech
- IndexedDB access via [`idb`](https://github.com/jakearchibald/idb) (~1KB, promise
  wrapper) — open to raw IndexedDB if a new dep isn't desired.
- Service worker via SvelteKit's native `$service-worker` — no Workbox.

### Open questions for maintainers
- Is offline read mode something you'd accept in principle / want to own?
- Acceptable to add `idb`, or prefer raw IndexedDB?
- Preferred service-worker caching + versioning strategy (skipWaiting / clients.claim)?
- Preference for attachment offline support: SW runtime-cache vs. blobs in IndexedDB?
- Any guidance on storage quota / eviction handling you'd want surfaced to users?

I'm happy to implement both PRs if there's appetite.

_[The text for this issue was generated by Claude Code.]_

LoneRifle:
image

This pull request introduces robust offline support to the chat UI by implementing a new IndexedDB-backed caching layer for conversations and a reliable online/offline detection mechanism. The UI is now aware of connectivity status, disabling actions that require a network connection and providing clear feedback to users. The most significant changes are grouped below.

**Offline detection and UI adaptation:**

- Introduced a new `isOnline` store (`useIsOnline`) that reliably detects real connectivity by probing a healthcheck endpoint, improving on the limitations of `navigator.onLine`. UI components now reactively use this store to enable or disable actions depending on online status, and provide tooltips explaining why actions are unavailable when offline.

- Updated UI elements—including chat input, file uploads, attachments, new chat button, conversation rename/delete, and voice recording—to be disabled when offline, with appropriate visual cues and accessibility titles.

Closes #2344

### DIFF
--- a/src/lib/components/NavConversationItem.svelte
+++ b/src/lib/components/NavConversationItem.svelte
@@ -12,6 +12,9 @@
 	import EditConversationModal from "$lib/components/EditConversationModal.svelte";
 	import DeleteConversationModal from "$lib/components/DeleteConversationModal.svelte";
 	import { requireAuthUser } from "$lib/utils/auth";
+	import { useIsOnline } from "$lib/stores/isOnline.svelte";
+
+	const isOnline = useIsOnline();
 
 	interface Props {
 		conv: ConvSidebar;
@@ -57,7 +60,7 @@
 
 <div
 	class="group flex h-8 flex-none items-center gap-1.5 rounded-lg pr-1.5 pl-2 text-base text-gray-600 hover:bg-gray-100 max-sm:h-10 sm:text-sm dark:text-gray-300 dark:hover:bg-gray-700
-		{conv.id === page.params.id ? 'bg-gray-100 dark:bg-gray-700' : ''}"
+        {conv.id === page.params.id ? 'bg-gray-100 dark:bg-gray-700' : ''}"
 >
 	{#if inlineEditing}
 		<input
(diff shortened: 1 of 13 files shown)


gary149:
that's a lot of changes I'll try to take a look when I've time but can't guarantee it will be soon

PR #2380, opened 2026-06-23
fix(artifacts): add a worked edit example to reduce empty update blocks

gary149:
Follow-up to #2379. An N=70 artifact-edit eval (GLM-5.2) found the dominant *actionable* failure isn't the matching algorithm — it's that ~11% of edit turns emit a bare `<artifact type="update"></artifact>` with **no pairs** inside (the "no-pairs" case). The model correctly decides an update is needed but doesn't produce any `old_str`/`new_str`.

The system prompt only showed the placeholder *schema*, never a filled-in example. This adds:

1. A concrete worked example (recoloring a button) with real `old_str`/`new_str` content.
2. An explicit rule: an update block must contain at least one complete pair; if exact `old_str` text can't be produced, re-emit the full artifact with the same identifier instead.

Prompt-only change. `npm run check` and `npm run lint` pass.

Note: the eval also confirmed two pi-mono ideas were correctly **not** ported in #2379 — "match-all-against-original" diverged in 0/70 runs (we re-search per pair, so no stale-offset bug), and identifier drift / typography normalization are model-dependent safety nets (GLM-5.2 copies verbatim; other routed models like Kimi-K2.6 drift).

### DIFF
--- a/src/lib/server/textGeneration/artifacts.ts
+++ b/src/lib/server/textGeneration/artifacts.ts
@@ -36,7 +36,15 @@ Editing an artifact you created earlier in the conversation:
 <new_str>replacement text</new_str>
 </artifact>
 
+For example, to recolor a button in an existing "signup-form" artifact, emit exactly:
+
+<artifact identifier="signup-form" type="update">
+<old_str>background: #16a34a;</old_str>
+<new_str>background: #2563eb;</new_str>
+</artifact>
+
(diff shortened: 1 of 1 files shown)


gary149:
Superseded by #2382, which now includes this worked-example + never-empty-update prompt guidance (folded in commit 61a46793) alongside the parser fix for swapped close tags. Consolidating all artifact-edit-reliability changes into one PR to avoid the overlapping prompt edit.

PR #2407, opened 2026-07-02
perf: compress dynamic SSR and API responses

gary149:
## What

Adds a thin production server (`server.mjs`) that wraps the adapter-node handler with `compression`, so dynamic responses (SSR HTML, `/api/v2/*` JSON) are served with brotli/gzip. Static assets were already pre-compressed at build time via sirv, but everything rendered at runtime went out raw.

Measured against hf.co/chat in production: the HTML document is 318-355 KB with no `content-encoding` on every page view, and every conversation-switch JSON payload (20-70 KB) is also uncompressed.

## Why streaming is excluded

The chat token stream (`POST /conversation/[id]`) responds with `Content-Type: application/jsonl`, and `/api/v2/conversations/updates` uses `text/event-stream`. A compression middleware buffers output, which would silently destroy time-to-first-token. Both content types are explicitly skipped, so token streaming behavior is unchanged.

## Deployment

- `entrypoint.sh` now starts `node /app/server.mjs` instead of `build/index.js` (PORT/HOST/SHUTDOWN_TIMEOUT env vars behave the same, body-size limiting stays inside the adapter handler)
- `Dockerfile` copies `server.mjs` into the image
- `npm run dev` / `vite preview` are unaffected

## Verification

- Behavioral test against the wrapper: HTML/JSON compressed (br), `application/jsonl` and `text/event-stream` pass through unencoded and unbuffered (first chunk arrives at ~100 ms of a ~500 ms stream, 5 distinct chunks received)
- Built the app and booted through `server.mjs`: `GET /chat/` went from 198,427 to 23,528 bytes transferred (-88%), `Content-Encoding: br`, `Vary: Accept-Encoding` set
- `npm run check`, `npm run lint`, `npm test` (392 tests) all pass

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,6 +31,7 @@ USER user
 
 COPY --chown=1000 .env /app/.env
 COPY --chown=1000 entrypoint.sh /app/entrypoint.sh
+COPY --chown=1000 server.mjs /app/server.mjs
 COPY --chown=1000 package.json /app/package.json
 COPY --chown=1000 package-lock.json /app/package-lock.json
 
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -16,4 +16,4 @@ fi;
 
 export PUBLIC_VERSION=$(node -p "require('./package.json').version")
 
-dotenv -e /app/.env -c -- node --dns-result-order=ipv4first /app/build/index.js -- --host 0.0.0.0 --port 3000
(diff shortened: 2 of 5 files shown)


gary149:
Superseded by #2409, which lands the same compression approach with a mime filter for both streaming content types (application/jsonl and text/event-stream), a Vary merge that survives SvelteKit's own header writes, brotli level pinned to 4, unix-socket support, and graceful shutdown. Verified end to end against a production build.

PR #2427, opened 2026-07-14
Fix scroll stability during mid-turn content collapse and add Safari manual anchoring

gary149:
## Summary

This PR fixes scroll instability when content collapses mid-stream (e.g., thinking blocks) and adds manual scroll anchoring for Safari, which lacks native `overflow-anchor` support. The changes ensure that readers following live content aren't yanked away by above-viewport layout shifts, and that post-stream collapses remain perfectly still.

### Scroll Anchoring System
- **One-way spacer during streaming**: While a response streams, the send-anchor spacer only shrinks (never re-inflates) as reply content grows. This prevents mid-stream collapses from re-anchoring the view and yanking readers away from the live edge.
- **Re-inflation after streaming**: Once streaming ends, the spacer can re-inflate to absorb post-turn collapses (manual toggles, auto-collapse) with zero motion via constant `scrollHeight`.
- **Container resize handling**: Viewport changes (keyboard close, window resize) re-derive anchor geometry mid-turn to keep the anchored message properly positioned.

### Manual Scroll Anchoring (Safari)
- Detects lack of native `overflow-anchor` support and enables manual anchoring as fallback
- Tracks the deepest element at the viewport top while detached (binary search + descent into messages)
- Stores ancestor chain to survive markdown re-renders and streaming updates
- Compensates for above-viewport content changes by restoring the tracked element's position
- Detects pending user scrolls (scrollTop mismatch) to avoid canceling user input

### Thinking Block Collapse Animation
- Unified template for streaming and settled phases so the thinking block component instance survives the answer-start transition
- Adds smooth `slide` transition (220ms) to collapse instead of instant layout jump
- Prevents "expand bounce" by keeping the capped streaming viewport shape during slide-out
- Uses `$effect.pre` to decide collapse before render, ensuring animation starts from correct height

### Controller Enhancement
- New `adjustBy(delta)` method for attribution-safe relative scroll adjustments (used by manual anchoring compensation)
- Detects clamp-jumped scrollTop to force instant re-pin (avoids visible bounce when spacer re-inflates)

## Implementation Details

- `ChatScroll` now tracks `streaming` state (mirrored from `ChatWindow.loading`)
- `updateSpacer()` accepts `allowGrow` flag; during streaming, computed height is clamped to current height (one-way)
- Manual anchoring stores `readAnchor` with element chain + captured scrollTop to detect stale captures
- `OpenReasoningResults` uses `$effect.pre` for loading transitions and conditional viewport rendering during slide-out
- New test suite covers mid-turn collapse stability, post-stream re-inflation, container resizes, and Safari manual anchoring edge cases

## Testing

Added comprehensive test coverage:
- In-turn content shrink (thinking-block collapse) with viewport stability verification
- Post-stream collapse with spacer re-inflation and zero motion
- Container resize mid-turn (mobile keyboard close scenario)
- Manual anchoring: above-viewport collapse, descendant tracking, node replacement, user scroll detection

https://claude.ai/code/session_013T9Ld83jc8eegpSf39wgyD

### DIFF
--- a/.claude/skills/verify/SKILL.md
+++ b/.claude/skills/verify/SKILL.md
@@ -1,0 +1,46 @@
+---
+name: verify
+description: Build, launch, and drive chat-ui end-to-end against a mock streaming backend (no real LLM key) to verify UI/streaming changes at the real surface.
+---
+
+# Verifying chat-ui end-to-end
+
+No LLM key is needed: point the app at a local OpenAI-compatible mock.
+
+1. **Mock backend** (node, no deps): serve `GET /v1/models` returning
+   `{"object":"list","data":[{"id":"test/thinker",...}]}` and
+   `POST /v1/chat/completions`; when `stream: true`, write SSE
+   `data: {"choices":[{"delta":{"content":"..."}}]}` chunks every ~60ms —
+   start with `<think>` + ~45 chunks + `</think>` to exercise the reasoning
+   block, then ~90 answer chunks; finish with `finish_reason:"stop"` then
+   `data: [DONE]`. Answer non-stream requests (title generation) with a
+   plain completion JSON.
+
+2. **Launch**: write `.env.local` (gitignored) with
+   `OPENAI_BASE_URL=http://127.0.0.1:<mockport>/v1`, `OPENAI_API_KEY=sk-mock`,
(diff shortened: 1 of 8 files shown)


Chirag123890:
Let go of my scroll and other stuffs. Help me Build productive business

PR #2431, opened 2026-07-17
fix: make .dockerignore ignore node_modules at any depth

### LINKED ISSUE #1920: Possible .dockerignore misconfiguration
Hello, and thank you for your work on this repository!

As part of my research, I am analyzing how developers configure .dockerignore in popular repositories.

During my analysis, I noticed that `/.dockerignore` might have been written under the assumption that .dockerignore and .gitignore follow the same pattern semantics, while they actually differ.[^1] In particular, the following pattern drew my attention:

https://github.com/huggingface/chat-ui/blob/ebeff50ac0ac4367a8e1a32b46dcc5ac2e8fc43f/.dockerignore#L7

According to Docker's specification, this pattern matches `node_modules` only in the top-level directory. (To ignore `node_modules` in any directory, the .dockerignore file needs to list `**/node_modules`)

Was this pattern written intentionally? If not, I'd be happy to submit a pull request to adjust it.

And if possible, could you tell me whether you were aware of the difference between .dockerignore and .gitignore? (I would like to know whether the differences are generally well recognized, or if they tend to slip developers' minds.)

Thank you again for maintaining this repository!

[^1]: See [.gitignore doc](https://git-scm.com/docs/gitignore), [.dockerignore doc](https://docs.docker.com/build/concepts/context/#dockerignore-files), and [a blog post about the differences](https://zzz.buzz/2018/05/23/differences-of-rules-between-gitignore-and-dockerignore/)

RudrenduPaul:
Closes #1920

## What this PR does

Fixes a `.dockerignore` pattern that was written assuming `.gitignore`
semantics, which Docker does not follow.

## Root cause

`.dockerignore` line 7 reads:

```
node_modules/
```

In `.gitignore`, a bare directory pattern like `node_modules/` (no leading
or embedded slash) matches that directory name at **any depth** in the
tree. Docker's `.dockerignore` parser does not carry over this behavior:
per the [Docker build context docs](https://docs.docker.com/build/concepts/context/#dockerignore-files),
a pattern without `**/` is anchored and only matches at the root of the
build context. So the current pattern only excludes the top-level
`./node_modules`, not any `node_modules` directory that might exist deeper
in the tree (e.g. under `scripts/`, `chart/`, or any subpackage added in
the future) — the opposite of what the pattern is clearly meant to do.

## Change

Single-line fix in `.dockerignore`:

```
# before
node_modules/

# after
**/node_modules/
```

The `**/` prefix is the documented Docker idiom for "match this pattern at
any depth," which restores the intended gitignore-like behavior the file
was already written to assume.

## Verification

- Read Docker's official `.dockerignore` syntax docs to confirm `**/` is
  required for depth-independent matching (unlike `.gitignore`).
- Confirmed via `grep`/directory listing that the rest of `.dockerignore`
  and the `Dockerfile` build context are unaffected by this change — no
  other rule relies on `node_modules/` staying root-only.
- No nested `node_modules` currently exists in this repo (single
  `package.json`, not a workspace/monorepo), so this change is a no-op for
  the current tree and only guards against future/incidental nested
  `node_modules` directories inflating the build context or getting
  copied into an image layer.

> **Note:** Claude Code was used to assist in drafting this fix. All changes were reviewed by the submitter.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,7 +4,7 @@ Dockerfile
 .gitignore
 LICENSE
 README.md
-node_modules/
+**/node_modules/
 .svelte-kit/
 .env*
 !.env


RudrenduPaul:
Just flagging for visibility: this is a single-line, narrowly-scoped fix (`.dockerignore` pattern only) that I've verified against Docker's own documented `.dockerignore` semantics. Happy to adjust scope, add tests, or close if this isn't a direction you want — let me know either way.

PR #2432, opened 2026-07-17
fix: allow addSibling to target the root message

### LINKED ISSUE #2386: fix: addSibling incorrectly rejects root message as sibling target
## Bug Description

`addSibling` in `src/lib/utils/tree/addSibling.ts` throws an error when called with the root message's ID as the sibling target:

```
Error: The sibling message is the root message, therefore we can't add a sibling
```

This is an unnecessary restriction. A sibling of the root message is a valid operation: the new message should be inserted with an empty `ancestors` array (same as the root), and since there is no parent to update, no parent's `children` list needs to change.

## Root Cause

The guard clause at lines 18–20 of `addSibling.ts` explicitly throws when `sibling.ancestors` is empty or undefined:

```ts
if (!sibling.ancestors || sibling.ancestors?.length === 0) {
  throw new Error("The sibling message is the root message, therefore we can't add a sibling");
}
```

However, if this block is removed, the existing logic already handles the root case correctly:
- `sibling.ancestors` is `[]`, so the new message is pushed with `ancestors: []`
- `nearestAncestorId` resolves to `undefined` (last element of empty array)
- `nearestAncestor` is not found, so the `children` update block is skipped

## Expected Behavior

`addSibling(conv, newMessage, conv.rootMessageId)` should succeed and return the ID of the newly created message, which has `ancestors: []`.

## Files Involved

- `src/lib/utils/tree/addSibling.ts` — contains the incorrect guard clause
- `src/lib/utils/tree/addSibling.spec.ts` — has a TODO comment acknowledging this should be fixed (line 50)

### LINKED ISSUE #2387: addSibling throws unnecessarily when target is the root message
## Bug

`addSibling` in `src/lib/utils/tree/addSibling.ts` throws an error when the caller tries to add a sibling to the conversation root message:

```
The sibling message is the root message, therefore we can't add a sibling
```

This restriction is unnecessary. A sibling of the root message is simply another top-level message with `ancestors = []`. The existing guard `if (nearestAncestor)` already handles the no-parent case correctly — it skips the children-update step, which is exactly right when there is no parent node to update.

## Affected files

- `src/lib/utils/tree/addSibling.ts` — lines 18-20, the premature throw
- `src/lib/utils/tree/addSibling.spec.ts` — line 50 carries a TODO noting this should be fixed

## Expected behavior

`addSibling(conv, newMessage, conv.rootMessageId)` should succeed and return a new message ID. The new message should be added to `conv.messages` with `ancestors = []` and `children = []`. No parent node exists, so no `children` array needs updating.

## Steps to reproduce

```ts
const siblingId = addSibling(conv, newMessage, conv.rootMessageId);
// throws: "The sibling message is the root message, therefore we can't add a sibling"
// expected: returns new message ID; conv.messages grows by 1; new message has ancestors = []
```

RudrenduPaul:
Closes #2386, Closes #2387

## What this PR does

`addSibling` in `src/lib/utils/tree/addSibling.ts` threw an error whenever
the sibling target was the conversation's root message:

```
Error: The sibling message is the root message, therefore we can't add a sibling
```

This restriction was unnecessary. A sibling of the root message is just
another top-level message with an empty `ancestors` array, and the rest of
the function already handles that case correctly.

## Root cause

The guard clause explicitly threw when `sibling.ancestors` was empty or
undefined:

```ts
if (!sibling.ancestors || sibling.ancestors?.length === 0) {
  throw new Error("The sibling message is the root message, therefore we can't add a sibling");
}
```

## Change

Removed the guard clause. With it gone, the existing logic already handles
the root case correctly:

- `sibling.ancestors` is `[]`, so the new message is pushed with
  `ancestors: []` (same as the root).
- `nearestAncestorId` resolves to `sibling.ancestors[sibling.ancestors.length - 1]`,
  i.e. `[][-1]`, which is `undefined`.
- `conv.messages.find(m => m.id === nearestAncestorId)` returns `undefined`
  since no message has `id === undefined`.
- The `if (nearestAncestor)` block is therefore skipped, which is exactly
  right: there is no parent node whose `children` array needs updating.

The non-root path is completely unaffected — the removed guard only ever
fired when `ancestors` was empty/falsy, so every other call path behaves
identically to before.

Also updated `addSibling.spec.ts`: the existing test asserted the old
throwing behavior (with a `// TODO: This behaviour should be fixed` comment
already acknowledging it needed to change). Replaced it with a test that
confirms `addSibling` now succeeds when targeting the root message and
returns a new message with the same (empty) `ancestors` as the root.

## Verification

- Traced the full function logic by hand for the root-message case to
  confirm no other code path re-throws or crashes once the guard is gone.
- `npx prettier --check` and `npx eslint` both pass clean on the changed files.
- Ran the full `addSibling.spec.ts` suite: 8/8 tests pass, including the
  updated root-message test and the pre-existing "should add a sibling to
  a message" test (confirms non-root behavior is unchanged).

> **Note:** Claude Code was used to assist in drafting this fix. All changes were reviewed by the submitter.

### DIFF
--- a/src/lib/utils/tree/addSibling.spec.ts
+++ b/src/lib/utils/tree/addSibling.spec.ts
@@ -47,16 +47,20 @@ describe("addSibling", async () => {
 		);
 	});
 
-	// TODO: This behaviour should be fixed, we do not need to fail on the root message.
-	it("should fail if the sibling message is the root message", async () => {
+	it("should add a sibling when the target is the root message", async () => {
 		const convId = await insertSideBranchesConversation();
 		const conv = await collections.conversations.findOne({ _id: new ObjectId(convId) });
 		if (!conv) throw new Error("Conversation not found");
 		if (!conv.rootMessageId) throw new Error("Root message not found");
 
-		expect(() => addSibling(conv, newMessage, conv.rootMessageId as Message["id"])).toThrow(
-			"The sibling message is the root message, therefore we can't add a sibling"
-		);
+		const rootMessage = conv.messages.find((m) => m.id === conv.rootMessageId);
+		if (!rootMessage) throw new Error("Root message not found in messages array");
+
+		const siblingId = addSibling(conv, newMessage, conv.rootMessageId as Message["id"]);
+
+		const sibling = conv.messages.find((m) => m.id === siblingId);
+		expect(sibling).toBeDefined();
+		expect(sibling?.ancestors).toEqual(rootMessage.ancestors);
 	});
 
 	it("should add a sibling to a message", async () => {
--- a/src/lib/utils/tree/addSibling.ts
+++ b/src/lib/utils/tree/addSibling.ts
@@ -15,10 +15,6 @@ export function addSibling<T>(conv: Tree<T>, message: NewNode<T>, siblingId: Tre
 		throw new Error("The sibling message doesn't exist");
 	}
 
-	if (!sibling.ancestors || sibling.ancestors?.length === 0) {
-		throw new Error("The sibling message is the root message, therefore we can't add a sibling");
-	}
-
 	const messageId = v4();
 
 	conv.messages.push({
--- a/src/lib/utils/tree/addSibling.ts
+++ b/src/lib/utils/tree/addSibling.ts
@@ -17,14 +17,16 @@ export function addSibling<T>(conv: Tree<T>, message: NewNode<T>, siblingId: Tre
 
 	const messageId = v4();
 
+	const siblingAncestors = sibling.ancestors ?? [];
+
 	conv.messages.push({
 		...message,
 		id: messageId,
-		ancestors: sibling.ancestors,
+		ancestors: siblingAncestors,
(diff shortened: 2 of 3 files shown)


RudrenduPaul:
Thanks @chatgpt-codex-connector for catching this — confirmed and fixed in the latest commit.

Root-level siblings had no parent to record them in a `children` array, so they were created but never appeared in any `messagesAlternatives` group, making them unreachable from the branch switcher UI. `createMessagesAlternatives` in `+page.svelte` now also emits a root-level alternatives group when more than one top-level message exists, so `ChatMessage`'s existing `alternatives={messagesAlternatives.find((a) => a.includes(message.id))}` lookup picks it up with no other UI changes needed.

While verifying this, `svelte-check` also caught that removing the original guard clause in `addSibling.ts` had silently dropped a TypeScript narrowing check on `sibling.ancestors` (2 new type errors) — fixed with an explicit `?? []` fallback. All 8 existing tests still pass.

PR #2437, opened 2026-07-20
Hide Top and Bottom buttons on mobile if loading

enzostvs:
While AI is thinking/replying hide the GoTop and GoBottom buttons to prevent glitching effect, especially for the bottom one


gary149 (approved):
didn't test

PR #2449, opened 2026-07-24
tighten SSRF protection

pngwn:
Hardens the server-side ssrf guard so raw IP literals and ipv6 edge-cases can no longer slip past the private-range blocklist before a request leaves the box. General hardneing and prep for mcp oauth.

- **up-front host check before connecting**: every request and redirect hop is now validated before the socket opens. Previously the check only ran in undici's dns lookup hook, which never fires for raw IP literals, so http://127.0.0.1 reached the network unchecked.
- **ipv4-mapped ipv6 fixed**: matched via the ::ffff:0:0/96 subnet instead of ip-address's is4(), which only recognised the dotted form and missed the canonical hex form (::ffff:7f00:1). Closes the ::ffff:127.0.0.1 / ::ffff:169.254.169.254 bypass.
- **ipv6 unspecified :: blocked**: routes to localhost, same as 0.0.0.0.
- **ipv6 unique-local fc00::/7 blocked**: the ipv6 counterpart of the private 192.168/172.16 ranges.


gary149:
this was discussed a lot with @rtrompier (adding him)

pngwn:
For reference, 10.0.0.0/8 which is used by internal load baancers is still allowed with this PR, and the existing tests for it still pass.

abidlabs:
You might want to whitelist hf-specific domains, like we do here:

https://github.com/gradio-app/gradio/blob/2e2a8baa1fb12100801c51d71af9627249eab307/gradio/processing_utils.py#L258

rtrompier (approved):

pngwn:
> You might want to whitelist hf-specific domains, like we do here:
>
> https://github.com/gradio-app/gradio/blob/2e2a8baa1fb12100801c51d71af9627249eab307/gradio/processing_utils.py#L258

@abidlabs all hf domains fall within 10/8, even with split DNS. This is secure: they are open at the app level and closed at the network level. If those domains ever fall outside of the allowed range then we can revisit.

PR #2451, opened 2026-07-24
prevent reattachment corruptions

pngwn:
Tests + fix for: https://github.com/huggingface/chat-ui/pull/2447#issuecomment-5068655342

Reattached generation streams could skip temporarily invisible events or resume past an event that later grew, causing dropped or duplicated text. The stream now advances only through contiguous events, and captured writer cursors seal pending events so persisted snapshots and replay logs remain consistent.

### DIFF
--- a/src/lib/server/generation/writer.spec.ts
+++ b/src/lib/server/generation/writer.spec.ts
@@ -1,5 +1,8 @@
-import { describe, it, expect } from "vitest";
-import { mergedStreamToken } from "./writer";
+import { afterEach, beforeAll, describe, it, expect } from "vitest";
+import { ObjectId } from "mongodb";
+import { randomUUID } from "node:crypto";
+import { collections, ready } from "$lib/server/database";
+import { createGenerationWriter, mergedStreamToken } from "./writer";
 import {
 	MessageUpdateType,
 	MessageReasoningUpdateType,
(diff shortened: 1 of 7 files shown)


pngwn:
@gary149 could you retest with the same example as the comment and see if the problem is addressed by this pr pls.

gary149:
Retested on this branch (aa66292e, checked out locally, new unit tests passing) with the same methodology as the original report: long generation with moonshotai/Kimi-K2.6, conversation opened in a second tab mid-run, second tab's rendered text sampled every second and each sample diffed against the final persisted text.

**The corruption still reproduces, though it is smaller than before.**

- Final persisted text: "north of **Greenland's** fractured coast"
- All mid-run samples of the reattached tab: "north of **Greenland** fractured coast" (the `'s` dropped at position 874 and stayed missing for the entire live view; self-healed at completion as usual)

The diagnostic detail: the very first sample (taken ~4s after the tab opened, 876 chars long) still **had** the correct "Greenland's", and it ends exactly at those characters. Later samples lost the last 2 characters of that initial snapshot and continued correctly with " fractured coast...". An appending consumer cannot delete mid-text characters, so the message content must have been replaced mid-run with a state that disagreed with the replay cursor by exactly one partial token, right at the page-load boundary. I saw the same "earlier sample had text a later sample lost" pattern in a pre-fix run (that time a 16-char drop at the boundary of an earlier snapshot).

So the two server-side fixes here look right (this run showed one 2-char artifact where pre-fix runs showed several larger ones), but a third seam appears to survive on the client: when conversation data is refreshed mid-run, the reattach is aborted and reopened from the new message's `materializedSeq` (the effect around `+page.svelte:601` / `reattachToRun`), and the freshly loaded content and that cursor can disagree by a partial coalesced token. Worth checking how the load-time `content` / `materializedSeq` pair can get out of sync with each other on the reader side, or making the client tolerant (e.g. trim/verify overlap against the first replayed event instead of trusting the cursor blindly).

Repro is cheap if you want to iterate: ask for a 2000-word story, open the conversation in a second tab while it streams, and diff the second tab's text against the final text after completion. Roughly one artifact per run at current rates, always surviving until the end-of-run merge repairs it.

pngwn:
Ah interesting, it wasn't the page refresh specifically. The problem was merging was compressing tokens and losing the correct cursor position. Pushed fix, tested locally, and added a test.

pngwn:
Will merge this for more testing, will follow up if there are any other issues you notice.

PR #2505, opened 2026-08-17
feat(mcp): add resources support

pngwn:
MCP servers can expose read-only **resources** (files, documents, records) alongside their tools. chat-ui listed and called tools but never touched resources, so half of every connected server's surface was dark.

## Approach

Resources are model-controlled, reached through **two synthetic functions** merged into the OpenAI tool array:

- `list_mcp_resources` — enumerates URIs, names, descriptions and media types across enabled servers
- `read_mcp_resource` — reads one by URI

Two functions rather than one per resource: a server may expose hundreds, and they would crowd its real tools out of the array sent to the provider. Neither is bound to a single server — `read` resolves a URI's owner at call time (enumerated resource first, then a matching URI template) and reports an unresolvable URI back to the model rather than broadcasting a round trip to every connected server.

## Listing shares one connection

Tools and resources are now listed together over a **single connection per server** and share a cache entry — same key, same TTL. Listing them separately would have doubled the connection churn the listing cache exists to avoid. Servers that never declare the `resources` capability are not asked for them, so nothing changes for them but the log line.

## Guards

- **Pagination** is followed to exhaustion within bounds (10 pages / 250 resources per server).
- **Read output is capped** at 32k chars with an explicit truncation marker — a resource is arbitrary server data, and a whole log file is a legitimate one.
- **Binary contents are described, not inlined.** base64 the model cannot read would otherwise spend the context window on bytes.
- **A broken resource listing never costs a server its tools** — that is what the caller is primarily after.
- Reads reuse the pooled client with the same reconnect-on-closed handling as `callMcpTool` (detectors are now exported from `httpClient.ts` instead of duplicated).

## Behaviour change worth flagging

The health check previously marked a server ready whenever the response carried a `tools` field — including `tools: []`, since an empty array is truthy. It is now ready if it has **tools or resources**, and `listTools` is allowed to fail, because a resources-only server is valid per spec and would otherwise fail the health check outright. A server with neither now gets a 503 saying so, which is what the original message already intended.

## UI

Server cards show a resource count next to the tool count and an expandable list.

## Manual testing

`scripts/mock-resources-mcp.mjs` is a throwaway MCP server exposing one tool and four resources, one per branch of the read path: a text file, a URI template, a binary blob, and a 44k-char log.

1. Set `MCP_ALLOW_INSECURE_URLS=true` in `.env.local`, or chat-ui rejects the loopback URL.
2. `node scripts/mock-resources-mcp.mjs` (listens on 8793; override with `MOCK_RESOURCES_MCP_PORT`).
3. Start the app, **log in** — the health check is behind auth, and resources only populate after one.
4. Add `http://127.0.0.1:8793/mcp` in the MCP panel, enable it, hit **Health Check**.

Expected — the card reads `1 tool · 4 resources` with an expandable list. Then ask **"what's the rollback codeword?"**: the answer should contain `PELICAN-7734`, which exists only inside `file:///briefing.md`, so it can't be produced without a real read. The mock logs `read <uri>` on every read, which distinguishes "never called" from "called and failed".

Worth trying alongside: `note:///alpha` (routes via a template, never enumerated), `image:///logo.png` (described, not inlined as base64), `file:///huge.log` (truncated at 32k), and a made-up URI (rejected without contacting any server).

### DIFF
--- a/src/lib/server/mcp/resources.ts
+++ b/src/lib/server/mcp/resources.ts
@@ -7,6 +7,8 @@ import { logger } from "$lib/server/logger";
 /** A resource is arbitrary server data; unbounded, one read can swallow the context window. */
 const MAX_RESOURCE_TEXT_CHARS = 32_000;
 const MAX_LISTED_RESOURCES = 200;
+/** Templates get their own budget so a long resource listing can never starve them out. */
+const MAX_LISTED_TEMPLATES = 50;
 
 export type McpResourceReadResult = {
 	text: string;
@@ -46,21 +48,19 @@ function describeResource(server: string, resource: CachedServerResource): strin
 	return resource.description ? `${line}\n    ${resource.description}` : line;
 }
 
-/** No match is reported back, never broadcast: a typo must not cost a round trip per server. */
-function resolveOwner(
+/** Every server that could serve `uri`; an enumerated match outranks a template one. */
+function resolveOwners(
 	servers: McpServerConfig[],
 	catalogs: ServerCatalog[],
 	uri: string
-): McpServerConfig | undefined {
-	for (const [index, server] of servers.entries()) {
-		if (catalogs[index].resources.some((resource) => resource.uri === uri)) return server;
-	}
-	for (const [index, server] of servers.entries()) {
-		for (const template of catalogs[index].templates) {
-			if (templateToRegExp(template.uriTemplate)?.test(uri)) return server;
-		}
-	}
-	return undefined;
+): McpServerConfig[] {
+	const enumerated = servers.filter((_, index) =>
+		catalogs[index].resources.some((resource) => resource.uri === uri)
+	);
+	if (enumerated.length > 0) return enumerated;
+	return servers.filter((_, index) =>
+		catalogs[index].templates.some((template) => templateToRegExp(template.uriTemplate)?.test(uri))
+	);
 }
 
 export async function listMcpResources(
@@ -85,8 +85,11 @@ export async function listMcpResources(
 	}
 
 	const templateLines: string[] = [];
+	let templateTotal = 0;
 	for (const [index, server] of servers.entries()) {
 		for (const template of catalogs[index].templates) {
+			templateTotal += 1;
+			if (templateLines.length >= MAX_LISTED_TEMPLATES) continue;
 			const parts = [`- ${template.uriTemplate}`];
 			if (template.name) parts.push(`name: ${template.name}`);
 			if (template.mimeType) parts.push(`type: ${template.mimeType}`);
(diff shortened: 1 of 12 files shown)


abidlabs:
Hey @pngwn I tested this with an MCP server that returned a basic text resource and resource template and it worked well: https://huggingface.co/spaces/gradio/mcp_resources_and_prompts_main. 

But one thing I noticed is that this PR does not try to render MCP resources that are non-text (e.g. image, html)?  

I'm not sure if we will need it for MCP, so I'll go ahead and LGTM this, but I think generally MCP clients do check for mime type and handle the resource appropriately. If you want, you can test with this MCP server which returns a html resource: https://huggingface.co/spaces/gradio/mcp_letter_counter_app_main/blob/main/run.py

abidlabs (approved):

pngwn:
Yeah. I just didn't really know what to do with images or html. There are a bunch of options and I didn't want to spend time on it when i just don't think the usage is there right now. MCP servers using resources are pretty rare. 

Happy to revisit if we think it is important though.

src/lib/server/mcp/tools.ts
```diff
@@ -203,18 +362,21 @@ async function listServerTools(
 
 		const response = await client.listTools({});
```
evalstate: This throws "Server does not support tools (required for tools/list)" for "resource only" servers. (the health endpoint handles this case with "[MCP Health]" check)

src/lib/server/mcp/resources.ts
```diff
@@ -0,0 +1,194 @@
...
+			total += 1;
+			if (lines.length >= MAX_LISTED_RESOURCES) {
+				omitted += 1;
+				continue;
+			}
+			lines.push(describeResource(server.name, resource));
+		}
+	}
+
+	const templateLines: string[] = [];
```
evalstate, line 87: Templates here are unbounded -- could cause unbounded tool result.

pngwn:
Correction: html should work fine as it is just text. Only binary blobs aren't supported.

evalstate (commented):
Honest feedback is I would feature flag this.

Main usage in the wild at the moment is for MCP Apps - and in that case you don't want the extra 2 tools added to context, or that content made available to the LLM (unless you are in a contorted debug scenario).

The other - Skills over MCP - would be better served by abstracting a specific Skills tool/flow for the Harness (although there is value in exposing skills).

Finally, of we want Resource Template handling it should handle namespace collisions (where 2 MCP Servers use the same template name). Current handling wouldn't be robust for complex template paths (or handle ?*+ type operatores). Honestly, I would remove support for Templates unless you have a specific MCP Server you want to use them with. If we do have the use I think it's worth adding UX for the User to handle templates and Completions. I think GH might be the only server in the wild to do this now.

PR #2506, opened 2026-08-17
Add MCP elicitation support

pngwn:
Lets an MCP server interrupt a tool call to ask the user something — a confirmation, details it's missing, or a link to visit. The request renders as a form in the conversation and the answer goes back to the server, which then finishes the call. Form and URL mode both supported.

Built on the multi-round-trip request pattern (SEP-2322) that arrived with the TS SDK v2 bump, with the 2025-era push style kept working for servers that haven't moved.

## Two eras, one handler

**2026-07-28 — nothing stays live.** The server answers `tools/call` with `input_required` instead of sending a request, so the call is already over by the time the prompt appears. The prompt, the tool, its arguments and the server's opaque `requestState` go into Mongo; when the answer arrives the call is re-issued with `inputResponses` + `requestState`. Neither side holds state in between, so a prompt survives a reload, a redeploy, and the connection being closed and reopened — and it can be answered on a different pod than the one that raised it. A tool may ask more than once (`double_check` does); each round parks the same way rather than handing the model a round that never finished.

**2025-era — the call blocks.** The server sends `elicitation/create` down the connection and waits, so the prompt lives exactly as long as that request. The UI shows a countdown here and none on the modern path, where there is nothing left to expire.

The SDK normalizes both into the same request shape, so there is one `elicitation/create` handler rather than a path per era.

## Notable bits

**Attribution and isolation.** MCP gives a server-initiated request no link back to the `tools/call` that provoked it — there is no correlation id on the client handler in either era — so the only clue is which calls are in flight on that connection. `clientPool` shares one client per URL + headers, which makes that clue ambiguous as soon as two conversations are mid-call, and an unattributable prompt is declined rather than shown to the wrong person. To stop that ambiguity arising at all, a call that can be interrupted for input gets a **per-conversation connection when the server is legacy**; modern servers push nothing, so they keep one shared connection each. Alongside that: every `mcpElicitations` query is scoped by `conversationId`, the client-side resume signal carries the conversation it belongs to, and evicting a client that looked broken to one caller no longer closes it under another's in-flight call.

**Answers go through Mongo, not the stream**, since the pod serving the user's POST need not be the pod holding the tool call — the same split `abortedGenerations` already uses. Side benefit: reloading mid-prompt re-renders a form that still works.

**The call deadline moved out of the SDK.** `Protocol.request` arms a wall-clock timer that only a server-sent progress notification can reset, so a human filling in a form would expire the call waiting on them. It's now enforced via the abort signal and paused while a prompt is open, which puts `MCP_TOOL_TIMEOUT_MS` back to meaning "time the server may go silent".

**Server-authored input is untrusted throughout.** The six spec spellings of a select box normalize to one field kind, field/option/text counts are capped, control characters are stripped from labels, non-http(s) URLs are refused, and the submitted answer is re-validated server-side against the stored schema.

**Answered prompts collapse.** A settled prompt becomes an expandable row — `Answered` / `Declined` / `Expired unanswered` and the server name — that opens to show the questions and what was sent, instead of leaving a filled-in form sitting in the transcript. It reads the same on reload, since the submitted values are persisted on the resolved update.

`extra.signal` is honoured, so a legacy server that gives up on its own request (60s by SDK default) closes the prompt instead of leaving a form that can't be answered.

## Config

- `MCP_ELICITATION_TIMEOUT_MS` — how long a prompt stays answerable, default 1 hour. Independent of the tool timeout. On a modern connection this is the *only* bound, since nothing else is waiting; on a legacy one the server's own timeout usually gives up first.
- `MCP_DISABLE_ELICITATION=true` — kill switch, stops advertising the capability entirely.

## Manual testing

`scripts/mock-elicitation-mcp.mjs` serves a 2026-era server via `createMcpHandler`, with one tool per case worth seeing.

1. Set `MCP_ALLOW_INSECURE_URLS=true` in `.env.local`, or the loopback URL is rejected
2. `node scripts/mock-elicitation-mcp.mjs` (port 8792)
3. Add `http://127.0.0.1:8792/mcp` as a user server, enable it, pick a tools-capable model
4. Ask for each tool by name:

| Tool | Expected |
| --- | --- |
| `book_meeting` | Form covering every field kind — text, email, date, number, select, multi-select, checkbox |
| `double_check` | Two prompts in one call, the second appearing after the first is answered (type the word it asks for — a mismatch is meant to fail) |
| `impatient_confirm` | The pre-MRTR style. Fails fast here, since a modern server cannot raise an unsolicited request; point a 2025-era server at it to exercise the "server stopped waiting" path |
| `sign_in` | URL mode — a link, not a form |

Also worth trying: reload the page while a prompt is open (comes back answerable), answer one from a second tab, and stop the response while one is open (closes as cancelled).

## Notes

- 62 tests across schema normalization, answer validation, the pausable deadline, resume, routing and pool isolation. Full suite green — 700 node, 100 browser.
- Declaring the capability changes behaviour for existing deployments — servers that previously got method-not-found will start raising prompts. Defaulted on, since the capability *is* the feature; `MCP_DISABLE_ELICITATION` is the lever to stage it.
- URL mode responds `accept` once the user confirms they've visited the link. The out-of-band `notifications/elicitation/complete` path isn't wired up, so a server relying on it to auto-close the prompt will see it stay open until confirmed or expired.
- Only one input request per round is accepted. The spec permits several, but no server we have met asks for more, and half an answered round is not resumable.

### DIFF
--- a/.env
+++ b/.env
@@ -149,8 +149,9 @@ MCP_DISABLE_ELICITATION=
 # How long (ms) an elicitation prompt stays answerable (default: 3600000 = 1 hour), so
 # someone who walks away can still answer when they get back. Independent of
 # MCP_TOOL_TIMEOUT_MS — the tool call's deadline is paused while the prompt is open.
-# In practice the MCP server's own timeout usually gives up sooner; this mainly bounds a
-# run that is never answered and never stopped, which would otherwise never finish.
+# On a 2026-era connection this is the ONLY bound: the server answered with
+# `input_required` and is no longer waiting, so nothing else expires. On a 2025-era one
+# the server's own timeout on its request (60s by SDK default) usually gives up first.
 MCP_ELICITATION_TIMEOUT_MS=
 # How often (ms) to sweep for generations whose pod died mid-run and mark them
 # interrupted. Default 60000.
--- a/.env
+++ b/.env
@@ -137,11 +137,21 @@ MCP_SERVERS=
 MCP_FORWARD_HF_USER_TOKEN=
 # Exa API key (injected at runtime into mcp.exa.ai URLs as ?exaApiKey=)
 EXA_API_KEY=
-# Timeout in milliseconds for MCP tool calls (default: 120000 = 2 minutes)
+# How long an MCP tool call may go without the server responding (default: 120000 = 2
+# minutes). Time spent waiting on a user to answer an elicitation does not count against it.
 MCP_TOOL_TIMEOUT_MS=
 # Set to "true" to allow MCP server URLs over http on localhost / private LAN addresses.
 # Local development and the e2e harness only — never enable in production.
 MCP_ALLOW_INSECURE_URLS=
+# Set to "true" to stop advertising the `elicitation` client capability, so servers can
+# never interrupt a tool call to ask the user for input.
+MCP_DISABLE_ELICITATION=
+# How long (ms) an elicitation prompt stays answerable (default: 3600000 = 1 hour), so
+# someone who walks away can still answer when they get back. Independent of
+# MCP_TOOL_TIMEOUT_MS — the tool call's deadline is paused while the prompt is open.
+# In practice the MCP server's own timeout usually gives up sooner; this mainly bounds a
+# run that is never answered and never stopped, which would otherwise never finish.
+MCP_ELICITATION_TIMEOUT_MS=
 # How often (ms) to sweep for generations whose pod died mid-run and mark them
 # interrupted. Default 60000.
 GENERATION_REAP_INTERVAL_MS=
--- a/scripts/mock-elicitation-mcp.mjs
+++ b/scripts/mock-elicitation-mcp.mjs
@@ -196,6 +196,182 @@ function buildServer() {
 		}
 	);
 
+	server.registerTool(
+		"edge_defaults",
+		{
+			description:
+				"Schedules a reminder. Defaults arrive as RFC 3339 and two switches are optional.",
+		},
+		async (...args) => {
+			const ctx = args.at(-1);
+			const given = acceptedContent(ctx?.mcpReq?.inputResponses, "when");
+			if (!given) {
+				console.log("[mock-elicitation-mcp] edge_defaults -> input_required");
+				return inputRequired({
+					inputRequests: {
+						when: inputRequired.elicit({
+							message: "When should this run, and how loudly?",
+							requestedSchema: {
+								type: "object",
+								properties: {
+									// RFC 3339 is what the schema asks servers to send, but a date input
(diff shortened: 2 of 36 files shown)


abidlabs:
Can you remind me @pngwn why elicitation support is needed? Does the HF MCP use elicitation or are we planning on adding tools with elicitation?

pngwn:
Because Shaun said they might use it in the hf mcp server (though it isn't used currently, I didn't know this when started it), it's a pretty basic mcp feature, and we can use part of the same code path to ask the user questions like Claude code/ codex do (although we won't do this via mcp specifically we can reuse some of the UX).

abidlabs (approved):
I tested with https://github.com/mcp-use/mcp-elicitation-demo and didn't run into any issues. Nice stuff @pngwn!

evalstate:
> Can you remind me @pngwn why elicitation support is needed? Does the HF MCP use elicitation or are we planning on adding tools with elicitation?

Yep, Elicitation support is fairly widespread: https://huggingface.co/spaces/evalstate/mcp-adoption -- and with the new MRTR architecture becomes viable for remote MCP Servers - so for high stakes actions and secondary authorisations with URL Elicitations this is a very good bet.

evalstate:
Love this! Great UX.

Some pathological testing (specific issue in each screenshot):

02-untouched-optional-boolean-submits-false
03-maxitems-enforced-only-after-submit
04-proto-field-removed-before-render
01-rfc3339-default-renders-blank

The boolean and date are definitely worth addressing IMO

evalstate:
Adding a couple more for URL Elcitiations:


Would be nice to warn on no HTTPS / local URIs.

05-http-private-network-url-has-no-warning

Spec says MUST provide clear Decline and Cancel options; only Decline is present.

01-no-explicit-cancel-control

clients SHOULD warn about ambiguous or suspicious URIs, explicitly including Punycode.
03-punycode-hostname-has-no-warning

abidlabs:
Hmm I'm seeing a weird issue where now the tool calls are all getting canceled before I have a chance to submit anything:


https://github.com/user-attachments/assets/6fa34ded-94b7-4320-8136-1b51e4e0998e

abidlabs:
Ok it doesn't happen consistently, but I think it happens if there's an active elicitation in another chat session:


https://github.com/user-attachments/assets/0733ffb3-13fd-4fdf-8558-aff5f4fe1cd2

abidlabs (approved):
LGTM @pngwn! 

Btw I noticed that all my chat titles are "We need to produce a title....". I don't think it's due to this PR, but I wonder if it's related to other tool calling / reasoning in the history PR you had.

Image

pngwn:
@abidlabs yeah i noticed that as well, will take a look and make a new PR.

PR #2513, opened 2026-08-20
Add an ask_user_question tool

pngwn:
Lets the model put a decision to the user as options they can click, instead of guessing or asking in prose. It calls `ask_user_question`, the run parks, and the answer comes back to it as the tool result.

Builds directly on the elicitation work in #2506 — the hard part there was never MCP-specific.

## No new mechanism

Resume already worked by persisting a tool-result update and letting history replay feed it back to the model, so a native tool is a different branch at two points rather than a parallel system:

- **Dispatch** (`toolInvocation.ts`) — the call is intercepted before the MCP mapping lookup, records a durable prompt and parks the run.
- **Resume** (`resumeElicitation.ts`) — nothing is re-issued, because the stored answer *is* the result.

Everything else is reused as-is: the durable record, the answer endpoint, the park/resume plumbing in the conversation route, and the settled transcript row.

Each question normalizes to one `select` field, so its validation and the collapsed summary are the ones already there. `pending` gains a `kind` discriminator rather than the collection being reshaped, so rows written before this read as MCP and there is no migration. The collection is still called `mcpElicitations` while serving both — worth renaming, but not in the change that introduces the second use.

## The UX is modelled on the harness tools, not on a form

An open question is a decision the run is waiting on, so it is **not part of the conversation**. It leaves the transcript entirely: a store hands it to the composer, which draws it as a panel above the input — a sibling of the scroll container, so it cannot scroll away or be scrolled past.

- **Options are rows, not a dropdown.** One click on the text, the model's explanation under each label, and a radio or checkbox marker depending on `multiSelect`.
- **One question at a time**, with a counter and Back, rather than a wall of fields.
- **A single answer carries itself forward.** Picking one advances on its own, first time and when an earlier answer is changed, so there is usually no Next button at all — it appears only where a click cannot finish the question: several answers, or one being typed. Sending is never automatic, however the last question is answered.
- **"Something else…"** reveals a text box, because the options are the model's guesses. The typed value replaces the marker before sending, so nothing internal reaches the server or the transcript, and submitting it blank is refused rather than silently dropping a required answer.
- **Answering settles it** into the same collapsed row an MCP prompt leaves behind, which is the part worth keeping in the stream.

Visually the rows carry no border or fill of their own — they were boxes inside the panel's box. Selection reads from the marker, the label's weight, and a hover wash, with a rule above and below the group so they sit in a band rather than floating. The row beneath holds one height throughout, so the panel does not grow when Send appears.

Selects otherwise only accept values from their own option list, so the validator gained an explicit, bounded exception. MCP fields never set the flag, so server-driven prompts stay closed to exactly what the server offered.

## The prompt was the reason it never fired

Worth calling out, because no tool description was going to fix it. The existing tool preprompt says:

> `IMPORTANT: Do NOT call a tool unless the user's request requires capabilities you lack … When in doubt, do not use a tool.`

Asking the user is neither a missing capability nor external information, so that instruction sat directly on top of the new tool. This carves it out and adds positive guidance on when to ask. Both are conditional on the tool being present, so a deployment with the kill switch set gets a byte-identical preprompt — there is a test asserting that.

## Gating

Offered only to conversations running the **ML Assistant preset** from #2520, and only where MCP tools already exist. That mode is a build flag, so a build either ships this or has no way to reach it — no separate runtime switch, which would have contradicted how the preset is gated.

## Notes

- **Still requires MCP tools to exist**, on top of the mode. It is added after the zero-tool bail, so a conversation with no MCP server keeps the plain generation path rather than being pulled onto the tool-calling one. In practice the preset supplies its own server, so this mostly matters for a mode conversation whose listing failed. A follow-up lifts the requirement: the dispatch here is being extracted into a shared builtin-tool seam that counts builtins in that gate. Sequenced deliberately, so the abstraction is shaped where a second consumer exists rather than guessed at here.
- **Firing is deliberately conservative.** The guidance tells the model to prefer picking sensibly and saying what it assumed. A request that reads as conversational, or that asks for options *as the answer*, correctly does not trigger it.
- 22 tests dedicated to it — normalization and the answer→result mapping, and the panel's behaviour — plus additions to five existing suites for the "Other" path, the preprompt carve-out, the tool-offering gate, and the three wiring points. Full suite green: 762 node, 144 browser.
- The wiring tests were checked by breaking each mechanism and confirming the right ones fail — dispatch, resume, and the transcript/composer split.
- Driven end to end against a live model on desktop. **Not verified on mobile** — the panel's fit against the composer at small breakpoints has not been reviewed.


abidlabs:
> Only offered when MCP tools already exist. It is added after the zero-tool bail, so a conversation with no MCP server keeps the plain generation path. Making it available standalone means engaging the tool-calling flow for every conversation, which is a much larger behaviour change than this warrants — but it does mean no MCP server, no clarifying questions. A follow-up lifts this: the dispatch here is being extracted into a shared builtin-tool seam, which counts builtins in that gate. Sequenced deliberately, so the abstraction is shaped where a second consumer exists rather than guessed at here.

Oh I got caught by this. I was wondering why the chat UI didn't seem to have access to the tool unless I added an MCP server, which seems unrelated to this capability. I don't think this will trip up users since they wouldn't necessarily expect this tool to exist, but it did confuse me as a review. Once I added an MCP server, everything worked as expected. 

Small UI nit: in the "something else" textbox, clicking enter/return should submit the form

abidlabs (approved):

PR #2517, opened 2026-08-21
Let users toggle planning per model

pngwn:
Stacked on #2516 (targets `harness-update-plan-ui`; retarget as the stack merges). Final piece of the plan-tool stack — the enablement surface.

## What this does

- **`planningOverrides` setting** — per-model `Record<string, boolean>`, persisted and round-tripped through both settings endpoints. Patterns after `artifactsOverrides`: planning is prompt/protocol-level, not provider-determined, so the override is **not** wiped on HuggingChat (unlike `toolsOverrides`).
- **Threading** — `planningOverride` joins `TextGenerationContext`, flows through `runMcpFlow` into the builtin-tool registry, where the resolution chain is now: user setting → per-model `supportsPlanning` → inferred default (`PLANNING_ENABLED` + `supportsTools`, router alias excluded). The model-default half of the chain is extracted as `resolvePlanningDefault()`.
- **Client model payload** — both model serializers now emit `supportsPlanning` as the *resolved* default rather than the raw flag, so the settings page shows each model's effective state without the client knowing about the env switch.
- **Settings UI** — a "Planning" switch on the per-model settings page, mirroring the Artifacts row (same Beta pill, enabled on HuggingChat): "Let the model keep a visible task plan for complex, multi-step work."

## Tests
Registry: user override beats the resolved default in both directions (including with the env switch off). Flow: `planningOverride: false` withdraws the tool even with `PLANNING_ENABLED=true`. Full suite green (772); `npm run check` 0 errors; svelte autofixer clean on the settings page change.


pngwn:
Closing: superseded by the ML Assistant gate. Peter asked for the plan tool to be gated behind ML Assistant mode (#2520), which replaces this PR's entire enablement surface — `PLANNING_ENABLED`, `supportsPlanning`, `planningOverrides`, and the per-model settings toggle are all gone in favor of `isMlAssistantConversation` in the builtin-tool registry (see the updated #2514). A per-model or per-user knob can be revisited if the mode ever needs one.

PR #2521, opened 2026-08-24
Feed the ML Assistant plan strip from update_plan

pngwn:
Stacked on #2516 (targets `harness-update-plan-ui`; retarget as the stack merges). Wires the plan tool into the seam #2520 left for it: the `MlAssistantPlanProgress` dots strip now has a plan source.

## What this does

- **`planStepsToMlSteps`** (`src/lib/utils/planProgress.ts`): maps `PlanStep[]` onto the strip's `MlPlanStep` shape — statuses onto its lifecycle (`in_progress`→`running`, `completed`→`done`), the full step text as the dot's tooltip description, and a short word-boundary cut as the dot label / running status label. The plan schema carries one string per step, so labels are derived; a model-authored label field can supersede this later if dogfooding wants prettier dots.
- **Streaming**: `consumeMessageUpdates` gains an optional `onPlan` callback fired on every `MessageUpdateType.Plan` snapshot — one wiring point shared by the send and reattach paths, so both render identically. The conversation page passes `onPlan` → `mlAssistant.setPlan(...)`, guarded by the `ML_ASSISTANT_MODE` build constant so the branch folds out of flag-off builds.
- **Resume mid-plan**: the conversation payload now carries `plan` (the persisted `Conversation.plan` snapshot), and ChatWindow's existing `syncConversation` effect seeds the strip from it when a mode conversation is (re)opened — only when `syncConversation` actually reset, so live streamed steps are never clobbered and the adopt-after-create path keeps its streamed state. Exactly the wholesale-replacement semantics the store's `setPlan` documents.
- A run that hasn't reported a plan keeps the strip on its tool note (no placeholder emitted), per the strip's designed degradation.

## Tests
`planProgress.spec.ts` covers the status mapping, tooltip fidelity, and label cutting. Full suite green (795); `npm run check` 0 errors; svelte autofixer clean on the touched code.

### DIFF
--- a/src/lib/components/chat/ChatWindow.svelte
+++ b/src/lib/components/chat/ChatWindow.svelte
@@ -58,6 +58,8 @@
 	import MlAssistantStrip from "./MlAssistantStrip.svelte";
 	import { ML_ASSISTANT_MODE } from "$lib/utils/mlAssistantFlag";
 	import { mlAssistant } from "$lib/stores/mlAssistant.svelte";
+	import { planStepsToMlSteps } from "$lib/utils/planProgress";
+	import type { PlanState } from "$lib/types/Plan";
 	import {
 		ML_ASSISTANT_EFFORT,
 		ML_ASSISTANT_PLACEHOLDER,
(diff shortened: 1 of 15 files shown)


pngwn:
Added a second commit addressing the label-quality issue the ML Assistant work flagged: a word-boundary cut of the step text made the dot's accessible name a fragment ("Read the — running"), spammed the aria-live row with the same fragment on each transition, and made the tooltip's bold lead a literal prefix of its own description. `update_plan` now carries a model-authored one-or-two-word `label` per step — advertised as required in the tool schema so models supply it, validated leniently so pre-field plans and label-less updates still parse — and `planStepsToMlSteps` prefers it, keeping the truncation only as a fallback. `statusLabel` is produced from the same value on this side, so #2520's type needs no change for this to work (making `statusLabel` optional there remains a nice-to-have cleanup).

PR #2540, opened 2026-08-31
fix: honor PUBLIC_ORIGIN when building the OAuth callback redirect URI

### LINKED ISSUE #2489: Serving over HTTP (not https) broken in docker container
## Bug description

This was a really tricky one to track down. The crux of the bug is that in the most vanilla, default install of chat-ui using a Docker container, accessing chat-ui over HTTP only is broken out of the box. What I mean by broken is that any rendered content that relies on `page.url.origin` will have it's connection protocol rendered as HTTPS instead of HTTP. This results in broken images, such as fave-icons, apple touch icons, omni-welcome images, and logos. More critically, however, is that oAuth callback_uri's are also incorrect, causing logins to fail due to invalidate callback uri's in the upstream OIDC provider.

The reason for the unintentional protocol render is the node-adpater for Svelt. The issue is documented [here](https://github.com/sveltejs/kit/issues/3248), [here.](https://github.com/sveltejs/kit/issues/10078), and ultimately indirectly fixed [here.](https://github.com/sveltejs/kit/pull/3384).

Also: Running the app from source code with the Vite dev build to try to reproduce the issue doesn't work because in dev move the node-adapter doesn't assume/force HTTPS. It must be tested on the "production" build. That was a fun one to realize.

## Steps to fix

Add an environment variable simply called "ORIGIN" to match the exact hostname. This will inform the node-adapter of what protocol to render as, and not assume HTTPS for all non-development builds of the app.

## Steps to reproduce

Deploy a fresh Docker container for chat-ui, configure it with:

```env
OPENAI_BASE_URL=https://********
OPENAI_API_KEY=********
OPENID_CONFIG={PROVIDER_URL: "****",CLIENT_ID: "chatui",CLIENT_SECRET: "****",SCOPES: "openid profile email"}
COOKIE_SECURE=false
#PUBLIC_ORIGIN=http://mylocaldocker.internal:5173/ # Note: This doesn't appear to work as expected.
```
Access the site over insecure http, observe the broken functionality, icons, and inability to oAuth.
To fix, add:

```env
ORIGIN=http://mylocaldocker.internal:5173/
```

## Screenshots

Image

After setting ORIGIN:

Image

## Notes

For some reason, PUBLIC_ORIGIN has a detrimental effect, causing 404's for rendered assets, but crucially, has **zero** effect on the incorrect change of the protocol. Assets and oAuth callback uri's are sill assuming HTTPS.

Also not explored: I did not check to see if PUBLIC_ORIGIN is supposed to also translate to the [ORIGIN](https://svelte.dev/docs/kit/adapter-node#Environment-variables-ORIGIN-PROTOCOL_HEADER-HOST_HEADER-and-PORT_HEADER) header, and this is the _real_ underlying issue.

Now for the obligatory acknowledgements: Of course, don't do this for production workloads. My use-case is an isolated testing lab with internal certificate authority, OIDC provider, and web host purely for testing and anlaysis. Of course, nobody should ever oAuth over insecure HTTP. However, if the chat-ui documentation suggests it supports an http-only scenario (by use of insecure cookies and PUBLIC_ORIGIN, one would assume the intent is that it work over HTTP after being configured a such.

And lastly, this is probably not an issue for 90% of users who properly deploy an https reverse-proxy in front of this container, which of course I do for for my real/production instance.

## Recommendation:

Here's a few possible avenues to fix:

* A simple documentation change to the .env example file to include ORIGIN.
* A code change that PUBLIC_ORIGIN also sets ORIGIN internally.
* Update [auth.ts#L533](https://github.com/huggingface/chat-ui/blob/main/src/lib/server/auth.ts#L553) and change `url.origin` to `config.PUBLIC_ORIGIN`

revanth-045:
Fixes #2489

## Problem

`triggerOauthFlow` builds the OAuth `redirect_uri` from `url.origin` directly:

```ts
let redirectURI = `${url.origin}${base}/login/callback`;
```

SvelteKit's node adapter renders `url.origin` as `https://` in production builds regardless of the actual inbound protocol, unless its own `ORIGIN` env var is set ([sveltejs/kit#3248](https://github.com/sveltejs/kit/issues/3248)). So any deployment served over plain HTTP — or behind a reverse proxy/port-mapping where the app-visible request doesn't match the externally-visible origin — produces a callback URI the OIDC provider was never configured to redirect back to, and login fails silently. It also breaks favicons/logos for the same reason, though that half is a SvelteKit deployment concern (`ORIGIN`), not something this codebase controls.

`config.PUBLIC_ORIGIN` exists for exactly this and is already used correctly one function over, in `getOIDCClient`'s CIMD `client_id` URL (`config.PUBLIC_ORIGIN || url.origin`) — `triggerOauthFlow` just never picked up the same fallback.

## Fix

Pulled the `config.PUBLIC_ORIGIN || url.origin` fallback into a small, directly-testable `resolveExternalOrigin(url)` helper (`src/lib/server/resolveOrigin.ts`), and used it at both the existing call site and the previously-unguarded one in `triggerOauthFlow`. Also added an inline comment on `PUBLIC_ORIGIN` in `.env` so its purpose is documented where users set it.

No behavior change when `PUBLIC_ORIGIN` is unset (the common case, behind an HTTPS-terminating proxy) — it still falls back to `url.origin` exactly as before.

## Testing

- Added `src/lib/server/resolveOrigin.spec.ts`: 3 unit tests covering the fallback-to-`url.origin` case and two `PUBLIC_ORIGIN`-set cases (including the exact HTTP-Docker scenario from #2489).
- Full suite: `npx vitest run --project=server --project=ssr` → 1073/1073 passed (no regressions).
- `npx prettier --check .` and `npx eslint .` on changed files: clean.
- `npx svelte-check`: 0 errors, 0 warnings.


revanth-045:
Good catch — fixed in 92c71a2: `resolveExternalOrigin` now routes the configured value through `new URL(...).origin` (with a trailing-slash-trim fallback if it isn't a parseable absolute URL), so a trailing slash or accidental path/query on `PUBLIC_ORIGIN` can't produce a double-slash callback URI. Added 3 more unit tests for it (6 total now). Full suite (1076 tests), prettier, eslint, and svelte-check all pass locally.

PR #2543, opened 2026-08-31
Surface and embed Trackio dashboards, etc. in the side panel in ML Intern mode

abidlabs:
In ML Intern mode a run's output lives somewhere else — a Trackio dashboard on a Space, a demo Space, a dataset. This PR reuses the artifact pane's real estate to show that live, next to the conversation, starting with Trackio dashboards.

Supports multiple Trackio dashboards, multiple artifacts, or a combination of both by letting users scroll using the <> arrows at the top as shown here:

https://github.com/user-attachments/assets/6bbd2e2e-564a-43e5-9e47-4b4326e5f631

A couple of other things that this PR does:

* The ML Intern frequently got the syntax for launching jobs or reading logs wrong, got the Trackio syntax wrong, or forgot to pass in the user's hf token, which would result in a wasted run and wasted time/cost. I added some information about how to launch a Jobs and when/how to use Trackio in the System Prompt
* Addressed feedback from @pngwn/codex

----------------- 

## What it does

When a training run syncs to Trackio, the dashboard Space URL is printed into the job's output. That output already reaches us as a tool result on the message, so the pane picks the URL up from there and frames the live dashboard beside the chat — auto-opening the first time a dashboard appears, with a chip on the tool-call row to get back to it after closing.

## Where the URL comes from

**Tool output only, never the model's prose.** A URL we frame inside chat-ui has to be one a tool actually returned; the model is free to mention a dashboard in its reply and that mention will not open a pane. `collectTrackioDashboards` scans successful results from the tools that run user code (`hf_jobs`, `hf_sandbox*`) and is a pure function of the messages, like `collectArtifacts` — nothing new is persisted, so reopening a conversation rebuilds the same list.

Matching is deliberately narrow: only `*.hf.space` is embeddable, a URL is accepted only when its own log line names Trackio or the Space itself is named for it, and every candidate is re-parsed through `URL` so a userinfo lookalike (`https://x.hf.space@evil.com`) cannot reach the iframe. Both printed forms are handled — the direct `*.hf.space` origin and the Hub page (`huggingface.co/spaces/<owner>/<name>`, converted to the embed subdomain) — normalized so the same dashboard found either way is one entry.

The tool match is by family (`/^hf_(jobs|sandbox)(_|$)/`) rather than exact name. Observed live: a run's dashboard URL arrived in an `hf_sandbox_fs` read of the training log, because `hf_jobs` was unresponsive and the whole run had moved to the sandbox. An exact-name list silently dropped it.

## One pane, one axis

The pane is now a shared shell (`SidePane.svelte`) with the resizable desktop column, mobile fullscreen overlay, drag-to-resize and Escape handling extracted out of `ArtifactPanel`, plus per-view bodies. `artifactPanel` store → `sidePane`, with a `view` discriminant. `ArtifactPanel` lost ~150 lines of geometry and gained nothing else.

Artifacts and dashboards share one slot, so they share one next/previous axis: `collectPaneItems` builds a single ordered list — interleaved by the message each appeared in, dashboards before artifacts within a message since tool output precedes prose — and `PaneItemNav` renders `2/4` with chevrons in every view's header. An artifact contributes one item however many versions it has; versions stay on their own axis in the artifact footer, so a five-version artifact can't bury a training run. The nav hides itself when there is only one item, leaving single-artifact conversations exactly as they were.

## Embedding

Trackio needs a real cross-origin frame, not a `srcdoc`, so the sandbox differs from artifact previews by necessity: a Space needs `allow-same-origin` to reach its own backend, which `PREVIEW_SANDBOX` deliberately withholds. Granting it is safe precisely because the framed origin is never chat-ui's — enforced by the `*.hf.space` allowlist at extraction. The sandbox still denies top-level navigation, so an embedded Space cannot navigate the tab away.

The frame follows chat-ui's resolved light/dark via `subscribeToTheme` (`__theme`), and hides the dashboard's own sidebar (`sidebar=hidden`) since the pane is narrower than a tab and already names the run in its header.

## Not in this PR

- No polling and no job-status chip: the dashboard is live on its own, and the status comes from the last tool result. No new endpoints.
- Embedded Gradio Spaces and datasets, which the item list is now shaped to carry.
- Auto-open still reopens the pane for a *new* dashboard after you have closed one, and a streaming artifact can displace a dashboard you are watching. Both are one-line behaviour decisions, left until the ordering has been used in anger.

## Notes for review

Two pre-existing bugs surfaced while testing this and are **not** fixed here:

1. Static `Authorization` headers configured on a base server in `MCP_SERVERS` never reach the MCP call. `/api/mcp/servers` strips headers before the list reaches the browser (correctly), but the client sends every enabled server back and `runMcpFlow` lets that headerless copy overwrite the env-configured entry ("request takes precedence"). The client should not be able to unset credentials it was never shown.
2. `withMlAssistantServers` seeds its map with the preset first, so a same-named user entry is skipped — an ML Intern conversation silently downgrades a correctly authed `Hugging Face` server to the anonymous `?login` one. Prod is unaffected (it configures `?login` with no static header); self-hosted deployments with a static PAT are not.


pngwn:
Thanks for this!

This looks awesome, excited to try it out!

src/lib/components/chat/ChatWindow.svelte
```diff
@@ -168,7 +170,21 @@
...
+	let trackioDashboards = $derived(collectTrackioDashboards(messages));
+
+	// Open the newest dashboard the first time it shows up: the point of the run
+	// is watching it train. Once per URL, so the user can close it and read the
+	// chat while the run continues.
+	$effect(() => {
+		const latest = trackioDashboards.at(-1);
+		if (!latest) return;
+		sidePane.maybeAutoOpenTrackio(latest.url, latest.label);
 	});
```
pngwn: Two issues with this effect, one fix location.

**1. It's ungated.** Both artifact auto-open paths are carefully gated: the streaming one requires `loading`, and the shared-conversation one requires desktop width ("on mobile the panel is a fullscreen overlay that would hide the conversation entirely"). This one has no `loading`, desktop, or shared guard — and `reset()` clears `autoOpenedKeys` on every conversation switch. Net effect:

- every hard load of a conversation with any past training run re-opens the pane, forever — including runs that finished long ago, at which point the Space may be asleep and the pane frames HF's sleep page;
- on mobile it opens as a fullscreen overlay covering the whole chat on arrival;
- shared conversations copy `messages` verbatim (tool results included), so every viewer gets the auto-open on every device.

**2. It's cancelled on client-side navigation.** This effect is declared *before* the conversation-switch effect that calls `sidePane.reset()` (~L437). Navigating into a conversation with dashboards runs both in the same flush in declaration order: this opens the pane, then `reset()` closes it and clears the key set. The shared-artifact effect below documents this exact hazard and deliberately sits after the reset effect ("Declared after the conversation-switch effect so its reset() can never close the panel after this opens it within the same flush").

Combined: the auto-open fires when unwanted (hard-load revisits of historic conversations) and silently dies when wanted (navigating into a live run).

Suggested fix: move this effect below the conversation-switch effect, and gate it — at minimum on `loading` so it only fires during a live run (which also fixes the historic re-open), and ideally mirroring the shared-artifact desktop guard for mobile.

src/lib/utils/trackio.ts
```diff
@@ -0,0 +1,166 @@
...
+ * one a tool actually returned. The model is free to mention a dashboard in its
+ * reply; that mention will not open a pane.
+ *
+ * Only `*.hf.space` is embeddable (see TrackioPane): Trackio on Spaces is the
+ * default deployment, the origin is never chat-ui's own, and it keeps the
+ * allowlist to a single host suffix we can reason about.
+ */
+
+/** Tools whose output may carry a dashboard URL: the ones that run user code. */
+const TRACKIO_SOURCE_TOOLS = new Set(["hf_jobs", "hf_sandbox"]);
```
pngwn: On the HF MCP server, `hf_sandbox` only creates and manages sandboxes (`create/status/terminate/ps/kill`) — command output, which is where trackio prints its URL, flows through the separate `hf_sandbox_exec` tool. As written this entry can match essentially nothing, and a training script run in a sandbox never surfaces a dashboard.

(Verified against the live HF MCP server's tool schemas; worth re-confirming against the exact server this deployment configures in `MCP_SERVERS`.)

Suggest adding `hf_sandbox_exec`, or matching on the `hf_sandbox` prefix.

src/lib/components/chat/TrackioPane.svelte
```diff
@@ -0,0 +1,94 @@
...
+	 * Unlike an artifact preview this is a real cross-origin document, not
+	 * generated content in a `srcdoc`, so the sandbox is different by necessity:
+	 * a Trackio Space needs `allow-same-origin` to reach its own storage and
+	 * backend, which `PREVIEW_SANDBOX` deliberately withholds from artifacts.
+	 * Granting it here is safe precisely because the framed origin is never
+	 * chat-ui's — `*.hf.space` only, enforced when the URL is extracted (see
+	 * `$lib/utils/trackio`). What the sandbox still buys us is the default deny
+	 * on top-level navigation: an embedded Space cannot navigate the tab away.
+	 */
+	const SANDBOX = "allow-scripts allow-same-origin allow-forms allow-downloads";
```
pngwn: The `allow-same-origin` reasoning in the comment above is sound — the framed origin is never chat-ui's. But the resulting trust rule deserves an explicit sign-off: any `*.hf.space` URL that shares a log line with the string "trackio", printed by a code-running tool, gets auto-framed with these grants. A prompt-injected run (poisoned dataset README, compromised dependency) that prints `trackio: https://attacker-space.hf.space` gets an attacker-controlled Space opened inside trusted chrome labeled "Training dashboard" — a phishing surface for HF tokens.

The blast radius is well-bounded (hf.space origins only, no top-navigation, no popups, no-referrer), so this may be acceptable — but two asks:

1. Does the dashboard actually need `allow-downloads`? If Trackio's export flow doesn't require it, drop it.
2. Pin this string in a test the way `previewSrcdoc.svelte.test.ts` pins `PREVIEW_SANDBOX`, so the grants can't widen silently in a future edit.

pngwn (commented):
**Adversarial review: architecture, security, UX, tests.**

Overall the baseline decisions here are right: extracting the shared `SidePane` frame, deriving dashboards from messages the way `collectArtifacts` does (pure function, nothing persisted, replay-safe), and making Trackio a pane *view* rather than an artifact kind are all the correct calls. The URL validation also held up against spoofing: the re-parse + userinfo rejection + hostname re-check defeats `https://x.hf.space@evil.com` and `https://x.hf.space.evil.com` lookalikes, URLs can only come from tool output (never model prose), and I verified `spaceIdToEmbedOrigin`'s folding against HF's authoritative `subdomain` API field on real Spaces (`yhavinga_t5-v1.1-…` → `yhavinga-t5-v1-1-…`).

The blocking issues are in the line comments:

- the auto-open effect is ungated (re-opens on every hard load of the conversation forever, fullscreen on mobile, and for every viewer of a shared conversation) and is cancelled by `sidePane.reset()` on client-side navigation due to effect ordering;
- `TRACKIO_SOURCE_TOOLS` omits `hf_sandbox_exec`, so sandbox-run training can never surface a dashboard;
- the iframe sandbox grants deserve an explicit decision (`allow-downloads` in particular) plus a pin-test.

**Tests: none of the new code is covered.** `npm run check` and all 1,070 existing tests pass on this branch, but no test touches `trackio.ts`, the renamed store, or the panes. `trackio.ts` is 166 lines of pure functions and is also the security boundary — exactly the shape this codebase already unit-tests (`artifacts.spec.ts` sits next door, and `previewSrcdoc.svelte.test.ts` pins `PREVIEW_SANDBOX` so grants can't widen silently). Before merge this needs:

- `trackio.spec.ts`: both URL forms (`*.hf.space` and `huggingface.co/spaces/...`), line scoping, lookalike/userinfo rejection, trailing-punctuation trimming, root-path normalization + dedupe, and `spaceIdToEmbedOrigin` folding;
- a pin-test for `TrackioPane`'s sandbox string, like the existing `PREVIEW_SANDBOX` assertions;
- store tests for the `maybeAutoOpenTrackio` open-once semantics and `reset()`.

One more note on the seam itself, as a follow-up rather than a blocker: the dashboard URL travels as prose in job logs and is recovered by regex, so a Trackio print-format change breaks this silently. Since HF controls both ends of the pipe, the durable contract would be structured metadata on the tool result (the outputs already carry metadata entries that `ToolUpdate` parses), with the regex kept as a fallback.

src/lib/utils/trackio.ts
```diff
@@ -0,0 +1,166 @@
...
+/**
+ * Whether a URL found on this line is a Trackio dashboard rather than some
+ * other Space the job happened to print. Two deterministic signals, either of
+ * which is enough: the line names Trackio (which is how Trackio's own init
+ * output reads), or the Space itself is named for it (`space_id="user/trackio"`
+ * being the documented default).
+ */
+function isTrackioContext(line: string, spaceSlug: string): boolean {
+	return /trackio/i.test(line) || /trackio/i.test(spaceSlug);
+}
```
pngwn: Nit: the two signals collapse into one — `spaceSlug` is always a substring of `line` (both regexes matched it out of that same line), so the second test is subsumed by the first and this reduces to `/trackio/i.test(line)`. Either drop the param, or if the intent was for the Space's own name to be a stricter independent signal, the doc comment above oversells what's implemented.

pngwn (approved):
Looks great! And thanks for adding the prompt modifications to make trackio the default.

Thanks for this @abidlabs!

PR #2545, opened 2026-09-01
Add Hugging Face Hub mention autocomplete

abidlabs:
https://github.com/user-attachments/assets/3efcac88-48c9-4dac-a21b-00bbc4f3a6f4

## Summary

- add an @-triggered autocomplete for Hugging Face models, datasets, and Spaces
- group Hub quicksearch results by resource type with mouse and keyboard selection
- debounce searches, cancel stale requests, and replace only the mention at the caret
- add accessible loading, empty, error, and active-option states

## Reference implementation

I used [hysts/gradio-huggingface-hub-search](https://github.com/hysts/gradio-huggingface-hub-search) as a reference for:

- the public https://huggingface.co/api/quicksearch endpoint and its q/limit parameters
- the grouped models, datasets, and spaces response shape, including the shared id field and optional Space emoji
- the general live-search pattern of debouncing input, cancelling stale requests, and grouping results by resource type

No Gradio component code or assets are reused. The textarea mention detection and replacement, Svelte state management, dropdown UI, keyboard interactions, accessibility behavior, and tests are implemented specifically for Chat UI.

## Testing

- npm run check
- npm run build
- npx vitest run src/lib/utils/hfHubSearch.test.ts src/lib/components/chat/HfHubMentionAutocomplete.svelte.test.ts --project=server --project=client
- ESLint and Prettier checks on changed files


pngwn:
This looks great! I'll review properly later but i think it could be good if the model was a little tighter to the text and a bit more compact in general.

I think the modal should appear where the caret is, so it feels like an autocomplete for that text, and i think we can drop the fon't sizes/ spacing a touch so it feels abit neater with the rest of the UI.

I think the way codex/ claude do it is tidy. We should keep the repo type subheadings tho.

Screenshot 2026-09-01 at 04 07 15

pngwn (commented):
Took a careful pass over this. The feature is a nice idea and the factoring is mostly right — pulling `findHfHubMention` / `replaceHfHubMention` / `parseHfHubQuicksearch` out as pure functions with real unit tests is exactly the seam I'd want, and the caret/tail handling in `findHfHubMention` is thoughtful.

That said, I don't think this is mergeable as-is. 13 comments inline; two I'd treat as blockers:

1. **Enter is hijacked for any `@word`** (`ChatInput.svelte:300`). The first option is always pre-highlighted and there's no "nothing selected" state, so typing `can you ping @john` and hitting Enter rewrites the text to an unrelated Hub repo id *and* swallows the send. The same branch is also missing the `!isVirtualKeyboard()` guard the submit path deliberately has, so Enter stops inserting newlines on mobile.

2. **The Hub call isn't gated by `publicConfig.isHuggingChat`** (`hfHubSearch.ts:15`). `CLAUDE.md` rule 1 requires it and `+layout.svelte:153` is the existing precedent. As written, every self-hosted deployment ships composer keystrokes to `huggingface.co` on each debounce tick, with no flag and no opt-out.

Two more correctness issues worth fixing before merge:

3. The autocomplete state survives submit — `draft = ""` in `ChatWindow` fires no `input` event, so the panel reappears over the emptied composer and selecting a result writes through stale offsets.
4. Accepting a suggestion immediately re-opens the panel and refetches, because nothing marks a mention as accepted or dismissed. Escape has the same problem — one click undoes it.

The rest are a11y gaps (the listbox owns non-option children; the textarea is missing `role="combobox"`/`aria-expanded`), an altitude note about extracting the search state machine into a `.svelte.ts` class so the racy parts are testable, and three minor cleanups.

Happy to talk through any of these if you'd rather push back on some.

src/lib/components/chat/ChatInput.svelte
```diff
@@ -183,7 +208,107 @@
...
+				activeHubResultIndex = (activeHubResultIndex + 1) % hubSearchResults.length;
+				return;
+			}
+			if (event.key === "ArrowUp" && hubSearchResults.length > 0) {
+				event.preventDefault();
+				activeHubResultIndex =
+					(activeHubResultIndex - 1 + hubSearchResults.length) % hubSearchResults.length;
+				return;
+			}
+			if ((event.key === "Enter" || event.key === "Tab") && hubSearchResults.length > 0) {
```
pngwn: **Enter/Tab is hijacked for any `@word`, rewriting ordinary at-mentions and swallowing the send.**

There is no "nothing selected" state — `activeHubResultIndex` starts at `0` and `HfHubMentionAutocomplete.svelte` always highlights `resultIndex === activeIndex`, so the first result is pre-selected the moment results land.

Repro: type `can you ping @john` and press Enter to send. Quicksearch returns something for `john`, so this branch fires, `preventDefault()`s, runs `selectHubResult`, and `return`s before the submit branch below. The message is **not sent** and the text silently becomes `can you ping @johnsnowlabs/… `.

This will fire for every non-Hub `@word` that happens to match anything on the Hub. Suggest starting with `activeHubResultIndex = -1` and only treating Enter as "accept" once the user has explicitly arrowed into the list (Tab can still accept the first result).

src/lib/components/chat/ChatInput.svelte
```diff
@@ -183,7 +208,107 @@
...
+		resetHubSearch();
+
+		await tick();
+		textareaElement?.focus();
+		textareaElement?.setSelectionRange(replacement.caret, replacement.caret);
+		adjustTextareaHeight();
+	}
+
 	function handleKeydown(event: KeyboardEvent) {
+		if (isHubMentionOpen && !isCompositionOn) {
```
pngwn, line 261: **This branch is missing the `!isVirtualKeyboard()` guard that the submit path below has.**

The submit branch at line 312 checks `!isVirtualKeyboard()` precisely so that Enter stays the newline key on touch devices. The mention branch has no such guard, so on a phone a user writing a multi-line prompt containing `@bert` gets a suggestion inserted instead of a line break — and Escape isn't reachable on a soft keyboard, so there's no way out until the mention resolves.

src/lib/utils/hfHubSearch.ts
```diff
@@ -0,0 +1,99 @@
...
+	emoji?: string;
+}
+
+export interface HfHubMention {
+	query: string;
+	start: number;
+	end: number;
+}
+
+const HUB_QUICKSEARCH_URL = "https://huggingface.co/api/quicksearch";
```
pngwn, line 15: **This needs to be gated behind `publicConfig.isHuggingChat`.**

From `CLAUDE.md` at the repo root:

> 1. **HuggingChat vs self-hosted**: Wrap HuggingChat-specific features with `publicConfig.isHuggingChat`

The existing precedent does exactly that — `src/routes/+layout.svelte:153`:

```js
if (publicConfig.isHuggingChat && data.user?.username) {
    fetch(`https://huggingface.co/api/users/${data.user.username}/overview`)
```

As written, every self-hosted deployment makes a cross-origin request to `huggingface.co` carrying a prefix of whatever the user typed after any `@`, on every debounce tick, with no config flag and no opt-out. That's a real problem for air-gapped/enterprise installs, and on deployments that block `huggingface.co` or set a restrictive `connect-src` CSP the feature will just render the error panel on every `@`.

src/lib/components/chat/ChatInput.svelte
```diff
@@ -74,6 +82,17 @@
...
 	let blurTimeout: ReturnType<typeof setTimeout> | null = $state(null);
+	let hubMention = $state<HfHubMention | null>(null);
+	let hubSearchResults: HfHubResource[] = $state([]);
+	let hubSearchStatus: "idle" | "loading" | "success" | "error" = $state("idle");
+	let activeHubResultIndex = $state(0);
+	let hubSearchTimeout: ReturnType<typeof setTimeout> | null = null;
+	let hubBlurTimeout: ReturnType<typeof setTimeout> | null = null;
+	let hubSearchAbortController: AbortController | null = null;
+	let hubSearchSequence = 0;
+	const isHubMentionOpen = $derived(
```
pngwn: **The autocomplete state survives message submit, so the panel reappears over the freshly-cleared composer.**

`isHubMentionOpen` only depends on `hubMention` and `hubSearchStatus` — nothing ties it back to the current `value`.

Repro: type `Compare @bert` and press Enter within the 250 ms debounce (or while the fetch is in flight). `hubSearchResults` is still empty, so the Enter branch is skipped and the submit branch runs. `ChatWindow.svelte:194` then does `draft = ""` — a programmatic assignment, which fires no `input` event, so nothing calls `resetHubSearch()`. `hubMention = {query:"bert", start:8, end:13}` and `hubSearchStatus === "loading"` both persist, this derived stays `true`, and the panel keeps hovering over the now-empty composer. When the debounced fetch resolves it flips to `success` and renders results for a message that was already sent.

Clicking one then calls `replaceHfHubMention("", {start:8, end:13}, id)` — both slices yield `""` — producing `@org/model ` in the brand-new draft, with a caret index past the end of the string.

A `$effect` that resets when `value` no longer contains the tracked mention would cover this and the externally-set-`value` case generally.

src/lib/components/chat/ChatInput.svelte
```diff
@@ -183,7 +208,107 @@
...
+
+	async function selectHubResult(result: HfHubResource) {
+		if (!hubMention) return;
+		const replacement = replaceHfHubMention(value, hubMention, result.id);
+		value = replacement.value;
+		resetHubSearch();
+
+		await tick();
+		textareaElement?.focus();
+		textareaElement?.setSelectionRange(replacement.caret, replacement.caret);
```
pngwn, line 256: **Accepting a suggestion immediately re-opens the panel and fires a fresh search for the id that was just inserted.**

Nothing marks a mention as already accepted (or dismissed), so `findHfHubMention` happily matches a completed `@org/model` token.

Walk through accepting a result in `Compare @distilb with BERT`: value becomes `Compare @distilbert/distilbert-base-uncased with BERT` and this line calls `setSelectionRange(43, 43)`. Assigning `.value` had put the caret at the end (52), so 43 is a change, and the HTML "set the selection range" algorithm fires a `select` event → `onselect={syncHubMentionFromTextarea}` (line 410) → `findHfHubMention` sees `beforeCaret` ending in the full id and returns `query: "distilbert/distilbert-base-uncased"`. `queryChanged` is `true` because `resetHubSearch()` just nulled `hubMention`, so the panel re-opens with a spinner and a new request.

Even setting the `select` event aside, the same thing happens on the two certain paths: clicking inside a completed mention (`onclick`, line 409) or arrowing into it (`onkeyup`, line 404).

It also defeats Escape — `resetHubSearch()` at line 307 clears state but leaves no "dismissed" flag, so a single click re-opens the panel the user just dismissed.

src/lib/components/chat/ChatInput.svelte
```diff
@@ -183,7 +208,107 @@
...
+		const queryChanged = nextMention.query !== hubMention?.query;
+		hubMention = nextMention;
+		if (!queryChanged) return;
+
+		const sequence = ++hubSearchSequence;
+		if (hubSearchTimeout) clearTimeout(hubSearchTimeout);
+		hubSearchAbortController?.abort();
+		hubSearchAbortController = null;
+		hubSearchResults = [];
+		hubSearchStatus = "loading";
```
pngwn: **Setting `loading` synchronously means the panel flashes over the composer during ordinary typing.**

`isHubMentionOpen` flips `true` here, before the 250 ms debounce has even started, so typing `hey @a` immediately pops the "Searching the Hugging Face Hub…" overlay across the composer (covering the ML Assistant strip). It re-renders on every subsequent keystroke, so writing `@alice` produces five spinner flashes and five quicksearch requests for `a`/`al`/`ali`/`alic`/`alice`.

Worth adding a minimum query length (2–3 chars) and only surfacing the loading state once the request is actually in flight, rather than during the debounce window.

src/lib/components/chat/ChatInput.svelte
```diff
@@ -205,10 +330,21 @@
 			clearTimeout(blurTimeout);
 			blurTimeout = null;
 		}
+		if (hubBlurTimeout) {
+			clearTimeout(hubBlurTimeout);
+			hubBlurTimeout = null;
+		}
 		focused = true;
+		syncHubMentionFromTextarea();
```
pngwn: **Focusing the composer now auto-opens the panel and fires a request with no user action.**

If a draft ends in an `@token` — restored draft, or the user just clicking back in with the caret parked there — `onMount` → `focusTextarea()` → `handleFocus` → this call finds the mention and opens the autocomplete with a spinner and a `huggingface.co` request on page load, before anything is typed. Enter to send is then hijacked per the comment on line 300.

src/lib/components/chat/HfHubMentionAutocomplete.svelte
```diff
@@ -0,0 +1,100 @@
...
+		listboxElement
+			?.querySelector<HTMLElement>(`[data-result-index="${activeIndex}"]`)
+			?.scrollIntoView({ block: "nearest" });
+	});
+</script>
+
+<div
+	bind:this={listboxElement}
+	id="hf-hub-mention-listbox"
+	role="listbox"
```
pngwn: **`role="listbox"` here owns children that are neither options nor groups.**

ARIA requires a listbox to own only `option` elements (or `group`s containing them). This container directly holds the `<div data-resource-header>` category headers (line 58) and, in the loading/error/empty branches, `<div role="status">` (lines 43/47/51). Screen readers computing the option set can mis-report positions ("1 of 6", counting headers) or drop options entirely.

Separately, those `role="status"` live regions are inserted into the DOM at the same instant the listbox mounts — live regions that don't pre-exist the content change are frequently not announced at all, so the loading and error states are silent.

Suggest wrapping each category in `role="group"` with an `aria-label`, and moving the status region outside the listbox so it's present before the state changes.

src/lib/components/chat/ChatInput.svelte
```diff
@@ -238,24 +374,52 @@
...
+				onactivechange={(index) => (activeHubResultIndex = index)}
+			/>
+		{/if}
+
+		<textarea
+			rows="1"
+			tabindex="0"
+			inputmode="text"
+			autocomplete="off"
+			aria-autocomplete="list"
```
pngwn, line 378: **The combobox ARIA is incomplete — missing `role="combobox"` and `aria-expanded`.**

Per ARIA 1.2, `aria-activedescendant` and `aria-controls` are only interpreted in the combobox pattern, which requires the input to expose `role="combobox"` with `aria-expanded` reflecting popup state.

As written, a screen-reader user typing `@bert` gets a plain multi-line text field: no announcement that suggestions appeared, no announcement of the active option while arrowing through the list, and no indication that Enter will now insert a completion instead of sending the message.

src/lib/components/chat/ChatInput.svelte
```diff
@@ -183,7 +208,107 @@
 		adjustTextareaHeight();
 	});
 
+	function resetHubSearch() {
```
pngwn: **Altitude: this whole async state machine probably belongs in a `.svelte.ts` rune class rather than inlined here.**

`ChatInput` now carries three independent timeout handles (`blurTimeout`, `hubSearchTimeout`, `hubBlurTimeout`), an `AbortController`, a monotonic `hubSearchSequence`, four pieces of `$state` and their `onDestroy` teardown — interleaved with the pre-existing focus/blur/virtual-keyboard logic.

The repo already has the seam for this: `src/lib/stores/mlAssistant.svelte.ts` and `src/lib/utils/scroll/chatScroll.svelte.ts` are rune classes with `.spec.ts` coverage.

The practical cost is test coverage. Because the logic lives in markup-coupled component scope there's no `ChatInput` test at all — this PR tests only the pure helpers and the presentational child, so the debounce/abort/sequence races, the Enter-vs-submit precedence, and the reset-on-submit gap are all untested. A `HubMentionState` class with `update(value, caret)` / `accept(result)` / `dismiss()` / `destroy()` would make each of those directly testable.

src/lib/components/chat/ChatInput.svelte
```diff
@@ -238,24 +374,52 @@
...
-		{disabled}
-		onfocus={handleFocus}
-		onblur={handleBlur}
-		onbeforeinput={requireAuthUser}
-	></textarea>
+	<div class="relative">
+		{#if isHubMentionOpen}
+			<HfHubMentionAutocomplete
+				results={hubSearchResults}
+				status={hubSearchStatus === "idle" ? "loading" : hubSearchStatus}
```
pngwn: Minor: this ternary is unreachable. It sits inside `{#if isHubMentionOpen}`, and that derived (line 93) already requires `hubSearchStatus !== "idle"` — the `"idle"` branch can never be taken. It only exists to satisfy the child's narrower prop type.

Same for `hubMention.query.length > 0` in the derived itself: `updateHubMention` calls `resetHubSearch()` whenever `!nextMention?.query`, so a non-null `hubMention` always has a non-empty query.

Narrowing `hubSearchStatus` to `"loading" | "success" | "error" | null` (null = closed) drops both and removes a fourth status the component never renders.

src/lib/utils/hfHubSearch.ts
```diff
@@ -0,0 +1,99 @@
...
+const REPO_ID_CHARS = "A-Za-z0-9._/-";
+
+const isRecord = (value: unknown): value is Record<string, unknown> =>
+	typeof value === "object" && value !== null;
+
+/** Find the Hub mention being edited at the current textarea caret. */
+export function findHfHubMention(value: string, caret: number): HfHubMention | null {
+	const safeCaret = Math.max(0, Math.min(caret, value.length));
+	const beforeCaret = value.slice(0, safeCaret);
+	const match = new RegExp(`(?:^|[\\s([{])@([${REPO_ID_CHARS}]*)$`).exec(beforeCaret);
```
pngwn: Minor: both regexes are built from the `REPO_ID_CHARS` module constant, so they never vary — but `new RegExp` recompiles them on every call (here and line 32). `findHfHubMention` runs from `handleInput` plus four other textarea handlers, so fast typing recompiles them several times per character.

Hoisting them to module-level literals is free — neither uses the `g` flag, so there's no `lastIndex` state to share.

src/lib/components/chat/HfHubMentionAutocomplete.svelte
```diff
@@ -0,0 +1,100 @@
...
+					data-resource-header={type}
+					class={[
+						"border-t border-gray-200 px-4 py-2 text-sm font-semibold first:border-t-0 dark:border-gray-800",
+						headerClasses[type],
+					]}
+				>
+					{labels[type]}
+				</div>
+				{#each group as result (result.id)}
+					{@const resultIndex = results.indexOf(result)}
```
pngwn: Minor: the template walks `results` three times with `.filter` (line 56) and then does a linear `.indexOf` per option here to recover the flat index — O(n²) plus three extra allocations, re-run on every render (each arrow key, each `mouseenter` that changes `activeIndex`).

The data already arrives flat and in category order from `parseHfHubQuicksearch`. It's also a bit fragile: the index is recovered by object identity rather than carried with the item, so it breaks silently if `results` is ever mapped or copied. A single `$derived` producing `[{ type, options: [{ result, index }] }]` computes it once and removes the identity dependency.

abidlabs:
I think everything should be addressed @pngwn! Here's what it looks like now:

https://github.com/user-attachments/assets/e509b346-728e-476e-9608-4a6c0400cc50

pngwn (approved):
Amazing! Thanks for addressing those comments @abidlabs. I noticed a small bug and pushed a fix. 

Will merge when CI is green.

PR #2549, opened 2026-09-01
Let the user check a parked wait early

pngwn:
check-now-pill

When the ML Intern parks a turn on the `wait` tool, the countdown is not always worth sitting out — the user can see what it is waiting for and often knows the job is worth a look now. Today the only answer is to type a message asking it to check.

The wait pill grows a play control at its end: a connected segment sharing the pill's border, icon only, no label.

### How the wake works

- `POST /conversation/[id]/wake` with the parked message id moves the parked row's `resumeAt` to now, and nothing else. The row stays `waiting`, so the ordinary claim still decides which pod resumes it — an early wake racing a sweeper cannot produce two producers for one turn. The endpoint then kicks a sweep, purely so the user doesn't sit through the sweep interval (10s by default).
- `wokeEarlyAt` on the row is what the model reads on the round it wakes into: *"The user asked you to check on this early, so the wait was cut short."* Without it, a user-shortened wait looks like the model's own choice and it mis-sizes the next one.
- The banner switches to "checking now…" on acceptance and stays there — the resumed turn's own output replaces it when `running` arrives. A failed request puts the countdown back, since the timer is untouched and the turn still wakes on its own.
- No control on a read-only/shared view, or once the deadline is far enough past that a sweep already owns the turn.

### Testing

`wakeParkedCallEarly` (deadline moved, row still claimable, only the turn asked about), the resume tool result's early-check wording, and the pill itself (posts the wake, stops counting down, restores the countdown on failure, hidden when read-only or overdue). Full suite: 1076 passing.

### DIFF
--- a/src/lib/components/chat/ChatMessage.svelte
+++ b/src/lib/components/chat/ChatMessage.svelte
@@ -64,7 +64,7 @@
 		message,
 		loading = false,
 		isAuthor: _isAuthor = true,
-		readOnly: _readOnly = false,
+		readOnly = false,
 		isTapped = $bindable(false),
 		alternatives = [],
 		editMsdgId = $bindable(null),
@@ -93,7 +93,6 @@
 	$effect(() => {
 		// referenced to appease linter for currently-unused props
 		void _isAuthor;
-		void _readOnly;
 	});
 	function handleKeyDown(e: KeyboardEvent) {
 		if (e.key === "Enter" && (e.metaKey || e.ctrlKey)) {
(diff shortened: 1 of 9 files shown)


abidlabs:
Works great @pngwn! 

image

One small thing I noticed is that if you click "check on this now" then the llm thinks that you waited the full timer amount and there was no result. So it tends to _increase_ the timer value when it runs `wait` again. Might be good to prevent this from happening.

abidlabs (approved):

PR #2559, opened 2026-09-02
Say that the billing setting covers inference

abidlabs:
Ideally the organization you select in the billing would also be the org that is charged for Jobs, etc. but until we have that, we can add this minimal language to indicate it's only for inference:

- `ad30186d` — Billing setting description now says "for inference": the selection only redirects inference spend (`X-HF-Bill-To` is attached to the OpenAI-compatible client and nothing else reads it), while Jobs bill the namespace they run in.


pngwn (approved):
Looks great. Thanks @abidlabs!

PR #2560, opened 2026-09-02
ML Intern: fixed model set with pinned providers, plus a model probe script

pngwn:
## What

- `ML_ASSISTANT_MODELS`: a JSON5 list of the models ML Intern conversations may run on, each pinning the inference provider it was verified on. First entry is the default; `parameters` merge over the catalog entry for mode conversations.
- Creating a mode conversation swaps any unlisted model (omni included) for the default. Switching a mode conversation to an unlisted model returns 400. The pinned provider overrides the user's per-model preference and `auto` at request time, in the conversation route and the parked-turn sweeper.
- With the ML Intern pill on, the Models page and the settings sidebar list only the set, in order, with the default marked.
- `scripts/probe-model.ts` (`npm run probe-model -- --model <id>`): sends the tool loop's exact request shape with a generated ML Intern history at 100k to 1M tokens, grades recall of verbatim log lines, and gives each model/provider run a verdict (OK, DEGRADED, SWALLOWED, TRUNCATED, ERROR).
- Chart pins for prod and dev, docs, `.env` reference.

## Why

Users saw ML Intern turns "stop mid-flow" on very long contexts. Probing GLM-5.3-Flash on the router showed it is provider-specific: on baseten the model intermittently fabricated, looped ("I I I I…"), or emitted garbage above ~150k prompt tokens and then ended with a normal `finish_reason: "stop"`, which chat-ui finalizes as a completed answer. The same requests on together and novita answered verbatim at 1,034k tokens. Baseten also silently discards a tool call whose name is not in the `tools` array (finish "stop", no delta, tokens still billed). The router's `auto` choice landed on baseten, so pinning the provider per model is the only reliable guard.

Matrix used for the chart pins (recall of verbatim log lines):

| model | pinned | evidence | notes |
|---|---|---|---|
| GLM-5.3-Flash | together | 5/5 at 924k | novita 4/5; fireworks 504 at 924k; baseten inconsistent |
| Kimi-K3 | baseten | 6/6 at 1,046k, 5/5 at 892k | fireworks 6/6 (slow prefill); together reasons to the cap near 1M; deepinfra's real window is 262k; featherless drops streams |
| GLM-5.3 | novita | 3/4 at 827k, 2/5 at 924k | degrades above ~800k on every provider |
| DeepSeek-V4-Pro-0813 | together | 4/4 at 761k, 5/5 at 911k | fireworks equal; novita and baseten 0/4 at 761k; deepinfra 1/5 at 910k; baseten 529 at 911k |
| DeepSeek-V4-Flash-0731 | novita | 3/4 at 761k, 5/5 at 911k | fireworks 4/5 at 911k; together and baseten 0/4 at 761k; deepinfra 0/5 at 910k |

## Tests

- Unit tests for the model set module; route tests for creation substitution and the switch guard.
- `npm run test`: 96 files, 1190 tests pass. `npm run check`: 0 errors.

### DIFF
--- a/chart/env/dev.yaml
+++ b/chart/env/dev.yaml
@@ -73,6 +73,12 @@ envVars:
   LLM_ROUTER_ENABLE_TOOLS: "true"
   LLM_ROUTER_TOOLS_MODEL: "moonshotai/Kimi-K2.6"
   TRANSCRIPTION_MODEL: "openai/whisper-large-v3-turbo"
+  # ML Intern model set: provider pinned per model from scripts/probe-model.ts runs
+  # (2026-09-02). GLM-5.3-Flash on together recalled 5/5 at 924k tokens; Kimi-K3 on
+  # baseten 6/6 at 1,046k (fireworks-ai is the alternate: 6/6, slower prefill);
+  # GLM-5.3 degrades above ~800k on every provider — novita was the best of them.
+  ML_ASSISTANT_MODELS: >
+    [{"id": "zai-org/GLM-5.3-Flash", "provider": "together"}, {"id": "moonshotai/Kimi-K3", "provider": "baseten"}, {"id": "zai-org/GLM-5.3", "provider": "novita"}]
   MCP_SERVERS: >
     [{"name": "Web Search (Exa)", "url": "https://mcp.exa.ai/mcp?tools=web_search_exa,get_code_context_exa,crawling_exa"}, {"name": "Hugging Face", "url": "https://hf.co/mcp?login"}]
   MCP_TOOL_TIMEOUT_MS: "120000"
--- a/chart/env/dev.yaml
+++ b/chart/env/dev.yaml
@@ -77,8 +77,11 @@ envVars:
   # (2026-09-02). GLM-5.3-Flash on together recalled 5/5 at 924k tokens; Kimi-K3 on
   # baseten 6/6 at 1,046k (fireworks-ai is the alternate: 6/6, slower prefill);
   # GLM-5.3 degrades above ~800k on every provider — novita was the best of them.
+  # DeepSeek-V4-Pro-0813 on together 5/5 at 911k (fireworks-ai equal; novita and
+  # baseten 0/4 at 761k); DeepSeek-V4-Flash-0731 on novita 5/5 at 911k (together
+  # and baseten 0/4 at 761k, deepinfra 0/5 at 910k).
   ML_ASSISTANT_MODELS: >
-    [{"id": "zai-org/GLM-5.3-Flash", "provider": "together"}, {"id": "moonshotai/Kimi-K3", "provider": "baseten"}, {"id": "zai-org/GLM-5.3", "provider": "novita"}]
+    [{"id": "zai-org/GLM-5.3-Flash", "provider": "together"}, {"id": "moonshotai/Kimi-K3", "provider": "baseten"}, {"id": "zai-org/GLM-5.3", "provider": "novita"}, {"id": "deepseek-ai/DeepSeek-V4-Pro-0813", "provider": "together"}, {"id": "deepseek-ai/DeepSeek-V4-Flash-0731", "provider": "novita"}]
   MCP_SERVERS: >
     [{"name": "Web Search (Exa)", "url": "https://mcp.exa.ai/mcp?tools=web_search_exa,get_code_context_exa,crawling_exa"}, {"name": "Hugging Face", "url": "https://hf.co/mcp?login"}]
   MCP_TOOL_TIMEOUT_MS: "120000"
--- a/.env
+++ b/.env
@@ -230,6 +230,14 @@ APP_BASE="" # base path of the app, e.g. /chat, left blank as default
 # set). Enforced server-side: submissions reserve their worst case (flavor
 # price x timeout) and settle to actual runtime.
 ML_ASSISTANT_MODE=
+# Models ML Intern conversations may run on, JSON5 array; first entry is the
+# default. Each entry pins the inference provider the model was verified on for
+# 1M-token tool loops (scripts/probe-model.ts) — the user's provider preference
+# and "auto" are ignored for mode conversations. `parameters` merge over the
+# catalog entry's. The router alias cannot be listed. Empty: the mode refuses
+# to start a conversation.
+# e.g. [{"id":"zai-org/GLM-5.3-Flash","provider":"together"},{"id":"moonshotai/Kimi-K3","provider":"together"}]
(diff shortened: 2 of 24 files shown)


chart/env/dev.yaml
```diff
@@ -73,6 +73,15 @@ envVars:
...
   TRANSCRIPTION_MODEL: "openai/whisper-large-v3-turbo"
+  # ML Intern model set: provider pinned per model from scripts/probe-model.ts runs
+  # (2026-09-02). GLM-5.3-Flash on together recalled 5/5 at 924k tokens; Kimi-K3 on
+  # baseten 6/6 at 1,046k (fireworks-ai is the alternate: 6/6, slower prefill);
+  # GLM-5.3 degrades above ~800k on every provider — novita was the best of them.
+  # DeepSeek-V4-Pro-0813 on together 5/5 at 911k (fireworks-ai equal; novita and
+  # baseten 0/4 at 761k); DeepSeek-V4-Flash-0731 on novita 5/5 at 911k (together
+  # and baseten 0/4 at 761k, deepinfra 0/5 at 910k).
+  ML_ASSISTANT_MODELS: >
+    [{"id": "zai-org/GLM-5.3-Flash", "provider": "together"}, {"id": "moonshotai/Kimi-K3", "provider": "baseten"}, {"id": "zai-org/GLM-5.3", "provider": "novita"}, {"id": "deepseek-ai/DeepSeek-V4-Pro-0813", "provider": "together"}, {"id": "deepseek-ai/DeepSeek-V4-Flash-0731", "provider": "novita"}]
```
abidlabs, line 84: Have we confirmed that Kimi K3 works with Baseten? Or should we choose a different provider?

abidlabs: Ah just read the results you had above, nice

abidlabs:
Nice find on the provider-specific degradation — the probe script earns its place. Two gaps where a mode conversation can still reach an unverified provider, which is the thing this is guarding against:

**1. `/api/v2/conversations/[id]` PATCH bypasses the switch guard.**

The `mlAssistantModelIds()` check only lives in the legacy handler (`src/routes/conversation/[id]/+server.ts:862-869`). The v2 endpoint validates `model` with `validModelIdSchema` alone (`src/routes/api/v2/conversations/[id]/+server.ts:97-101`) and then calls `applyConversationSettings` directly, so an ML Intern conversation can be moved onto any catalog model — `omni` included — straight past the pin.

`conversationSettings.ts` exists precisely because these two endpoints diverging was a past silent bug ("an endpoint that changes `model` with a plain `$set` reintroduces exactly that, which is what happened while only the legacy handler had the pipeline"). This reintroduces the same divergence one layer up, at the validation step. Probably wants to be a shared validator, or hoisted into `applyConversationSettings` alongside the backfill.

**2. The router alias isn't actually excluded by construction.**

`src/lib/server/mlAssistantModels.ts:12-13` says "The router alias is excluded by construction — it can't be listed here", and `.env:237` says "The router alias cannot be listed" — but I don't see anything enforcing it. The alias is an ordinary entry in `models` (`src/lib/server/models.ts:414`), so `parseMlAssistantModels`'s `knownIds` check at line 52 accepts `[{"id": "omni", "provider": "together"}]` happily. The pin is then dropped by the `!model.isRouter` guard at the two call sites, and you get a silently unpinned mode conversation — the exact failure mode the PR is closing.

Looks like a one-line fix: drop `isRouter` entries in `parseMlAssistantModels` (with the same error log as an unknown id), so the invariant the comments assert is real.

---

A few smaller things, non-blocking:

- **`DEFAULT` badge mislabels under search.** `src/routes/models/+page.svelte:154` is now `index === 0 && (model.isRouter || mlModelsOnly)`. `isRouter` was a property of the model itself; `mlModelsOnly` isn't — so with a filter typed that excludes the actual default, whatever lands at index 0 gets the badge. Comparing against `data.mlAssistantModels[0]` would be exact.
- **Empty set fails late.** A build with `ML_ASSISTANT_MODE=true` and no `ML_ASSISTANT_MODELS` 400s "ML Intern has no models configured" at send time, with the pill still fully enabled. Worth hiding the toggle when `mlAssistantModels` is empty rather than letting the user compose into a dead end.
- **Substitution is silent.** Starting a chat from `/models/<some-other-model>` with the mode on posts the set default instead (`src/routes/models/[...model]/+page.svelte:41-47`) with nothing surfaced to the user.
- **`ModelSwitch` isn't filtered.** It renders for read-only conversations (`ChatWindow.svelte:912`) — exactly the since-removed-entry case the `mlAssistantProviderFor` comment anticipates — still lists every model, and its only failure path is a `console.error`. Picking an unlisted model in a mode conversation is a silent no-op.
- **Self-hosted gets no pinning.** Both call sites gate on `config.isHuggingChat`, so the pin is inert off HuggingChat while `.env` documents the key for everyone. Might be worth saying so in the docs.
- **Nit:** `runMcpFlow.ts:470` uses `targetModel.id ?? targetModel.name`; the convention elsewhere is `id || name` (e.g. `conversation/+server.ts:56`), which also handles an empty-string id.

chart/env/dev.yaml
pngwn: Yep, confirmed via the probe in the table above.

pngwn:
Thanks both, all addressed in `5d53368e`.

**Blocking**
- **v2 PATCH bypass**: the guard now lives in `applyConversationSettings`, so both endpoints share it. It looks up the conversation's mode flag and refuses an unlisted model with a 400. The route test covers the legacy and the v2 handler.
- **Router alias**: the parser now receives catalog entries (`{id, isRouter}`) and drops the alias with the same error log as an unknown id. Tested. The comments that said "excluded by construction" now describe something the code does.

**Codex**
- Share import: the mode's model is resolved after the `fromShare` block, which was the last thing able to change it.
- `provider` is required; an entry without one fails validation.
- Probe percentages are computed from the whole window and clamped only so prompt plus reply fits, so `recall@90%` now means 90%.

**Smaller**
- Default badge compares against the set's first id, so search can no longer move it.
- The composer hides the switch when `ML_ASSISTANT_MODELS` is empty.
- ModelSwitch lists only the set for a mode conversation and surfaces the server's error message instead of `console.error`.
- Docs and `.env` say the pin is a router feature and self-hosted deployments get the set without it.
- `id || name` nit fixed.

**Kept as is: silent substitution from `/models/<other-model>`.** That is the intended flow (pick any model, turn the pill on, get the set's default), and with both pickers filtered it only happens via a direct URL. The sidebar and header show the actual model as soon as the conversation exists. Happy to add a toast if you'd still prefer one.

PR #2568, opened 2026-09-09
Recover when a question's answer was recorded but never continued

pngwn:
## The symptom

A user gets stuck on an assistant question. Every Send or Skip shows "Already answered." in red under the panel, and reloading brings the same question back. The console shows the elicitation endpoint answering 409 to each click.

## Why it happens

Answering an assistant question is two steps. The answer endpoint records the answer, then the page starts a second run that continues the parked turn, and only that run writes the Resolved update into the message. If the page loses its cue between the two steps, the row is resolved but the parked message still lacks the Resolved update, so the transcript re-opens the question. A reload, a closed tab, an answer sent while the page was still loading, or a run that died before persisting all produce it.

Once there, the client had no way out: the 409 carried only a message, and both buttons post.

## The fix

- `submitElicitationAnswer` still refuses a repeat answer to a durable prompt with a 409, but the response now says what stands and whether the parked call was ever continued.
- "Continued" is judged by what could have consumed the answer: a producer live on the turn or a parked call being claimed means one is under way, and otherwise the parked message either carries the Resolved update a continuation writes or it does not. The turn-state document is deliberately not the judge, because a continuation that dies leaves it at `running` for good.
- The elicitation route returns that as JSON rather than through `error()`, which has nowhere to put it.
- `sendElicitationAnswer` queues the continuation on that signal exactly as it does after a successful answer.
- The composer question unregisters itself instead of showing the refusal. The transcript row settles when the continuation reports the recorded answer.

A user's earlier answer is what continues; a different pick on the repeat is not applied.

## Review follow-up

- The finalizer no longer closes durable prompts when a run is reaped. Nothing polls them, and their answer continues the turn whenever it comes, so closing one turned the user's real answer into a replayed cancel.
- Rows now record who resolved them (`resolution`). A system close that nothing consumed gives way to the user's answer instead of standing in for it; rows from before the field existed read a durable prompt's `cancel` as the system's.
- A failed lookup returns a retryable error rather than passing for "not resumable", so the question stays open instead of being dropped with nothing queued.

## Verification

- MCP and generation server specs: 214 pass across 18 files, including eight new repeat-answer cases (never continued, died after claiming the turn, already ran, under way, system close replaced, legacy cancel, consumed system close kept, lookup failure) and one for the finalizer.
- AskQuestion and ElicitationForm browser specs: 31 pass, including three new composer cases.
- svelte-check: 0 errors. eslint and prettier clean on the changed files.

## Unblocking a conversation already in this state

Nothing manual is needed once this is deployed: the next Send on the re-opened question recovers it.

## Left out

Two tabs answering within the same instant could both start the continuation. That window is much smaller than the dead end it replaces; closing it needs a compare-and-swap on the turn claim in the conversation route, which is a separate change.

### DIFF
--- a/src/lib/components/chat/AskQuestion.svelte
+++ b/src/lib/components/chat/AskQuestion.svelte
@@ -104,7 +104,10 @@
 		});
 		submitting = false;
 		if (!result.ok) {
-			error = result.error;
+			// The answer that stands is the earlier one, so there is nothing left to ask; the
+			// transcript row settles when the continuation reports it.
+			if (result.answered) unregisterQuestion(request.elicitationId);
+			else error = result.error;
 			return;
 		}
 		// Mirror a budget grant into the strip right away — the server applied it
--- a/src/lib/components/chat/AskQuestion.svelte.test.ts
+++ b/src/lib/components/chat/AskQuestion.svelte.test.ts
@@ -2,6 +2,13 @@ import AskQuestion from "./AskQuestion.svelte";
 import { render } from "vitest-browser-svelte";
 import { describe, expect, it, beforeEach, afterEach, vi } from "vitest";
 import type { ElicitationField, ElicitationRequestPayload } from "$lib/types/McpElicitation";
+import {
+	pendingQuestions,
+	registerQuestion,
+	unregisterQuestion,
+} from "$lib/stores/pendingQuestion";
+import { elicitationToResume } from "$lib/stores/elicitationResume";
+import { get } from "svelte/store";
 
 let sent: Array<Record<string, unknown>>;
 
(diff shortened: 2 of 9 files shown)


pngwn:
Addressed the three review findings in the second commit.

- **Crashed continuation read as continued forever** (Codex P1, Copilot): "continued" is now judged by whether a producer is live or a parked call is being claimed, and otherwise by whether the parked message already carries the Resolved update. The turn-state document is no longer consulted. Covered by "asks again after a continuation died before it persisted" and "leaves a continuation that already ran alone, whatever the turn state says".
- **System cancel replayed as the user's answer** (Copilot): the finalizer leaves durable prompts alone, rows record `resolution`, and a system close nothing consumed gives way to the user's real answer. Covered by the three "system close" cases and the new finalizer test.
- **Lookup failure dropped the question** (Codex P2, Copilot): a failed lookup returns a retryable 500 without the `answered` marker. Covered by "keeps the question open when it cannot tell whether the answer was continued".

PR #2571, opened 2026-09-10
Add resource group billing support

abidlabs:
This PR adds resource group billing support to Hugging Chat (and ML Intern). Specifically it will:

- expose the signed-in user's organization resource groups in the billing selector
- persist the selected organization and optional resource-group ID together
- bill Inference Providers to the resource-group ID via `X-HF-Bill-To`
- force ML Intern Jobs and Sandbox creation to the organization namespace plus selected resource group
- carry attribution through parked generations, the ML Intern session context, and budget reservation records
- clear resource-group selections that are no longer present in OAuth userinfo

This PR is based on #2567 and contains only the resource-group extension.

Depends on huggingface/hf-mcp-server#244 for the Jobs and Sandbox MCP arguments so that will need to be merged first before this can be tested.

But here's how the resource group selection will look (note that the resource groups appear for an organization after the forward slash):

image


pngwn:
I merged `main` into this branch (`66cad366`) so it applies on top of the squash-merged #2567. Each conflicted file is main's version plus this PR's own change, with three reconciliations worth knowing about:

- The flow bills through a payer target: the organisation with its resource group when one is selected, else the signed-in user. Personal stays enforced, as `main` made it.
- The billing-orgs endpoint stays on the shared userinfo helper (`billingOrganizations.ts`), which now carries each org's resource groups; the saved-value check in both settings endpoints covers a changed group as well as a changed org.
- The settings page keeps the awaited save with revert on refusal and gains the group options.

Type check, lint and the affected specs pass. I then reviewed the combined diff against `main`. Findings, most severe first. Nothing has been fixed yet.

### Correctness

1. **`hubBilling.ts:100` — sandbox create sends a flag the deployed server lacks.** The rewrite appends `--resource-group-id`, but the live `hf_sandbox` grammar has no such flag until hf-mcp-server#244 ships. With a group selected, every sandbox create either fails at the server's token parser or, if the flag is ignored, runs at the org root while the UI and the prompt's `BillTo` say the group pays. Nothing checks capability first; the tool mapping already keeps each tool's advertised schema and description, which is where a check belongs, or the group option needs a feature flag until the upstream deploys.
2. **`hubBilling.ts:58` — `resource_group_id` is added to `hf_jobs` read operations.** The set/delete sits after the `submitting` check, so `ps`/`logs`/`inspect`/`cancel` that name no namespace also carry a submission-only argument; the sandbox half correctly gates on `create`. Today the server's plain `z.object` schemas strip the key silently, but #244 adds it to submission schemas only, so the moment reads validate strictly every read and cancel in a group session fails, and the local preflight cannot catch it because `hf_jobs.args` is free-form.
3. **`billingOrganizations.ts:76` — the group's `role` is fetched but never enforced.** It travels from userinfo to the client and is rendered as an option, yet `assertBillableOrganization` checks membership only and the page lists every group. A member with `read` on a group can select it, the save is accepted, and every submission is then refused by the Hub as an opaque tool error. The org list has an eligibility filter (`plan || canPay`); the group list has none.
4. **`application/+page.svelte:143` — the reconcile path bypasses the setter's error handling.** After billing-orgs reports a cleared org or group, `onMount` calls `settings.instantSet` directly and discards its result; before this diff it went through the awaited setter. A failed POST there leaves `$settings` showing Personal while the persisted document still names the org and group, with no error and no revert.
5. **`mlAssistant.ts:119` — `BillTo` in the prompt is prose, not a namespace.** `mlAssistantBillingLabel` stamps `acme (resource group 65f…)`, while the jobs and sandbox contracts present `BillTo` as the namespace a job runs under. A model that copies it into a read's `namespace` gets it through untouched (reads that name a namespace are left alone) and the Hub 404s. Stamping the bare namespace and the group as a separate field keeps the identifier the prompt describes. The call-site comment in `textGeneration/index.ts` saying the prompt uses the dispatch rewrite's resolution is no longer true either.
6. **`settings/+server.ts:36` (both endpoints) — regex and refine failures surface as 500s.** `settingsSchema.parse` is uncaught, so a `billingResourceGroup` that fails the new 24-hex regex or the refine becomes a generic 500 rather than a 400 with the message. The client writes back `group.sub` verbatim, and the app never validates that format at the source. The uncaught-parse shape predates this PR; the regex adds a new client-reachable way to hit it.

### Reuse and simplification

7. **Schema and billing check duplicated across `api/v2/user/settings` and `settings/(nav)`.** The two blocks are byte-identical and this diff grew both with the same regex, empty-string union, refine and two-field comparison. One exported module would stop the next tweak landing in one file only.
8. **`X-HF-Bill-To` spread copy-pasted at five sites** (three in `endpointOai.ts`, one in `runMcpFlow.ts`, one in `api/transcribe`), each evaluating `inferenceBillingTarget` twice, and already non-uniform (only `runMcpFlow` gates on `isHuggingChat`). A `billToHeaders(locals)` helper in `billing.ts` collapses them.
9. **Two resolvers read the same billing locals unshared.** `inferenceBillingTarget` and `mlAssistantBillingTarget` each re-read and trim the two fields with their own inline locals type. The output shapes legitimately differ per wire protocol, but the shared read has no single home, so a precedence change in one splits a user's spend across payers with every spec still passing.
10. **Selection prefix encoding written in five places** on the settings page: formatted in `getBillingSelection` and both option templates, parsed by string slicing in `parseBillingSelection`. One encode/decode pair would do.

One candidate I refuted: sending the resource-group id in `X-HF-Bill-To` is what the Hub docs' cost-attribution section prescribes, so that part is right.

abidlabs:
Thanks for the review @pngwn! I resolved the critical comments from your review above, and have tested with a resource group to confirm everything works end to end:


https://github.com/user-attachments/assets/1d770095-580f-4d3c-9d5b-e94220ecba21

pngwn (approved):
This looks great!

Thanks for addressing those comments @abidlabs !

